PR #108, opened 2014-12-29
Subsite support for urlsegment

lekoala:
Current code does not take into account the Subsite module. The baseUrl should be the one of the owner subsite if it exists.

### DIFF
--- a/code/extensions/FluentSiteTree.php
+++ b/code/extensions/FluentSiteTree.php
@@ -78,8 +78,8 @@ class FluentSiteTree extends FluentExtension {
 		// Fix URLSegment field issue for root pages
 		if(!SiteTree::config()->nested_urls || empty($this->owner->ParentID)) {
             		$baseUrl = Director::baseURL();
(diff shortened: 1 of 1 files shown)


lekoala:
It's a good thing then, because it's working quite well! The only issue I have so far is the one mentioned here. I have two projects that use fluent/subsite together.
The only thing I find problematic is that languages are set for all subsites, but that's to be expected given the way fluent works (which is why I was wondering if it would be possible to set locale data at row level instead of column level). Currently I'm going around that with a little bit of code in the SiteConfig, but it has its limits.

tractorcow:
I have an idea in the works for solving the column number limit by using json column types (currently supported by postgres, see https://functionwhatwhat.com/json-in-postgresql/). I am thinking either a postgres-enhanced fork or addon-module would work, maybe once 3.2 is out. :P

tractorcow:
I've had long enough to come up with a better solution. sorry this took so long. :D

PR #119, opened 2015-06-03
FIX: Ensure deferred CMS panels are locale-specific (fixes #94)

### LINKED ISSUE #94: Site tree not updating titles
Hello, I have one problem with this otherwise great module. When locale is changed in the back-end, page titles in the Tree View don't update. This would be OK (they never update in Multilingual module, for instance) but if I reload the admin page, they do update to current locale. Moreover, the title of a saved page also updates, which results in the Tree View showing titles in mixed languages. Do I make any sense?

What I mean is, for example:
1. Locale is Polish. Tree View shows pages in Polish. OK.
2. Switch to English. Tree View still shows pages in Polish, although upon clicking them I can properly edit English data.
3. Edit some page and save it. Now it updates and shows in the Tree View in English, while other pages in the tree still show up with Polish titles. Confusing.
4. Reloading with CTRL-R fixes everything - whole Tree View shows titles in current locale.

Is this the intended behavior or have I done something wrong? Editing is fine and it's only a display issue with the Tree View, but it's very confusing for the editor.

kinglozzer:
- Moved `urlForLocale` so it’s not bound to a selector, changed it to accept a URL as an argument rather than assume `document.location.href` is the URL in question;
- Updated `$(".cms-fluent-selector .cms-fluent-selector-locales a").onclick()` to reflect new `urlForLocale` logic;
- Added `$('.cms-panel-deferred').onadd()` to add the fluent locale parameter to the deferred URL

I initially tried just disabling the deferred panels from ever being cached, but I think this is a better approach as it keeps that caching in-place. It’s not specific to the CMS tree view (#94 is the issue I’m trying to fix), but I don’t think that matters.

### DIFF
--- a/javascript/fluent.js
+++ b/javascript/fluent.js
@@ -1,20 +1,36 @@
-
-		
-				
 (function($) {
-	
 	/**
 	 * File: fluent.js
 	 * 
 	 * Injects locale navigation menu into the website
 	 */
 	$.entwine('ss', function($) {
+		/**
+		 * Determine the url to navigate to given the specified locale
+		 */
+		var urlForLocale = function(url, locale) {
+			// Get new locale code
+			param = {}
+			param[fluentParam] = locale;
+
+			// Check existing url
+			search = new RegExp('\\b'+fluentParam+'=[^&#]*');
+			if(url.match(search)) {
+				// Replace locale code
+				url = url.replace(search, $.param(param));
+			} else {
+				// Add locale code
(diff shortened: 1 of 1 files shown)


kinglozzer:
Thanks @tractorcow. Any chance of a new tag? :)

tractorcow:
https://github.com/tractorcow/silverstripe-fluent/releases/tag/3.2.1

PR #120, opened 2015-07-09
FEATURE: new reset translations button

unclecheese:
Adds a "reset translations to default" button to the CMS actions, which sets all translated fields to their current value in the default locale.

### DIFF
--- a/tests/FluentTest.php
+++ b/tests/FluentTest.php
@@ -822,10 +822,10 @@ class FluentTest extends SapphireTest {
 		$defaultValue = $page->$defaultField;
 		$page->resetTranslations();
 
-    	$this->assertEquals($page->Title_fr_CA, $defaultValue);
-    	$this->assertEquals($page->Title_en_NZ, $defaultValue);
-    	$this->assertEquals($page->Title_en_US, $defaultValue);
-    	$this->assertEquals($page->Title_es_ES, $defaultValue);
+    		$this->assertEquals($page->Title_fr_CA, $defaultValue);
+    		$this->assertEquals($page->Title_en_NZ, $defaultValue);
+    		$this->assertEquals($page->Title_en_US, $defaultValue);
+    		$this->assertEquals($page->Title_es_ES, $defaultValue);
 	}
 }
 


code/extensions/FluentCMSPageEditController.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+class FluentCMSPageEditController extends DataExtension {
+
+	public function doResetTranslations($data, $form) {
+		$record = $this->owner->getRecord($data['ID']);
+		$record->resetTranslations();
+
+        $this->owner->getResponse()->addHeader("X-Pjax","Content");
```
tractorcow: Indentation here is a bit broken. :D Can you also please add phpdoc on the method?:D

tests/FluentTest.php
```diff
@@ -813,6 +814,19 @@ public function testIsFrontendIgnorance() {
...
 	}
+
+
+	public function testResetTranslations() {
+		$page = $this->objFromFixture('FluentTest_TranslatedPage','translated1');
+		$defaultField = 'Title_'.Fluent::default_locale();
+		$defaultValue = $page->$defaultField;
+		$page->resetTranslations();
+
+    	$this->assertEquals($page->Title_fr_CA, $defaultValue);
```
tractorcow: Indentation is a bit weird here too.

tractorcow:
100% for actual code and unit tests. good job @unclecheese

tractorcow:
I just noticed an issue, if you reset the value for each locale to the default, what will happen is you remove the "inherit" behaviour from each field going forward.

E.g. you start with

```
field_default: val
field_locale1: val2
field_locale2: null
```

This means locale2 will failover to whatever is in the default.

What we are now doing is 

```
field_default: val
field_locale1: val
field_locale2: val
```

Which is good, until you edit the default locale again. This value in all other locales will STAY at val until saved in each language again.

A better reset solution is:

```
field_default: val
field_locale1: null
field_locale2: null
```

Which means all fields will fallback to the default, but will also inherit automatically if the default is changed, up to the point the page is saved in those locales again.

It might be a simple case of setting `$this->owner->$localeField = null` instead of `$defaultValue` but I'll test.

wernerkrauss:
Any progress on this?

tractorcow:
I would like to work on this but I keep getting bogged down with more important things. :) I need to stop putting this off...

dhensby:
Bump

tractorcow:
I'll fix this one day.

dhensby:
Today is one day!

tractorcow:
No it's not it's 6 days ago!

tractorcow:
I don't think this is going to get into 3.x sorry.

PR #126, opened 2015-07-15
FIX: Fixes #125 Duplicate locales in homepage URI

### LINKED ISSUE #125: Homepage view breaks when locale is used
SS 3.1.2 / Fluent 3.2.1 (using aliases) / CWP 1.0.7 (w/Translateable) PHP 5.5.9

The Problem: 

- Navigating to the homepage (clicking or futzing the URL) results in 404
- Seems to be to do with `FluentRootURLController::handleRequest()` line 93

If I dump `$localeURL.'/'.RootURLController::get_homepage_link().'/'` which is passed as a param to `$request->setUrl()` I get: `en/en/home/` and the page results in a 404 with the debugging removed.

This seems to be because `RootURLController::get_homepage_link()` is taking account of "Translateable" and appending the extra aliased locale from (decorated) SiteTree::$current_locale which Fluent also utlises.

The fix for me is simply to add `SiteTree::remove_extension('Translatable')` to mysite/_config.php which I think should be "Recommended" in Configuration.md, as it's currently buried in translatable.md which I hadn't read because (and call me pedantic..) I wasn't actually "Migrating"...thoughts?

phptek:


code/routing/FluentRootURLController.php
```diff
@@ -99,4 +99,22 @@ public function handleRequest(SS_HTTPRequest $request, DataModel $model = null)
...
+     * 
+     * With Translatable installed, don't pre-append the locale to the homepage 
+     * URL.
+     * 
+     * @return string
+     * @see {@link Translatable::get_homepage_link()}.
+     */
+    public static function fluent_homepage_link($localeURL) {
+        $homepageLink = parent::get_homepage_link();
+        if(class_exists('Translatable')) {
```
tractorcow: Is there ever a case where a user has translatable AND fluent, but wants to use the 'fluent home page' rather than the translatable one?

Should this be a configurable option that, by default (perhaps), acts as you have set above?

Really nice fix and looks good. :D

tractorcow: An in-code comment would help with future debugging as well.

```
// Here we don't append the prefix because <reason>. When translatable is installed <some change in behaviour> so this code is awesome and trustworthy because <comforting reassurance>
```

tractorcow:
Glad to have you help with this module @phptek !

code/routing/FluentRootURLController.php
phptek: Added warm, fuzzy comment block.
Not sure what you mean by "Fluent homepage". Do you mean a homepage ala `/en/home`? That works AFAIKT..

PR #131, opened 2015-07-17
FIX: Fixes #135 Increased range of native lang strings.

phptek:

Issue #130, opened 2015-07-17
Native language string isn't always returned


phptek:
Looking at `FluentExtension::LocaleInformation()`, for the `LanguageNative` value; In the event an ambiguous language is encountered such as Portugese which is spoken in two countries, then `false` is returned by the following out-of-the-box logic:

````php
i18n::get_language_name(i18n::get_lang_from_locale($locale), true)
````

The above makes use of `i18n::$common_languages` but is hamstrung by _not_ using a 2-char country code index for languages spoken in multiple countries (they use the full locale as keys). However there is `i18n::$common_locales` which _does_ have a "native" string we can utlise and _is_ keyed by the full locale.

Summary: FluentExtension::LocaleInformation() should have access to all native language translations available in the framework...fix incoming

phptek:
N.b I haven't looked at the _entire_ differences between the languages contained in each array, so a bit of self one-upmanship might be (if required) to merge and filter the two arrays for our own purposes, such that we only have the one source.

Thoughts??

tractorcow:
Fixed with https://github.com/tractorcow/silverstripe-fluent/pull/131

PR #134, opened 2015-07-20
FIX: Fixes #124 Bad data-url encoding causes CMS page-search bug.

### LINKED ISSUE #124: CMS Page Filter is broken
If I install Fluent in SS 3.1.12, I can no longer filter on PageTypes or perform a standard "Content" text search from within the CMS' "Filter" located in the centre pane. Instead, a complete list of _all_ page objects is returned as if I had instructed the CMS to build me the standard SiteTree. It isn't even a search-result that I can "clear".

I can try and futz the URL to include/exclude the "l" parameter to no avail e.g. 

`/?locale=en_NZ&q[FilterClass]=CMSSiteTreeFilter_Search&q[ClassName]=MyPageType`

or 

`/?locale=en_NZ&l=en_NZ&q[FilterClass]=CMSSiteTreeFilter_Search&q[ClassName]=MyPageType`

Removing the module, and it works again. By "Remove" I mean delete the "fluent" directory.

phptek:


tractorcow:
Ok, it looks like the actual bug is in `$.path.addSearchParams(url, param);` in https://github.com/silverstripe/silverstripe-framework/blob/3.2/admin/javascript/lib.js#L164

It doesn't seem to know how to add the `&l=en_US` to a querystring that contains escaped components. It ends up double-escaping each rather than simply appending the additional param.

I suggest to instead fix this in the fluent `urlForLocale` method to manually add the parameter, rather than using the lib.js version. It'd probably be as simple as checking if we should use `?` or `&` and then adding the param on the end of the url (but before the hash).

I am worried that, although the suggested fix does fix this issue, it feels like the problem is being fixed in the wrong place.

phptek:
@tractorcow Fair enough :-) As it is though, I'm having to run a Gitlab fork with outstanding (incorrect) PRs merged there, because I don't have much more time to spend patching some of the issues I find (and I _know_ my client will find) perhaps to your honourable satisfaction :-)

The more important one I'm keen to see you respond to is #131.

Thanks man :-)

tractorcow:
Fixed with https://github.com/tractorcow/silverstripe-fluent/commit/32778f8ea91e3332db17af7efaf07d28e31ddfb4

tractorcow:
Well I've pushed up a fix to this issue, as well as merging #130.

I hope you can continue your work now. :D

PR #135, opened 2015-07-23
Small Input: Hide Flydown w/ <2 added languages

Xarinor:
This is not a solid fix/solution but should help us keeping the backend
clean when a website is prepared for fluent but has no secondary languages assigned yet.


Xarinor:
Closed due to comitting the wrong/bugged version.

PR #136, opened 2015-07-24
BUG Fix creation of records in non-default locale

### LINKED ISSUE #132: Default translated field value doesn't seem logical
## The Problem:

First-time page-edits, post module install using the non-default locale in the CMS, will render the non-localised field's value as per the value of selected locale's field-value.

## Reproduction

Assuming a default locale of `en_NZ` and some other locale is configured e.g. `ar_EG` (Arabic, Egypt):

* Login to the CMS
* Navigate to any page in the CMS
* Observe that page's "Title" reads in English (Probably)
* Perform a DB query on the SiteTree table for the "Title", "Title_en_NZ", "Title_ar_EG" fields
* Observe that by default, "Title" contains an English string, "Title_en_NZ" is NULL, "Title_ar_EG" is NULL
* In the CMS select "Arabic (Egypt)" from the CMS' locale selector
* Enter an "Arabic" title
* Press "Save"
* Observe the Title field in the CMS now reads in Arabic
* Run the same SQL query as above
* Observe that the "Title" column contains Arabic, "Title_en_NZ" is still NULL, and "Title_ar_EG" is in Arabic
* Observe in the frontend in "Stage" mode, that that page's title is also in Arabic, regardless of any frontend language selection made
* In the CMS select "English (New Zealand)" from the CMS' locale selector
* Observe that the visible title is in Arabic

## Summary

If I had selected "Save and Publish" instead of just "Save" then that page's title would also be in Arabic, and that change would be live.

...this doesn't seem to be logical behaviour to me. Perhaps I'm coming at things from an odd angle, but I'd expect the YML-configured `default_locale` version of the field to be pre-populated with whatever is already in the "Title" field (or make this behaviour configurable) and for the un-translated field to be left-alone for subsequent updates.

Possibly I might even expect that any query augmentation totally ignores the un-translated fields while `Fluent` is installed. But I'm not sure of the wider impact of this sort of thinking.

Might this be related to a response in #128 regarding an inherent difficulty in creating fields with the values of other, pre-existing fields?

tractorcow:
Fixes #132

### DIFF
--- a/code/extensions/FluentExtension.php
+++ b/code/extensions/FluentExtension.php
@@ -403,21 +403,30 @@ class FluentExtension extends DataExtension {
 		// E.g. if loaded as zh_CN, then even if the current locale is now en_NZ we should save
 		// writes to this object as though it was still in chinese locale
 		$locale = $this->owner->getSourceQueryParam('Fluent.Locale') ?: Fluent::current_locale();
+		$defaultLocale = Fluent::default_locale();
 
 		// Prior to writing, we should flag any translated field as changed if its local value differs from
 		// its translated value, in case change detection would prevent this from being written to the DB
 		$translated = $this->getTranslatedTables();
 		foreach($translated as $table => $fields) {
 			foreach($fields as $field) {
-
-				// Extract both base value (which could have been extracted from the subfield on load)
-				// and compare it to the localised field value
-				$localeField = Fluent::db_field_for_locale($field, $locale);
+				// Get assigned value for this field
 				$value = $this->owner->$field;
-				$localeValue =  $this->owner->$localeField;
 
-				// If these values differ, but a change isn't detected, then force a change
-				if($this->owner->exists() && ($value != $localeValue) && !$this->owner->isChanged($field)) {
+				// If creating a new record in non-default locale, ensure a default value exists for each field
+				$defaultField = Fluent::db_field_for_locale($field, $defaultLocale);
+				if(!$this->owner->exists()) {
+					if($locale !== $defaultLocale && empty($this->owner->$defaultField)) {
+						$this->owner->$defaultField = $value;
+					}
+					continue;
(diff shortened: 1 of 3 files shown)


code/extensions/FluentExtension.php
```diff
@@ -403,21 +403,30 @@ public function onBeforeWrite() {
...
+				// Get assigned value for this field
+				$value = $this->owner->$field;
 
-				// Extract both base value (which could have been extracted from the subfield on load)
-				// and compare it to the localised field value
+				// If creating a new record in non-default locale, ensure a default value exists for each field
+				$defaultField = Fluent::db_field_for_locale($field, $defaultLocale);
+				if(!$this->owner->exists()) {
+					if($locale !== $defaultLocale && empty($this->owner->$defaultField)) {
+						$this->owner->$defaultField = $value;
```
tractorcow: This is the bit that ensures the default locale has a value.

code/extensions/FluentExtension.php
```diff
@@ -403,21 +403,30 @@ public function onBeforeWrite() {
...
+
+				// If the base (or localised) field has changed, but isn't detected, force a change
+				// Also ensure that if the main field is changed, mark all fields as changed
 				$localeField = Fluent::db_field_for_locale($field, $locale);
-				$value = $this->owner->$field;
 				$localeValue =  $this->owner->$localeField;
-
-				// If these values differ, but a change isn't detected, then force a change
-				if($this->owner->exists() && ($value != $localeValue) && !$this->owner->isChanged($field)) {
+				if( ($value != $localeValue) || $this->owner->isChanged($field)){
```
tractorcow: This is a fix for the default value being overwritten in subsequent writes.

tractorcow:
Postgres is being weird, but this fixes your issue.

PR #147, opened 2015-09-14
BUG Fix default row values not being properly loaded when saving with…

tractorcow:
Fix default row values not being properly loaded when saving with fluent for the first time
Postgres can now fail due to no proper support for numeric fields (maybe one day)

Fixes https://github.com/tractorcow/silverstripe-fluent/pull/141, probably.

@phptek do you want to review this?

Also, annoyingly, I couldn't get postgres to work properly with numeric fields (it complains when comparint int columns to '') so I've temporarily added it to allowed failures.

### DIFF
--- a/code/extensions/FluentExtension.php
+++ b/code/extensions/FluentExtension.php
@@ -496,15 +496,16 @@ class FluentExtension extends DataExtension {
 	 * @return string Select fragment
 	 */
 	protected function localiseSelect($class, $select, $fallback) {
-		return "CASE
-				WHEN (\"{$class}\".\"{$select}\" IS NOT NULL AND \"{$class}\".\"{$select}\" != '')
-				THEN \"{$class}\".\"{$select}\"
-				ELSE \"{$class}\".\"{$fallback}\" END";
+		return "CASE COALESCE(CAST(\"{$class}\".\"{$select}\" AS CHAR), '')
+				WHEN '' THEN \"{$class}\".\"{$fallback}\"
+				WHEN '0' THEN \"{$class}\".\"{$fallback}\"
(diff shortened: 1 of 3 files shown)


tractorcow:
Finally got postgres to work. :D

code/extensions/FluentExtension.php
```diff
@@ -547,10 +555,11 @@ public function augmentSQL(SQLQuery &$query, DataQuery &$dataQuery = null) {
 			// depending on field nullability.
 			// If the filterColumn is null or empty, then it's considered untranslated, and
 			// thus the query should continue running on the default column unimpeded.
+			$castColumn = "COALESCE(CAST($filterColumn AS CHAR), '')";
 			$where[$index] = "
-				($filterColumn IS NOT NULL AND $filterColumn != '' AND ($localisedCondition))
+				($castColumn != '' AND $castColumn != '0' AND ($localisedCondition))
```
tractorcow: I could probably simplify this as an `in ('', '0')` condition.

tractorcow:
Thanks for the peer review me! :D

uniun:
This fix breaks ability to have translated Boolean fields (Integer too?).
[FluentExtension.php#L545](https://github.com/tractorcow/silverstripe-fluent/blob/master/code/extensions/FluentExtension.php#L545)

Sometimes it is useful to have "tranlated" CheckboxFied.

tractorcow:
Can you make a new issue for this please?

tractorcow:
Boolean and Integer values are also ment to "fail over" if they are set to 0 (int 0 or false).

tractorcow:
I suppose I could add a tri-state boolean DBField; -1 is false, 0 is inherit, and 1 is true.

PR #152, opened 2015-10-06
turning spaces in tabs and adding some phpdoc blocks

spekulatius:
Nothing crazy here - but w=1 should display some minor adjustments (removing two unneeded lines and adding two phpdoc blocks)


tractorcow:
You're everywhere. ;)

PR #153, opened 2015-10-12
adding public for better readibility

spekulatius:


dhensby:
Makes sense

tractorcow:
Thanks @dhensby :D

PR #172, opened 2015-12-02
BUG Fix 

prij:
issue where the current locale would not be assigned to the correct domain

### DIFF
--- a/code/Fluent.php
+++ b/code/Fluent.php
@@ -91,11 +91,9 @@ class Fluent extends Object implements TemplateGlobalProvider
         // Check browser headers
         if (empty($locale)) {
             $locale = self::detect_browser_locale();
+        }
         
         // Fallback to default if empty or invalid (for this domain)
-        /*if(empty($locale) || !in_array($locale, self::locales(true))) {
-        $locale = self::default_locale(true);
-        }*/
         $caresAboutDomains = Fluent::is_frontend();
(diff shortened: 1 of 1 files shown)


tractorcow:
Looks like a parse error in your PHP @prij . I guess I can fix it for you. :)

code/Fluent.php
```diff
@@ -91,11 +91,14 @@ public static function current_locale($persist = true)
         // Check browser headers
         if (empty($locale)) {
             $locale = self::detect_browser_locale();
-        }
```
tractorcow: I'm not sure this should have been dropped off?

prij:
Woops sorry @tractorcow, I've just pushed that missing brace

code/Fluent.php
```diff
@@ -92,10 +92,14 @@ public static function current_locale($persist = true)
         if (empty($locale)) {
             $locale = self::detect_browser_locale();
         }
-
-        // Fallback to default if empty or invalid
-        if (empty($locale) || !in_array($locale, self::locales())) {
-            $locale = self::default_locale();
+        
+        // Fallback to default if empty or invalid (for this domain)
```
tractorcow: You also need to put this code back. :)

prij: As we discussed :) Pushed.

tractorcow:
Merged manually with https://github.com/tractorcow/silverstripe-fluent/commit/ceefc0a731585a1e296c85b0f0674e7b1e9e7d0d

PR #176, opened 2016-01-04
Silverstripe 3.2 cms site tree do not refresh properly on language se…

mkaczmarczyk:
Hi,
since on Silverstripe 3.2 cookies are flaged httpOnly by default, cms site tree do not refresh properly on language change.


tractorcow:
Thanks, that probably solved a LOT of problems!

mkaczmarczyk:
I'm glad to hear this :)

PR #177, opened 2016-01-18
added a language label next to the fluent icon

TheBnl:
Added a small label next to the icon.

Some users experience pages in the CMS not completely finishing the load. What happens is when a user that is editing the original language, for example en_NZ, and then wants to start translating a new language lets say nl_NL navigates to nl_NL. The loading animation starts.. and finishes even the sidebar is refreshed and the label on the fluent select box in the menu bar updates. But the new page is not really loaded. This is a CMS issue but has mayor impact in the usability of Fluent because a user gets confused thinking he/she is editing nl_NL but is really editing en_NZ. 

So by adding the small label a user can always check that he/she is editing the language that is requested.

Also in SS 3.2 the menu bar is collapsed by default, thus hiding the current language. With this label the current language is always visible.

screen shot 2016-01-18 at 11 19 57

Kind regards,
Bram


tractorcow:
Wow, this is very nice. :D

You'll need to be careful though, some language codes are 3 letters not 2. How about doing something like `strtok($locale, '_')` to get the language?

code/extensions/FluentExtension.php
```diff
@@ -715,7 +715,13 @@ public function updateCMSFields(FieldList $fields)
                 }
 
                 // Highlight any translated field
-                if ($field) {
+                if($field && !$field->hasClass('LocalisedField')) {
+                    $field->addExtraClass('LocalisedField');
```
tractorcow: I think this is duplicated below? That extra check you added is a good idea though.

TheBnl:
Solid addition, i've replaced substr() with strtok() as you suggested and removed the duplicate addExtraClass. Should have seen that one before commiting!

tractorcow:
Ta!

TheBnl:
Nice! thanks :smile:

tractorcow:
No, thanks to you. This module is only good because people use it. :)

PR #178, opened 2016-01-20
made the collapsed fluent menu readable by using shortened language labels

TheBnl:
Hi!

Back with another addition, added two labels to the fluent menu selector. One long label for the cms-menu open state and a shortened label for the cms-menu collapsed state. A title attribute is placed on the li with the full name of the language, so when the menu is collapsed and multiple english dialects are used the user can still find out what the dialect is by hovering over the label.

screen shot 2016-01-20 at 10 12 48

cheers!
Bram


javascript/fluent.js
```diff
@@ -42,9 +42,11 @@
 				// Create options
 				$.each(fluentLocales, function(locale, name){
 					var item = $("<li><a></a></li>");
+					var lables = "<span class='full-title'>"+name+"</span><span class='short-title'>"+locale.split("_")[0]+"</span>";
```
tractorcow: Ah, a slight XSS risk here; We should not trust `name` to be XML escaped.

Try making labels a html string and setting the properties this way, safely, using `.text()` (which does the escaping for us).


```javascript
var item = $("<li><a><span class='full-title'></span><span class='short-title'></span></a><li>");
$(".full-title", item).text(name);
$(".short-title", item).text(locale.split("_")[0]);
$("a", item)
    .attr('data-locale', locale)
	.attr('title', name);
$(".cms-fluent-selector-locales", selector).append(item);
```

tractorcow:
Thanks for the submission.. just noted an escaping issue, but easy to fix.

TheBnl:
Sharp! I've fixed the issue as you proposed

tractorcow:
Pointy!

Issue #171, opened 2015-11-26
Domain not passed in functions


zrii:
I had a problem with setting domain per locale, since we had a couple of websites running different languages, one per domain.

The fix was hardcoded, didn't have the time to look where the functions are called for a pull request.

Fluent.php

``````php
public static function default_locale($domain = null) {
```
with
``````php
public static function default_locale($domain = true) {
```

AND

```php
public static function locales($domain = null) {
```
with
``````javascript
public static function locales($domain = true) {
```

Maybe if I have time I will make a pull request.

tractorcow:
I think @prij had an issue with a fix that she was going to push up for me. :) Still waiting! :D

tractorcow:
@zrii can you please describe what the bug was? You have given me the fix but not actually explained what necessitated it. :D

tractorcow:
If I recall correctly, the fix @prij was going to add was to set the parameter correctly in Fluent::current_locale 

```php
// Fallback to default if empty or invalid
$domainMode = <something>;
if(empty($locale) || !in_array($locale, self::locales())) {
	$locale = self::default_locale($domainMode);
}
```

zrii:
OK the issue was the following.
One website would be set on 3 languages on 3 domains, one domain each language (nl, en, sr, with according TLD .nl, .en, .rs), with default dutch language.

Settings are the following in fluentconfig.yml:

```yml
---
Name: myfluentconfig
After: '#fluentconfig'
---
Fluent:
  detect_locale: false
  default_locale: nl_NL
  locales:
    - nl_NL
    - en_US
    - sr_RS
  aliases:
    nl_NL: nl
    en_US: en
    sr_RS: sr
  force_domain: true
  domains:
    example.nl:
      locales:
        - nl_NL
      default_locale: nl_NL
    example.us:
      locales:
        - en_US
      default_locale: en_US
    example.rs:
      locales:
        - sr_RS
      default_locale: sr_RS
---
Name: myfluenti18nconfig
After: '#fluenti18nconfig'
---
i18n:
  default_locale: nl_NL
```

Then pages would be made. 
Example page would be example.nl/page-one. In English it would be example.us/page-one-english same page in different language and different url segment.
If you would visit home page of English, example.us/ and then go to example.us/page-one-english there would be no problem.

But if you would open incognito view and open example.us/page-one-english directly, you would get a 404 error. It would 'fix' if you visit homepage and it switches language and then go to the page example.us/page-one-english and get it.

[...]

PR #183, opened 2016-02-16
extend domain-configuration.md

dkriegner:
In order to help prevent the noise I recently caused on #171 I propose some changes to the documentation which might help new users to better understand the structure of the config file in the case of multiple domains.


tractorcow:
Nicely put, thanks for testing it out !

PR #187, opened 2016-04-03
Adding LanguageAlias to LocaleInformation data

3Dgoo:
It might be useful for some people to use a language alias in some cases. I have added `LanguageAlias` to the data returned by `LocaleInformation` so a user can easily use it in a custom template.

### DIFF
--- a/code/extensions/FluentExtension.php
+++ b/code/extensions/FluentExtension.php
@@ -402,6 +402,7 @@ class FluentExtension extends DataExtension
             'Alias' => Fluent::alias($locale),
             'Title' => i18n::get_locale_name($locale),
             'LanguageNative' => Fluent::locale_native_name($locale),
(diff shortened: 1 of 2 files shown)


tractorcow:
Please fix the unit tests to include the new parameter.

I don't thin `LanguageAlias` is quite correct, since it's not an alias, and we already have an alias.

I think `Language` by itself is probably fine. What do you think?

3Dgoo:
`Language` sounds good to me. I'll try to fix up those things up when I get a chance.

tractorcow:
Ta; I'll wait for the tests and I'll merge. :D

3Dgoo:
I have changed the variable to `Language` and added tests to check these values. Does that look OK?

tractorcow:
I'll restart the tests and let you know. ;D

tractorcow:
Gr, test install timing out due to error. Will retry later.

PR #189, opened 2016-04-17
[#188] Add visual indicator for whether a field has been changed 

### LINKED ISSUE #188: Feature request: Indicator for when a value is changed from the default value
Assuming I have a basic Page, I install this module, and I configure two locales: en_NZ (default) and en_AU.

I change the Page's Title to "Foobar" and save it. When I move to the AU locale, the change shows through there as well. If I then change the Title at the AU locale to "Foobar Australia" and save, it shows the AU Title as expected. Navigating back to the NZ (default) locale shows "Foobar" again, as expected.

**Request:**

> Given that I have changed a value at a non-default locale level,
> I want to see a visual indicator that the value is different from the default locale's value,
> So that I am instantly aware that the value is different from the default value.
> 
> Given that I have _not_ changed a value at a non-default locale leve,
> I want to see a visual indicator that the value is the same as the default locale's value,
> So that I am instantly aware that changes will show here if changed at the default locale scope.

Hope this makes sense - it would just be a handy thing to be able to see where values are essentially "static", in that they've been customised, and will not change when the default locale's value changes.

**An example:**

In Magento, this very system is in place in the form of a checkbox with a label saying "Use default value." If the value is the same at the en_AU scope as it is at the en_NZ scope, the checkbox is ticked, and the textbox itself is in a disabled state. If you uncheck the "use default value" box, you can modify the textbox at that locale.

The implementation difference I see between that example and this module (possibly - not sure how this module works ATM) is that in Magento's ecosystem you're talking about EAV records, so by using a default value you're just _not looking for a locale value_ but going straight to the default value. Does this module copy the values from default scope to each locale on save, or does something similar exist?

Thanks for reading!

robbieaverill:
This pull request resolves #188 

It changes the green locale indicator to orange if the value of the field has been changed from the default locale's value.

If the value is the same as the default locale's value, or is null (inheriting) then it will remain green.

Tooltip/title attributes added in both cases to explain.

Feedback welcomed!


robbieaverill:
Example: If the Navigation label is changed from the default locale (en_NZ) value of "My Cool Page" to "My Cool Pag" at the en_GB locale, the indicator becomes orange:

my example

code/Fluent.php
```diff
@@ -669,4 +669,40 @@ public static function get_template_global_variables()
...
+            return false;
+        }
+
+        $defaultField = self::db_field_for_locale($field->getName(), $defaultLocale);
+        $localeField  = self::db_field_for_locale($field->getName(), $locale);
+
+        $defaultValue = $object->$defaultField;
+        $localeValue  = $object->$localeField;
+
+        if ((!is_null($defaultValue) && is_null($localeValue))
```
tractorcow: Actually, string values can be "mapped" to null, so can we use `empty()` instead of `is_null` to check for has-value?

Fluent fields do not distinguish between null / empty, since that distinction is often ambiguous in core. Thus we are not able to use, for example, `null` for untranslated, or '' for translated as empty.

The problem is that you can end up with empty strings appearing as though they are modified, but they aren't really. :P

tractorcow: Another trick you could try is to cast as string and then do type comparison.

E.g.

`return (string)$defaultValue === (string)$localeValue;`

Which will eliminate the risk of non-significant changes appearing as though they have delta.

tractorcow:
This is an excellent A+ change. My only concern is the checking for empty values... could you please review my feedback and let me know if you concur?

code/Fluent.php
tractorcow: Sorry, removed my previous second reply... also need to treat `0` as empty values, even if they are strings. :P

robbieaverill: I'd tend to agree with that - but should we be concerned with treating `0` as an empty value? I suppose that if the field is numeric it shouldn't be translatable...? I'll change anyway and add a unit test then re-push

robbieaverill: Do you have a preference for implementing the JavaScript to make the visual indicator dynamic at this point?

robbieaverill: So we're on the same page, you're talking about changing to this?

```php
 
-        if ((!is_null($defaultValue) && is_null($localeValue))
+        if ((!empty($defaultValue) && empty($localeValue))
             || ($defaultValue === $localeValue)
         ) {
             // Unchanged from default
```

robbieaverill:
Hey @tractorcow - I've added the JavaScript handler for this. It's a separate commit for now in case you want to shoot it - take a look and let me know, can squash if you'd like. Thanks!

robbieaverill:
Added: click the locale label when it's in the "modified" state to reset to the default locale's value.

Similar to #120

tractorcow:
Wow, I'm really impressed... nice last minute addition!

Since we're adding front-end functionality, I'm going to check this PR out and do a bit of local testing. Really nice job. :D

robbieaverill:
Happy to help! Let me know if you want any changes

On Tue, Apr 19, 2016 at 1:15 PM, Damian Mooyman <notifications@github.com>
wrote:

> Wow, I'm really impressed... nice last minute addition!
>
> Since we're adding front-end functionality, I'm going to check this PR out
> and do a bit of local testing. Really nice job. :D
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/tractorcow/silverstripe-fluent/pull/189#issuecomment-211668217>
>



-- 
*Robbie Averill*
LA PETITE MANOUCHE - NZ Gypsy Jazz Guitar Duo
Phone: +64 27 321 4758
contact@lapetitemanouche.com
www.lapetitemanouche.com

tractorcow:
Any changes? Can you change it so I never have to deal with daylight saving again? :P

robbieaverill:
Perhaps if you remove times and dates altogether you could get around that.
AJAX call to Google for the time instead...? :)

On Tue, Apr 19, 2016 at 2:16 PM, Damian Mooyman <notifications@github.com>
wrote:

> Any changes? Can you change it so I never have to deal with daylight
> saving again? :P
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/tractorcow/silverstripe-fluent/pull/189#issuecomment-211691192>
>



-- 
*Robbie Averill*
LA PETITE MANOUCHE - NZ Gypsy Jazz Guitar Duo
Phone: +64 27 321 4758
contact@lapetitemanouche.com
www.lapetitemanouche.com

tractorcow:
ok, I've tested it and I have some slight feedback.

Since the change icons are actually buttons, I think we need some kind of hover-indicator to help users realise that they can be clicked on.

Can we make it:
same value: `green` (cursor: default)
different value: `orange` (cursor: pointer to show that you can click)
different value hover: `green` again (to indicate you are restoring this to green)

robbieaverill:
Sure thing - will add now

On Wed, Apr 20, 2016 at 12:21 PM, Damian Mooyman <notifications@github.com>
wrote:

> ok, I've tested it and I have some slight feedback.
>
> Since the change icons are actually buttons, I think we need some kind of
> hover-indicator to help users realise that they can be clicked on.
>
> Can we make it:
> same value: green (cursor: default)
> different value: orange (cursor: pointer to show that you can click)
> different value hover: green again (to indicate you are restoring this to
> green)
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/tractorcow/silverstripe-fluent/pull/189#issuecomment-212182390>
>



-- 
*Robbie Averill*
LA PETITE MANOUCHE - NZ Gypsy Jazz Guitar Duo
Phone: +64 27 321 4758
contact@lapetitemanouche.com
www.lapetitemanouche.com

tractorcow:
Thanks @robbieaverill , great job!

PR #199, opened 2016-04-28
[#198] Visual indicator: fix for textareas

### LINKED ISSUE #198: Visual indicator: Resetting to default locale value doesn't work on TextareaFields
Just logging this as a bug for reference, I will patch it tomorrow.

The "Click here to reset to the default value" visual indicator button works for text inputs, but not for textareas.

Looks like the problem is to do with how the default value (JSON encoded) is inserted into the DOM. Will probably need to be escaped first.

Caused by #189

robbieaverill:
* Support inputs and textareas (resolves #198)
* Decode JSON from an HTML attribute before parsing it

### DIFF
--- a/javascript/fluent.js
+++ b/javascript/fluent.js
@@ -166,6 +166,10 @@
 					input.val(defaultValue);
 				} else if (input.is('textarea')) {
 					input.text(defaultValue);
+					// If it's a WYSIWYG editor, trigger it to update
+					if ('true' === input.attr('tinymce')) {
+						tinyMCE.get(input.attr('id')).setContent(defaultValue);
+					}
 				}
 				input.change();
 			}
--- a/javascript/fluent.js
+++ b/javascript/fluent.js
@@ -128,13 +128,18 @@
 			}
 		});
 
-		$('div.LocalisedField input.LocalisedField').entwine({
+		$('div.LocalisedField .LocalisedField').entwine({
 			/**
 			 * Check for changes against the default value
 			 */
 			onchange: function() {
-				var newValue = this.val();
-				var defaultValue = $.parseJSON(this.data('default-locale-value'));
+				if (this.is('input')) {
+					var newValue = this.val();
+				} else if (this.is('textarea')) {
+					var newValue = this.text();
+				}
+
+				var defaultValue = this.data('default-locale-value');
 
 				if (!defaultValue) {
 					// We'll turn this off on the default locale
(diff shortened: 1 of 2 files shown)


javascript/fluent.js
```diff
@@ -125,16 +125,29 @@
 				} else {
 					label.attr('title', 'Using default locale value');
 				}
+			},
+			/**
+			 * The JSON in the data attribute will be encoded, so this is a bit of a hack to decode it before parsing
```
tractorcow: Sorry, but how is it encoded? It's XML attribute encoded right?

tractorcow: Is it possibly being double encoded?

robbieaverill: If you view the source you'll see something like this: `data-default-locale-value="&quot;i&#039;m a default value&quot;" />`. This is a result of `json_encode('I'm a default value')`, which would be `"I'm a default value"`, yet when it's inserted into the DOM it gets encoded as such (assuming by the SS framework)

tractorcow: Should we perhaps not json encode the string at all?

robbieaverill: I'm guessing it's [`Convert::raw2xml`](https://github.com/silverstripe/silverstripe-framework/blob/3/core/Convert.php#L99-L106) that makes it so, called from `FormField::getAttributesHTML`

robbieaverill: @tractorcow let me run a quick test without JSON encoding and see - I guess it should work the same way :)

robbieaverill: Yep, works fine without JSON, pushing the updated commit now...

javascript/fluent.js
```diff
@@ -128,13 +128,18 @@
...
 			/**
 			 * Check for changes against the default value
 			 */
 			onchange: function() {
-				var newValue = this.val();
-				var defaultValue = $.parseJSON(this.data('default-locale-value'));
+				if (this.is('input')) {
+					var newValue = this.val();
+				} else if (this.is('textarea')) {
+					var newValue = this.text();
```
tractorcow: Have you tested this for tinymce? it uses a textarea I think.

robbieaverill: Damn - I'd become pretty happy having gotten rid of the damn WYSIWYG editor from my environment :) Anywho - it does update the textarea, but the iframe that is run by Tiny doesn't update with the value in the hidden textarea. Pushing an update for Tiny now...

javascript/fluent.js
```diff
@@ -166,6 +166,10 @@
 					input.val(defaultValue);
 				} else if (input.is('textarea')) {
 					input.text(defaultValue);
+					// If it's a WYSIWYG editor, trigger it to update
+					if ('true' === input.attr('tinymce')) {
+						tinyMCE.activeEditor.setContent(defaultValue);
```
tractorcow: Is it possible (for the sake of multi-editor environments) to get the correct instance?

`tinymce.get('editor');` will get an editor by ID. That way multiple editors won't end up getting all the same content. :P

http://archive.tinymce.com/wiki.php/API3:method.tinymce.get

robbieaverill: Good call - I'd only skim read the docs. Updated. #whatsanedgecaseanyway

tractorcow:
Don't forget to post a reply when you update the pull request. I don't get a notification if you just push up a new commit. ;D

PR #202, opened 2016-05-02
Add config option to disable locale prefixes in links

### LINKED ISSUE #201: Cannot disable URL prefix on Link()
In some cases I've found it may be useful to be able to call `Page::Link()` and have it *not* add the locale alias prefix to the return, e.g.: `/en_NZ/my-pages/my-page/child-here`.

I'd like to be able to retrieve the locale-less page link in some situations (but not all!).

OK if I PR a config addition for it?

robbieaverill:
* Add test
* Add to fluent.yml as example
* Resolves #201

### DIFF
--- a/_config/fluent.yml
+++ b/_config/fluent.yml
@@ -44,6 +44,8 @@ Fluent:
   persist_id_cms: 'FluentLocale_CMS'
   # Request parameter to store the locale in. Used for some Controllers
   query_param: 'l'
+  # Option to disable adding the locale prefix to URLs. This can be set inline for one off use.
+  disable_link_locale_prefix: false
   # Determine if the site should detect the browser locale for new users.
   # Turn this off to disable 302 redirects on the home page
   detect_locale: false
--- a/code/extensions/FluentSiteTree.php
+++ b/code/extensions/FluentSiteTree.php
@@ -34,6 +34,10 @@ class FluentSiteTree extends FluentExtension
 
     public function updateRelativeLink(&$base, &$action)
     {
+        // Option to prevent the prefix being added if required
+        if (Fluent::config()->disable_link_locale_prefix) {
+            return;
+        }
 
         // Don't inject locale to subpages
         if ($this->owner->ParentID && SiteTree::config()->nested_urls) {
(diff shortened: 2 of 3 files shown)


tractorcow:
I think the change is well coded, but I feel that it lacks justification.

As it stands, turning on this feature will immediately stop the locale menu from working, since all the links in the menu will be the same. How will the locale navigation work with this feature on?

robbieaverill:
I intended it to be used inline rather than always, so perhaps the example should be taken out of `fluent.yml` to avoid confusion, e.g. If I want `/my-page` instead of `/en_NZ/my-page` I could run the following:

```php
$config = Config::inst();
$config->update('Fluent', 'disable_link_locale_prefix', true);
$link = $page->Link();
$config->update('Fluent', 'disable_link_locale_prefix', false);
```

The rest of the execution would then continue as per normal.

As I say, I didn't intend for this change to be applied always as I'm sure there are better ways to configure the module to avoid the prefix, e.g. domains or whatnot.

In my application I'd like to keep the prefixes, but in some situations I'd like to be able to retrieve the full page link without the prefix. `URLSegment` isn't complete, and the extension doesn't provide any way to prevent `updateRelativeLink()` from running other than `SiteTree::config()->nested_urls` which I assume would be a broader change.

That being said, the same concept as the example above could probably be applied to that config setting as well, although I'm almost entirely sure that it would be a bad idea :)

Thoughts?

tractorcow:
How about another method that returns UntranslatedLink() ?

robbieaverill:
So thinking out loud that would need to duplicate logic from somewhere
right (possibly `SiteTree::RelativeLink`)? If it's just calling `Link()`
then it would have the `updateRelativeLink` extension called

robbieaverill:
So thinking out loud that would need to duplicate logic from somewhere right (possibly `SiteTree::RelativeLink`)? If it's just calling `Link()` then it would have the `updateRelativeLink` extension called

tractorcow:
I think instead of this feature, could I add a "disable prefix for default locale" option?

robbieaverill:
It's up to you- your code after all! Personally I need the URLs without
locale prefixes everywhere, but I'm also happy to remove them manually or
make hacks to get rid of them instead of merging in a useless configuration
value



-- 
*Robbie Averill*
LA PETITE MANOUCHE - NZ Gypsy Jazz Guitar Duo
Phone: +64 27 321 4758
contact@lapetitemanouche.com
www.lapetitemanouche.com

tractorcow:
> It's up to you- your code after all! 

I know, but I want to make sure that users can solve the problems they encounter. :D It's no good writing the code for myself if no one uses it.

PR #213, opened 2016-06-13
FIX: Fixes #211 

### LINKED ISSUE #211: Use of Browser navigation buttons causes confusion
When using a browser's "Back" and "Forward" buttons, I can reproduce a scenario where what is displayed in the CMS' locale switcher, differs from the locale saved in the `FluentLocale_CMS` cookie and what is displayed in the admin URL's GET param represented by `Fluent::query_param`.

For step-by-step reproduction, see comment no's 20-22 in #191.

phptek:
Allowing locale switcher to maintain correct highlighted default when using browser navigation.

### DIFF
--- a/javascript/fluent.js
+++ b/javascript/fluent.js
@@ -173,6 +173,17 @@
 				}
 				input.change();
 			}
-		})
+		});
+
+		/**
+		 * Maintain UI state of the CMS locale switcher when using standard browser navigation buttons.
+		 */
+		$('.cms-edit-form').entwine({
+			redraw: function() {
+				var localeMenuLocale = $.cookie('FluentLocale_CMS')
+				$('.cms-fluent-selector .text').text(fluentLocales[localeMenuLocale]);
+				this._super();
+			}
+		});
 	});
 })(jQuery);


javascript/fluent.js
```diff
@@ -173,6 +173,17 @@
...
 			}
-		})
+		});
+
+		/**
+		 * Maintain UI state of the CMS locale switcher when using standard browser navigation buttons.
+		 */
+		$('.cms-edit-form').entwine({
+			redraw: function() {
+				var localeMenuLocale = $.cookie('FluentLocale_CMS')
```
tractorcow: This string is soft-coded in https://github.com/tractorcow/silverstripe-fluent/blob/master/_config/fluent.yml#L44

Can you make sure the fluent config includes the cookie name, then soft-coding it here?

phptek: > This string is soft-coded in https://github.com/tractorcow/silverstripe-fluent/blob/master/_config/fluent.yml#L44

Indeed it is sir, but I did notice that it's already baked-in on line 103 thus:

    newUrl = urlForLocale(url, $.cookie('FluentLocale_CMS'));

...want me to change that too?

tractorcow: Sure :D

tractorcow:
I think that this improves the overall value of the module, even if the cookie key is hard-coded.

Feel free to soft-code the key in another PR if you like. :D

phptek:
> Feel free to soft-code the key in another PR if you like. :D
Friday :-)

tractorcow:
Yeah, I can't be too demanding can I? :P

PR #215, opened 2016-06-20
UPDATE: Fluent.FilterAdmin is now taken from config

tony13tv:
I removed the variable (const FILTER_ADMIN) as it wasn't fitting the regular SS programming style. Also the class is taking Fluent.FilterAdmin key from config file and adding its value to the queryParams as this wasn't actually working.

### DIFF
--- a/code/extensions/FluentFilteredExtension.php
+++ b/code/extensions/FluentFilteredExtension.php
@@ -9,6 +9,13 @@
 class FluentFilteredExtension extends DataExtension
 {
     /**
+     * Data query key necessary to turn on admin filtering
+     *
+     * @var string
+     */
+    const FILTER_ADMIN = 'Fluent.FilterAdmin';
+
+    /**
      * Set the filter of locales to the specified locale, or array of locales
      *
      * @param string|array $locale Locale, or list of locales
--- a/code/extensions/FluentFilteredExtension.php
+++ b/code/extensions/FluentFilteredExtension.php
@@ -99,14 +99,14 @@ class FluentFilteredExtension extends DataExtension
     }
 
     /**
-     * Amend freshly created DataQuery objects with the current locale and frontend status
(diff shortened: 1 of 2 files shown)


code/extensions/FluentFilteredExtension.php
```diff
@@ -113,6 +106,7 @@ public function updateCMSFields(FieldList $fields)
      */
     public function augmentDataQueryCreation(SQLQuery $query, DataQuery $dataQuery)
     {
+        $dataQuery->setQueryParam('Fluent.FilterAdmin', Fluent::config()->FilterAdmin);
```
tractorcow: Can you please document this in the docs, and make sure the config option follows the naming convention (`filter_admin`).

You'll need to make sure there's a default option set, which should be false, right? Add to https://github.com/tractorcow/silverstripe-fluent/blob/master/_config/fluent.yml

code/extensions/FluentFilteredExtension.php
```diff
@@ -9,13 +9,6 @@
 class FluentFilteredExtension extends DataExtension
 {
     /**
-     * Data query key necessary to turn on admin filtering
-     *
-     * @var string
-     */
```
tractorcow: I just realised something... if we remove this, it breaks backwards compatibility, since existing code could refer to it.

Can you please leave this block in? You don't have to use it, so the below code is ok.

tractorcow:
> I removed the variable (const FILTER_ADMIN) as it wasn't fitting the regular SS programming style.

Actually technically, it is in style. :)

tony13tv:
I was referring exactly to that in particular. I'm not so used to see constants in SS code. Sorry &#128517;

tractorcow:
Maybe that's because it's so in-CONST-istant? :P

tractorcow:
Good job @tony13tv :D thanks for your contribution!

PR #217, opened 2016-07-07
BUG Ensure that new pages are overwritten on first save, even if saved in non-default locale

### LINKED ISSUE #214: "New Page" vs "What the docs say"
**Setup:**
- PHP 5.6
- Framework + CMS 3.4 (Non CWP)
- Fluent 3.4.0

**fluent.yml:**

``` yml

---
Name: ss34fluentconfig
After: '#fluentconfig'

---
Fluent:
  default_locale: en_NZ
  locales:
    - en_NZ
    - fr_FR
  aliases:
    en_NZ: en
    fr_FR: fr
  field_include:
    - Title
    - MenuTitle
    - Content
  detect_locale: false
  force_domain: false

---
Name: ss34fluenti18nconfig
After: '#fluenti18nconfig'

---
i18n:
  default_locale: en_NZ
```

Re-reading the docs here: https://github.com/tractorcow/silverstripe-fluent/blob/master/docs/en/how-fluent-works.md#creating-translations

This needs to be made clearer:

> It's ideal, but not always necessary, to create a record in the default locale.

The problem is that it's not obvious what _should happen_ when a record is indeed created in a non-default locale.

Having tested Fluent 3.4.0 on SS 3.4, what happens for SiteTree objects created at the top level of the CMS' site tree, is that "New Page" appears for MenuTitle in the frontend main menu for the "EN" locale, and the expected French MenuTitle when in the "FR" locale.

IMO this should be documented perhaps with some tips as to how to work around it i.e. using `FluentFilteredLocale` or even an `augmentSQL()` workaround that checks for `MenuTitle === _t('CMSMain.NEWPAGE')`.

> If a record is NOT created in the default locale then the default value of that record will be the value last saved in any locale.

I'm either misunderstanding what this really means and its ramifications or it needs to be made clearer or there's a bug.

Given the scenario above, where "New Page" shows when I create a page in a non-default locale, then the documentation above is not accurate. What's _actually_ happening is that the default value for this record stays as the CMS' default "New Page" and _only_ the field(s) saved for the non-default locale, contain the expected content.

tractorcow:
Fixes #214

Just in case @phptek thinks I've stopped maintaining this module. :)

### DIFF
--- a/code/extensions/FluentExtension.php
+++ b/code/extensions/FluentExtension.php
@@ -124,6 +124,30 @@ class FluentExtension extends DataExtension
     }
 
     /**
+     * Copies current value of this record to the values in the given locale.
+     *
+     * Useful when writing a record to multiple locales in a single write.
+     *
+     * Note: Won't have any effect when writing to the current locale, since that
+     * locale will be written to anyway.
+     *
+     * @param string $locale
+     */
+    public function copyFieldsToLocale($locale) {
+        // Iterate over all localised fields in all tables
+        $includedTables = $this->getTranslatedTables();
+        foreach ($includedTables as $class => $fields) {
+            foreach ($fields as $field) {
+                // Copy to this locale
+                // (Title => Title_fr_FR)
+                $value = $this->owner->getField($field);
+                $updateField = Fluent::db_field_for_locale($field, $locale);
+                $this->owner->setField($updateField, $value);
+            }
+        }
+    }
+
+    /**
      * Splits a spec string safely, considering quoted columns, whitespace,
      * and cleaning brackets
      *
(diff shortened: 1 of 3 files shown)


tractorcow:
@phptek I've given you collab permissions; Please merge this if you like it. :)

It'd be good if we could work together on supporting fluent, so we can peer review each other's code. :P

PS: don't merge your own pull requests without review. :P I'll try to make sure I am around to review future changes.

phptek:
Ooh tasty. I'll try and take a look Friday ( on leave atm pending new job on Monday) ta for merge privs, re: guidelines, got 'em :-)

tractorcow:
yeah, in fact I've had this issue before with a page that had "new zealand" as the title. There are other parts of the code base that rely on "new" as a "new page indicator".

This is the bit of code that gets the new page title.


```php
$newItem->Title = _t(
	'CMSMain.NEWPAGE',
	"New {pagetype}",'followed by a page type title',
	array('pagetype' => singleton($className)->i18n_singular_name())
);
```

I guess it's too hard to solve... at least with this solution. =(

PR #220, opened 2016-07-25
BUG Don't assume default_locale on home page if detect_locale is enabled

### LINKED ISSUE #218: Fluent::detect_browser_locale it's never being called
I tried to make my site "detect browser locale" through configuration but it didn't work.

Doing some debug I found out that there was a rule for '/' that sets the ```query_param``` to ```Fluent::default_locale(true)``` but that was breaking the ```detect_browser_locale``` feature.

So I commented it out. I want to know if I should PR this change or maybre create a config param to handle this use case.

tractorcow:
Fixes #218

@tony13tv please check this fixes your issue.

### DIFF
--- a/code/Fluent.php
+++ b/code/Fluent.php
@@ -43,8 +43,13 @@ class Fluent extends Object implements TemplateGlobalProvider
         // Home page route
         $routes[''] = array(
             'Controller' => 'FluentRootURLController',
-            self::config()->query_param => static::default_locale(true),
         );
+
+        // If we do not wish to detect the locale automatically, fix the home page route
+        // to the default locale for this domain.
+        if (!Fluent::config()->detect_locale) {
+            $routes[''][self::config()->query_param] = static::default_locale(true);
+        }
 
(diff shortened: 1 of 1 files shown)


code/Fluent.php
```diff
@@ -43,9 +43,14 @@ public static function regenerate_routes()
...
-            self::config()->query_param => static::default_locale(true),
         );
 
+        // If we do not wish to detect the locale automatically, fix the home page route
+        // to the default locale for this domain.
+        if (!Fluent::config()->detect_locale) {
+            $routes[''][self::config()->query_param] = static::default_locale(true);
+        }
+
         // If default locale doesn't have prefix, replace default route with
```
phptek: Looks good to me. @tractorcow I think there might be some benefit in having some rudimentary `FunctionalTests` for `Fluent` - not necessarily for _this_ commit, but going forward - what do you think?

tractorcow: Yes I think so... if I can get the time.

PR #225, opened 2016-09-15
API Support nullable fields

### LINKED ISSUE #156: Unable to set a translated field to an empty string.
If a user changes the field value of a "translation" to be empty, the default value gets copied over.

Example:
Fluent is configured with 3 Languages, DE, FR, IT, where DE is the default.
User edits a Page and fills in content for all 3 languages (say in the "Title" and "Content" field).

Now if the user wanted to remove the content in any of these languages (eg. selecting "Italian" and then deleting all the content inside the "Content" field), the content will then be reset to another language.

I know that this is probably due to a fallback mechanism in case a field isn't translated, but it makes for a very weird experience for CMS users (if they want the field value to be empty, they have to enter whitespace).

I guess this is a tricky problem to solve. Maybe a configuration-option to disable the fallback could be introduced? Maybe even on a "per field" level?

tractorcow:
Fixes #156

A long, long requested feature.

### DIFF
--- a/code/extensions/FluentExtension.php
+++ b/code/extensions/FluentExtension.php
@@ -8,6 +8,18 @@
  */
 class FluentExtension extends DataExtension
 {
+    // <editor-fold defaultstate="collapsed" desc="Config options">
+
+    /**
+     * Set to either a list of fields, or '*', to allow fields to be nullable.
+     * This will leave fields null instead of inheriting from default locale.
+     *
+     * @var array|string
+     */
+    private static $nullable_fields = null;
+
+    // </editor-fold>
+
     // <editor-fold defaultstate="collapsed" desc="Field helpers">
 
     /**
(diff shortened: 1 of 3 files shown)


code/extensions/FluentExtension.php
```diff
@@ -8,6 +8,18 @@
  */
 class FluentExtension extends DataExtension
 {
+    // <editor-fold defaultstate="collapsed" desc="Config options">
+
+    /**
+     * Set to either a list of fields, or '*', to allow fields to be nullable.
+     * This will leave fields null instead of inheriting from default locale.
+     *
+     * @var array|string
```
phptek: ...and `null` it would seem?

code/extensions/FluentExtension.php
```diff
@@ -493,21 +505,20 @@ public static function set_enable_write_augmentation($enabled)
      *
      * @param string $condition Condition SQL string
      * @param array $includedTables
-     * @param string $locale Locale to localise to
-     * @return string Column identifier in "table"."column" format if it exists in this condition
+     * @return array Array with items [$table, $column], or null if not found
```
phptek: So `@return array|null` then? Or return an empty array for consistency? Bloody nuills...

code/extensions/FluentExtension.php
tractorcow: Yeah I don't bother adding `|null` into phpdoc.

code/extensions/FluentExtension.php
tractorcow: I haven't standardised declaring null return types explicitly; I find it simpler to assume it's a possible response, and then just declare the non-null return types.

I might be doing it wrong though... but I could be convinced to adjust my practices. :)

tractorcow:
Ok so I spoke to @sminnee and it looks like we should be documenting |null return types explicitly. I'll do this in a later pull request.

PR #227, opened 2016-09-16
FIX Unset index on manipulation no longer causes error

### LINKED ISSUE #226: Undefined index issue on dev-master
Hey guys 

I just wanted to try a recent fix that was merged into master regarding not being able to totally clear translated fields (issue #204 ). It works.. But not without throwing the following error aswell: 

```
ERROR [Notice]: Undefined index: fields
IN POST /admin/pages/edit/EditForm
Line 723 in /Applications/MAMP/htdocs/casoefurniture/fluent/code/extensions/FluentExtension.php
``` 
The code at line 723:
```php 
                if (!array_key_exists($field, $updates['fields'])) {
                    continue;
                }
``` 

Not sure if there should be a check for the "fields" index aswell?

dhensby:
fixes #226

### DIFF
--- a/code/extensions/FluentExtension.php
+++ b/code/extensions/FluentExtension.php
@@ -720,7 +720,7 @@ class FluentExtension extends DataExtension
             foreach ($includedTables[$class] as $field) {
 
                 // Skip translated field if not updated in this request
-                if (!array_key_exists($field, $updates['fields'])) {
+                if (empty($updates['fields']) || !array_key_exists($field, $updates['fields'])) {
                     continue;
                 }
 


mleutenegger:
Just wanted to take care of my mistake, but this does the job. Just as a question: wouldn't it make sense to use either `empty()` or `array_key_exists()` for both checks, as they seem to be equivalent in that case?
And additionally, from the [PHP.net manual](http://php.net/manual/en/function.empty.php):

> That means empty() is essentially the concise equivalent to !isset($var) || $var == false.

This means, `empty()` would skip if 'fields' is set but contains `NULL` (or anything listed in the manual). So in the light of preferring  `array_key_exists()` over `isset()` wouldn't it be better to use `array_key_exists()` only?

dhensby:
I don't see that using `array_key_exists` has any advantage over the `empty` check. The `empty` check (IMO) is more concise and obvious what it's doing.

I read this code as:
> If there's nothing in the `fields` array OR the key we want isn't there, `continue`

`empty` is also more forgiving, the variable doesn't even need to be an array for the empty check to pass, where as `array_key_exists` is type sensitive so could open up edge-cases.

mleutenegger:
Agreed, thanks for clarification!

tractorcow:
Thanks!

PR #234, opened 2016-10-12
Add config setting for disabling the current locale message

robbieaverill:


dhensby:
I'd merge, but I think I'll leave @tractorcow to make the call

robbieaverill:
No problem @dhensby.

As background - I just figured that it would be sensible to be able to disable this new feature if you wanted to.

Issue #221, opened 2016-08-11
Feature: Locale page indicator


tractorcow:
When editing a page there should be a generic "locale indicator" to inform the user which locale this page will be saved in. This indicator should also tell the user if this is the default locale (or not) so that users doing initial data-entry of pages can be prompted to switch.

phptek:
Something like the "two-tone" build status SVGs included in many GIT repos might be a good start? LHS (Blue: = Current locale) RHS (Green: Default Locale) - or vice versa....although ioS users may think it's some sort of button...

tractorcow:
Yeah I think colour is a good way to get different states across. :)

PR #235, opened 2016-10-27
Fix addLocaleIndicatorMessage method to only add locale message once.

bummzack:
Fixes an issue that was introduced with #221 
When the `updateCMSFields` extension hook is called multiple times (eg. once in a base-class and once in a superclass), the message would be added for every `extend` call.

This PR introduces a check that ensures that the message only gets added once.
Also made the method API more consistent.

### DIFF
--- a/code/extensions/FluentExtension.php
+++ b/code/extensions/FluentExtension.php
@@ -821,11 +821,16 @@ class FluentExtension extends DataExtension
      * Adds a UI message to indicate whether you're editing in the default locale or not
      *
      * @param  FieldList $fields
-     * @return self
+     * @return $this
      */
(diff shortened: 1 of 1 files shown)


tractorcow:
Thanks :D

robbieaverill:
I noticed this the other day too when I applied FluentExtension to Page, forgetting that it's applied to SiteTree already.

It's worth noting from the documentation:

>In all cases, it is necessary to ensure that the `updateCMSFields` extension method is called once (and only once) on each `FieldList` object.

Perhaps it's just making a configuration mistake more obvious!

Anyway, worth fixing :)

PR #237, opened 2016-11-01
Add ability to force a return value for "is_frontend"

### LINKED ISSUE #236: Enhancement: Ability to force "is_frontend"
In the following scenarios it would be useful to be able to force Fluent to return a certain value:

* Unit testing
* Web API modules
* Accessing DataObject data programmatically

Currently `Fluent::is_frontend` has support for an option for to ignore the controller, which inspects the URL instead. This would probably solve the "web API modules" example in the list above, and obviously the unit testing too (since it's covered already!) but leaves open the ability to control this setting when accessing data programmatically.

If I have three locales and three Pages, and each Page is filtered to only display in a certain locale then I have no ability to retrieve a collection of Pages for that locale automatically (as in, without adding the `->canViewInLocale()` check to each Page. The frontend detects it automatically, so would be nice to be able to do it from code too.

robbieaverill:
Resolves #236

### DIFF
--- a/code/Fluent.php
+++ b/code/Fluent.php
@@ -9,6 +9,13 @@
 class Fluent extends Object implements TemplateGlobalProvider
 {
     /**
+     * If set, "is_frontend" can be forced to return a value
+     *
+     * @var boolean
+     */
+    protected static $force_is_frontend;
+
+    /**
      * Forces regeneration of all locale routes
      */
     public static function regenerate_routes()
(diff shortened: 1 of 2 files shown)


dhensby:
ping @tractorcow

tractorcow:
Sounds good to me :D

robbieaverill:
Thanks!

PR #239, opened 2016-11-16
Save null instead of the string literal when resetting to a default value from another locale

### LINKED ISSUE #238: Improvement: Save default values as null when resetting to default
If I change a value in a non-default locale and save a Page, for example the Title and save it, Things are saved as expected:

```plain
mysql> SELECT Title, Title_en_NZ, Title_en_US FROM SiteTree WHERE ID = 1; 
+----------+-------------+-------------------+
| Title    | Title_en_NZ | Title_en_US       |
+----------+-------------+-------------------+
| Homepage | NULL        | My American Title |
+----------+-------------+-------------------+
1 row in set (0.00 sec)
```

If I then click on the locale indicator to reset it to its default value and save it again, the value is saved as the default value but as the string literal rather than acting as a transparent fallback as it would if null:

```plain
mysql> SELECT Title, Title_en_NZ, Title_en_US FROM SiteTree WHERE ID = 1;
+----------+-------------+-------------+
| Title    | Title_en_NZ | Title_en_US |
+----------+-------------+-------------+
| Homepage | Homepage    | Homepage    |
+----------+-------------+-------------+
```

This means if I go through and change the default locale's value it doesn't reflect to the non-default locale:

```plain
mysql> SELECT Title, Title_en_NZ, Title_en_US FROM SiteTree WHERE ID = 1;
+------------------+------------------+-------------+
| Title            | Title_en_NZ      | Title_en_US |
+------------------+------------------+-------------+
| Homepage changed | Homepage changed | Homepage    |
+------------------+------------------+-------------+
```

## Improvement request

When resetting to a default value we should set the localised field to null to allow it to be transparent. This would match the expectation for when you first created the page and default changes would be reflected everywhere.

robbieaverill:
This ensures that underlying changes at the default locale will still come through at non-standard locales, as they would when you first create a DataObject.

Resolves #238

### DIFF
--- a/code/extensions/FluentExtension.php
+++ b/code/extensions/FluentExtension.php
@@ -740,6 +740,15 @@ class FluentExtension extends DataExtension
                     if (!empty($updates['fields'][$defaultField]) || $this->isFieldNullable($field)) {
                         $updates['fields'][$field] = $updates['fields'][$defaultField];
                     }
+
+
+                    // If the localised field has been reset to the default value (i.e. they're the same value)
+                    // we should store it as null to allow change transparency from the default locale in future.
+                    if ((!empty($updates['fields'][$defaultField]))
+                        && ($updates['fields'][$updateField] === $updates['fields'][$defaultField])
+                    ) {
+                        $updates['fields'][$updateField] = null;
+                    }
                 }
             }
 
--- a/code/extensions/FluentExtension.php
+++ b/code/extensions/FluentExtension.php
@@ -745,7 +745,7 @@ class FluentExtension extends DataExtension
                     // If the localised field has been reset to the default value (i.e. they're the same value)
                     // we should store it as null to allow change transparency from the default locale in future.
                     if ((!empty($updates['fields'][$defaultField]))
-                        && (!$this->isFieldNullable($updateField))
+                        && (!$this->isFieldNullable($defaultField))
                         && ($updates['fields'][$updateField] === $updates['fields'][$defaultField])
                     ) {
                         $updates['fields'][$updateField] = null;
@@ -831,7 +831,7 @@ class FluentExtension extends DataExtension
      * Adds a UI message to indicate whether you're editing in the default locale or not
      *
      * @param  FieldList $fields
-     * @return self
+     * @return $this
      */
     protected function addLocaleIndicatorMessage(FieldList $fields)
     {
--- a/code/extensions/FluentExtension.php
+++ b/code/extensions/FluentExtension.php
@@ -745,6 +745,7 @@ class FluentExtension extends DataExtension
                     // If the localised field has been reset to the default value (i.e. they're the same value)
                     // we should store it as null to allow change transparency from the default locale in future.
                     if ((!empty($updates['fields'][$defaultField]))
+                        && (!$this->isFieldNullable($updateField))
                         && ($updates['fields'][$updateField] === $updates['fields'][$defaultField])
                     ) {
                         $updates['fields'][$updateField] = null;
@@ -830,16 +831,14 @@ class FluentExtension extends DataExtension
      * Adds a UI message to indicate whether you're editing in the default locale or not
      *
      * @param  FieldList $fields
-     * @return $this
+     * @return self
      */
     protected function addLocaleIndicatorMessage(FieldList $fields)
     {
-        if (Fluent::config()->disable_current_locale_message) {
-            return $this;
-        }
-
-        // If the field is already present, don't add it a second time
-        if ($fields->fieldByName('CurrentLocaleMessage')) {
+        if ((Fluent::config()->disable_current_locale_message)
+            // If the field is already present, don't add it a second time
+            || ($fields->fieldByName('CurrentLocaleMessage'))
+        ) {
             return $this;
         }
 
--- a/tests/FluentTest.php
+++ b/tests/FluentTest.php
@@ -1142,6 +1142,30 @@ class FluentTest extends SapphireTest
         // Different from default locale
         $this->assertTrue(Fluent::isFieldModified($object, $fields->fieldByName('Root.Main.Title'), 'en_NZ'));
     }
+
+    /**
+     * When editing in a non-default locale and resetting the value for a field to the default value, or entering
+     * the same value as the default value, we should store null instead of the string literal to allow for transparent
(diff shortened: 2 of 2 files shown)


robbieaverill:
I'm not sure if this is the best way to achieve it @tractorcow, or whether it perhaps goes against something that you want to do deliberately - but I'll just leave this here for you :)

robbieaverill:
Yep - Travis is telling me that it's not the expected path, I'll close this for now.

tractorcow:
Hi @robbieaverill , actually I agree with your change. Please see my note above however regarding a risk with nullable fields.

The tests are correctly pointing out a change in behaviour, but I would be willing to accept this change for a new minor version release.

tractorcow:
Do you want to fix the nullable issue and demonstrate what changes are necessary to the tests to get them working? If so I'll review.

robbieaverill:
Thanks @tractorcow - I've updated the tests and relevant comments to show the behaviour changes, and impleleted a `FluentExtension::isFieldNullable` check before setting a null value. Over to you.

code/extensions/FluentExtension.php
```diff
@@ -740,6 +740,16 @@ public function augmentWrite(&$manipulation)
                     if (!empty($updates['fields'][$defaultField]) || $this->isFieldNullable($field)) {
                         $updates['fields'][$field] = $updates['fields'][$defaultField];
                     }
+
+
+                    // If the localised field has been reset to the default value (i.e. they're the same value)
+                    // we should store it as null to allow change transparency from the default locale in future.
+                    if ((!empty($updates['fields'][$defaultField]))
+                        && (!$this->isFieldNullable($updateField))
```
tractorcow: isFieldNullable() takes the untranslated $field, rather than the $updateField.

code/extensions/FluentExtension.php
```diff
@@ -821,16 +831,14 @@ public function updateCMSFields(FieldList $fields)
      * Adds a UI message to indicate whether you're editing in the default locale or not
      *
      * @param  FieldList $fields
-     * @return $this
+     * @return self
```
tractorcow: $this is correct. :) self means "An instance of the same type".

robbieaverill:
Can you link me a PHPDoc reference for that? I'm 80% sure that $this is not
the norm, but happy to proved wrong!

tractorcow:
Very very bottom of https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md

robbieaverill:
Ah, right. I normally stick to the de facto standard. Will update it in an
hour or so though. Thanks!

tractorcow:
@robbieaverill I'm being super mean by using standards that haven't yet been ratified and voted on, but they are close as we have to a convention. :)

tractorcow:
Once I've reviewed the tests and satisfied that the change in behaviour for storage is acceptable, I'll merge.

tests/FluentTest.php
```diff
@@ -1142,6 +1143,30 @@ public function testIsFieldModified()
...
+        // Give it the same value as the default value
+        Fluent::with_locale('en_US', function () use ($objectId) {
+            $record = FluentTest_TranslatedObject::get()->byId($objectId);
+            $record->Title = 'Cette couleur est le bleu'; // fr_CA is the default locale in this test suite
+            $record->write();
+        });
+
+        // Ensure the field is now null
+        $object = FluentTest_TranslatedObject::get()->byId($objectId);
+        $this->assertNull($object->Title_en_US);
```
tractorcow: 👍  for the test.

tests/FluentTest.php
```diff
@@ -913,12 +913,12 @@ public function testWriteWithLocale()
...
         $item2->write();
 
-        // check alternate locale title unchanged
+        // check alternate locale title reflects the default locale value
         $es2Title = Fluent::with_locale('es_ES', function () use ($item2ID) {
             $item2 = FluentTest_TranslatedObject::get()->byId($item2ID);
             return $item2->Title;
         });
-        $this->assertEquals($es2Title, 'Spanish 2');
+        $this->assertEquals($es2Title, 'English 2');
```
tractorcow: Hm, this raises an interesting question.

Create record in spanish locale, save.
Go to english locale, modify.
Going back to spanish locale, and this version has reverted to the english locale (because it was null and thus override it).

You agree @robbieaverill that this is bad behaviour right? Is there a way we can prevent this?

tractorcow: I would say, maybe we want to disable this behaviour until AFTER the record has been saved at least once in the default locale, right?

robbieaverill: Yea, you're right and I completely agree

tractorcow:
How about we modify writes as below:

# Use case 1

Initial write in default locale:

Field: "initial"
Field_default_locale: "initial"
Field_other_locale: null

Save in non-default value with new value ("second")

Field: "initial"
Field_default_locale: "initial"
Field_other_locale: "second"

Save in non-default value with same value ("initial")

Field: "initial"
Field_default_locale: "initial"
Field_other_locale: null (because Field_default_locale is set)

# Use case 2

Initial write in non-default locale:

Field: "initial"
Field_default_locale: null
Field_other_locale: "initial" (non-null because Field_default_locale is null)

Subsequent write in default locale with new value ("second")

Field: "second"
Field_default_locale: "second"
Field_other_locale: "initial"

Save in non-default value with same ("second") value

Field: "second"
Field_default_locale: "second"
Field_other_locale: null (null because Field_default_locale is non-null)

robbieaverill:
LGTM @tractorcow - so if you create a page in Spanish (non-default), then the default locale (say, en_NZ) would be null and thus inherit the Spanish version by default. That's desired behaviour I suppose, because you don't necessarily need to publish the page in English..

tractorcow:
I guess I'll have to work on this at some time... it's getting complicated. ;)

I might need to write some architectural docs to document all of this.

robbieaverill:
If you want to @tractorcow then go for it. This PR has "allow edits from maintainers" on, so you should be able to go straight into it.

I'm happy to make the requested changes as well, but I'm also quite busy at the moment with packing/moving cities!

robbieaverill:
Closing for now - will need to be re-evaluated.

PR #254, opened 2017-03-26
Fix virtualpage MenuTitle

### LINKED ISSUE #253: virtual pages navigation label not set
With fluent installed, it seems that the 'Navigation label'  in virtual pages does NOT get properly set (it stays on New Virtual Page). The workaround is to 're-publish' the original page which will then cause the navigation labels of the virtual pages to be re-written ... but only in the current language ... so the master page has to be re-published in each language separately ...

tractorcow:
Fixes #253


oetiker:
great! works fine!

PR #258, opened 2017-05-15
prevent cli bug

bhofstaetter:
HTTP_HOST is not present if you want to run something at the console

### DIFF
--- a/code/Fluent.php
+++ b/code/Fluent.php
@@ -246,7 +246,7 @@ class Fluent extends Object implements TemplateGlobalProvider
      */
     public static function locales($domain = null)
     {
-        if ($domain === true) {
+        if (!Director::is_cli() && $domain === true) {
             $domain = strtolower($_SERVER['HTTP_HOST']);
         }
 
--- a/code/Fluent.php
+++ b/code/Fluent.php
@@ -378,7 +378,7 @@ class Fluent extends Object implements TemplateGlobalProvider
      */
     public static function default_locale($domain = null)
     {
-        if ($domain === true) {
+        if (!Director::is_cli() && $domain === true) {
             $domain = strtolower($_SERVER['HTTP_HOST']);
         }
 


phptek:
@bhofstaetter How have you got your environment setup? Using `_ss_environment.php` correctly with the `$_FILE_TO_URL_MAPPING[]` array set correctly, should preclude this from being needed. It is used for exactly this purpose - lack of HTTP_HOST in CLI environments.

code/Fluent.php
```diff
@@ -378,7 +378,7 @@ public static function domain_for_locale($locale)
      */
     public static function default_locale($domain = null)
     {
-        if ($domain === true) {
+        if (!Director::is_cli() && $domain === true) {
             $domain = strtolower($_SERVER['HTTP_HOST']);
```
tractorcow: Instead of the is_cli() check, please use:

```php
$protocolAndHost = Director::protocolAndHost();
$protocol = Director::protocol();
$domain = substr($protocolAndHost, 0, strlen($protocol));
```

In ss 4.0 and above you can just use `$domain = Director::host()` but that's not yet implemented.
 
 `Director::protocolAndHost()` and remove the protocol.

tractorcow: Eh, I should really back-port that to 3.x. It's annoying you can't do that easily.

bhofstaetter: as soon as I put `Director::protocolAndHost();` in, I once again receive ` HTTP_HOST not set.  You probably want to define an entry in $_FILE_TO_URL_MAPPING`

I'm quite sure file to url mapping is set correctly and I also double checked it. If it would be wrong, all my custom tasks would also break

dhensby:
> I'm quite sure file to url mapping is set correctly and I also double checked it. If it would be wrong, all my custom tasks would also break

You wouldn't get this error if it were, tbh. unless you have some really unique setup that we haven't seen before

PR #263, opened 2017-07-31
Create a Finnish translation

Taitava:
This is a very quick translation and not the best possible quality, but at least it's something :). If anybody wants to make any changes to any of the translated terms, please feel free to do it! :)


tractorcow:
Looks good! thanks for the localisation. :)

PR #264, opened 2017-08-07
API Add InitStateMiddleware and LocaleDetectorMiddleware as part of routing framework

robbieaverill:
Contents of HandleRedirectMiddleware is still to come.


src/State/LocaleDetector.php
```diff
@@ -0,0 +1,271 @@
...
+        }
+
+        $hostname = strtolower(parse_url($this->getRequest()->getHeader('Referer'), PHP_URL_HOST));
+        // Check if internal traffic
+        if ($hostname == strtolower($this->getRequest()->getHeader('Host'))) {
+            return true;
+        }
+
+        // Check configured domains
+        $domains = static::config()->get('domains');
```
robbieaverill: This needs to be replaced with the domain model

src/Model/Locale.php
```diff
@@ -42,6 +43,7 @@ class Locale extends DataObject
     private static $db = [
         'Locale' => 'Varchar(10)',
         'Title' => 'Varchar(100)',
+        'Alias' => 'Varchar(20)',
```
tractorcow: `URLSegment` is the new `Alias`, so we don't need this anymore.

src/Middleware/InitStateMiddleware.php
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /**
+     * Gets the current domain from the request
+     *
+     * @return string
+     */
+    public function getRequestDomain()
+    {
+        return strtolower((string) $this->getRequest()->getHeader('Host'));
```
tractorcow: Can we use Director::host()? It'll be easier for supporting testing.

src/Middleware/InitStateMiddleware.php
```diff
@@ -0,0 +1,142 @@
...
+    protected $request;
+
+    /**
+     * URL paths that should be considered as admin only, i.e. not frontend
+     *
+     * @config
+     * @var array
+     */
+    private static $admin_url_paths = [
+        'dev/',
```
tractorcow: Don't forget `admin/` :P

src/Middleware/InitStateMiddleware.php
```diff
@@ -0,0 +1,142 @@
...
+ * InitStateMiddleware initialises the FluentState object and sets the current request locale and domain to it
+ */
+class InitStateMiddleware implements HTTPMiddleware
+{
+    use Configurable;
+
+    /**
+     * @var HTTPRequest
+     */
+    protected $request;
```
tractorcow: Instead of making this a state, let's make the middleware stateless, and pass in $request argument to each method. That way the middleware is more re-usable across requests (i.e. testing).

src/Middleware/InitStateMiddleware.php
```diff
@@ -0,0 +1,142 @@
...
+    public function process(HTTPRequest $request, callable $delegate)
+    {
+        $this->setRequest($request);
+
+        $state = FluentState::create();
+        if ($locale = $this->getRequestLocale()) {
+            $state->setLocale($locale);
+        }
+
+        if ($domain = $this->getRequestDomain()) {
```
tractorcow: `$domain = Director::host($request)`

src/Middleware/InitStateMiddleware.php
robbieaverill: I had it that way, but thought this looked better. Happy to revert

src/Middleware/InitStateMiddleware.php
robbieaverill: It's added in the method that uses it since it's dynamic. I could add it as a string here anyway and filter duplicates?

tractorcow: Oh, that's ok, leave it as is I think.

PR #265, opened 2017-08-07
API Add DetectLocaleMiddleware to detect and set persisted locale information

robbieaverill:
This replaces the `HandleRedirectsMiddleware`. I'll make a follow up pull request (or push another commit to this branch) to handle root URL redirects if they're configured.

I've removed the "persist locale" data from the FluentState because I think we can wrap it up within the "locale" data in the state instead. If we disagree it can be added back in, but should differentiate between the CMS and frontend persistence keys.


src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * Whether cookies have already been set during {@link setPersistLocale()}
+     *
+     * @var bool
+     */
+    protected $cookiesPersisted = false;
+
+    public function process(HTTPRequest $request, callable $delegate)
+    {
+        $localeDetector = LocaleDetector::singleton()->setRequest($request);
```
tractorcow: I suggest we make these services stateless; Instead of setRequest, we pass this in via each method that requires it.

src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        // Fallback to default if empty or invalid (for this domain)
+        if (empty($locale) || !Locale::getByLocale($locale)) {
+            // If on the frontend, filter locales by the current domain
+            $domain = $state->getIsFrontend($request) ? $state->getDomain() : null;
+
+            /** @var Locale $locale */
+            if ($locale = Locale::getDefault($domain)) {
+                $locale = $locale->Locale;
```
tractorcow: Re-assignment of different types to one variable is code smell. Let's separate locale object / strings into different variables.

src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Get the current locale from routing parameters, persistence, browser locale, etc
+     *
+     * @param  HTTPRequest $request
+     * @return string
+     */
+    public function getLocale(HTTPRequest $request)
+    {
+        $state = FluentState::singleton();
```
tractorcow: More of a question than a comment; If FluentState already has a locale set, what should this middleware do? I'm thinking about how this middleware will interact with the other middlewares (E.g. detect locale via querystring / url args).

tractorcow:
> I've removed the "persist locale" data from the FluentState because I think we can wrap it up within the "locale" data in the state instead. If we disagree it can be added back in, but should differentiate between the CMS and frontend persistence keys.

I agree with this, but I'll let you answer my question above and I'll have a think about this. :)

src/Middleware/DetectLocaleMiddleware.php
robbieaverill: I'm actually tempted to remove that class now and move all its logic to this middleware. Anything that needs to be "got" can be set to the state for user code to use.

src/Middleware/DetectLocaleMiddleware.php
robbieaverill: Do you think it might be useful to separate these middlewares further, so we have one for each: DetectQueryStringLocaleMiddleware, DetectUrlArgumentLocaleMiddleware and DetectPersistedLocaleMiddleware?

src/Middleware/DetectLocaleMiddleware.php
robbieaverill: For now though I'll update to be stateless

src/Middleware/DetectLocaleMiddleware.php
robbieaverill: To answer your question - this particular middleware should probably ignore the locale if it's already set, I'll add that in

PR #267, opened 2017-08-08
NEW Ensure locale detection priority is honoured for homepage and non-homepage routes

robbieaverill:
Order should be 1) route param, 2) request param, 3) persisted locale and if home page then 4) language header.

### DIFF
--- a/src/Middleware/DetectLocaleMiddleware.php
+++ b/src/Middleware/DetectLocaleMiddleware.php
@@ -84,8 +84,8 @@ class DetectLocaleMiddleware implements HTTPMiddleware
             $locale = $this->getPersistLocale($request);
         }
 
-        // Check for locale in browser headers
-        if (empty($locale)) {
+        // Home page only: check for locale in browser headers
+        if (empty($locale) && $request->getURL() === '') {
             $locale = LocaleDetector::singleton()->detectBrowserLocale($request);
         }
 
--- a/src/Middleware/DetectLocaleMiddleware.php
+++ b/src/Middleware/DetectLocaleMiddleware.php
@@ -59,8 +59,10 @@ class DetectLocaleMiddleware implements HTTPMiddleware
             $locale = $this->getLocale($request);
 
             $state->setLocale($locale);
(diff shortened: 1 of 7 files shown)


src/State/LocaleDetector.php
```diff
@@ -105,9 +105,10 @@ public function detectBrowserLocale(HTTPRequest $request, $currentDomain = false
         // Check each requested locale against loaded locales
         foreach ($prioritisedLocales as $priority => $parsedLocales) {
             foreach ($parsedLocales as $browserLocale) {
-                foreach ($this->getLocales($currentDomain) as $locale) {
-                    if (stripos(preg_replace('/_/', '-', $locale), $browserLocale) === 0) {
-                        return $locale;
+                foreach ($this->getLocales($currentDomain) as $localeObj) {
+                    /** @var Locale $localeObj */
+                    if (stripos(preg_replace('/_/', '-', $localeObj->Locale), $browserLocale) === 0) {
```
tractorcow: Can we put this preg_replace into a separate getter on Locale class? (After I merge this)

src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -84,8 +84,8 @@ public function getLocale(HTTPRequest $request)
             $locale = $this->getPersistLocale($request);
         }
 
-        // Check for locale in browser headers
-        if (empty($locale)) {
+        // Home page only: check for locale in browser headers
+        if (empty($locale) && $request->getURL() === '') {
```
tractorcow: Commenting on wrong line, but still in this file;

Shall we move $this->setPersistLocale($request, $locale); to outside of the loop in process? That way we capture the current locale more frequently, in case the next request doesn't have a safe way to guess the locale. (e.g. on immediate redirect).

robbieaverill: Makes sense. Users wanting to implement their own locale detection middleware would need to ensure it's called before this middleware, but after the InitStateMiddleware

PR #269, opened 2017-08-18
API Lots of refactoring of develop branch

tractorcow:
API Added flush to CacheableModel
Default locale is now stored in _Localised subtable instead of core table
API Updated default locale detection
API Updated domain behaviour
Removed unused redirection code from LocaleDetector, and made this an interface
API Replaced DefaultProvider with extension points in FluentExtension
Added new icons
BUG Always create localised table for base class, even if no fields are localised
API Ensure un-localised pages are hidden on the frontend
API Removed $key param from set/getPersistedLocale (redundant)
ENHANCEMENT If navigating a domain with only one locale, fix to that locale
BUG Fix inability to dev/build if database isn’t built properly already

For the attention of @robbieaverill :)


src/Extension/FluentExtension.php
```diff
@@ -278,10 +263,18 @@ public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null)
...
         }
 
-        // Get default provider
-        $providerName = Config::inst()->get(get_class($this->owner), 'defaultProvider');
-        /** @var DefaultProvider $defaultProvider */
-        $defaultProvider = $providerName ? Injector::inst()->get($providerName) : null;
+        // On frontend, ensure at least one localised version exists in localised base table (404 otherwise)
+        if (FluentState::singleton()->getIsFrontend()) {
+            $wheres = [];
+            $baseTableLocalised = $this->owner->baseTable() . '_' . self::SUFFIX;
```
robbieaverill: I think this might be useful to make into its own method, i.e. `$somethingFluent->getLocalisedTable()`. I can imagine user code taking advantage of this

src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -172,15 +167,101 @@ public function setPersistLocale(HTTPRequest $request, $locale, $key = null)
...
+            return null;
+        }
+        // Respect config disable
+        if (!FluentDirectorExtension::config()->get('detect_locale')) {
+            return null;
+        }
+
+        /** @var LocaleDetector $detector */
+        $detector = Injector::inst()->get(LocaleDetector::class);
+        $localeObj = $detector->detectLocale($request);
```
robbieaverill: Yay, this class finally has a clear purpose :D

src/Extension/FluentExtension.php
tractorcow: Yes I think so.

src/Model/Locale.php
```diff
@@ -220,4 +224,38 @@ public function isLocale($locale)
...
+        }
+
+        return $locales;
+    }
+
+    public function onAfterWrite()
+    {
+        parent::onAfterWrite();
+
+        // If this is the default locale, remove default from other locases
```
robbieaverill: Typo

src/Model/CachableModel.php
```diff
@@ -26,6 +28,11 @@ public static function getCached()
             return $list;
         }
 
+        // Check if db is ready for this model to be queried
+        if (!static::databaseIsready()) {
```
robbieaverill: I know PHP isn't case sensitive, but should be `databaseIsReady`

src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -172,15 +167,101 @@ public function setPersistLocale(HTTPRequest $request, $locale, $key = null)
     /**
      * Get the Fluent locale persistence key. See the "persist_ids" config static.
      *
-     * @param  string|null $key
      * @return string
      */
-    public function getPersistKey($key = null)
+    public function getPersistKey()
```
robbieaverill: Seeing as the methods below are now protected, do you think we should make these other methods protected as well and just leave `process` as public? We can probably just have integration tests for this class instead of smaller units

tractorcow: Yes I think so.

src/Extension/FluentExtension.php
```diff
@@ -316,6 +321,20 @@ public function augmentDataQueryCreation(SQLSelect $query, DataQuery $dataQuery)
     }
 
     /**
+     * Get the localised table name
+     *
+     * @return string
+     */
+    public function getLocalisedTable()
```
tractorcow: How about moving this to a method that takes:
- normal table
- optional locale (for alias)

That way we can use it for the non-base tables also.

src/Extension/FluentExtension.php
```diff
@@ -316,6 +321,20 @@ public function augmentDataQueryCreation(SQLSelect $query, DataQuery $dataQuery)
...
+    {
+        $tableName = $this->owner->baseTable() . '_' . self::SUFFIX;
+
+        $this->owner->extend('updateLocalisedTable', $tableName);
```
tractorcow: I'm not keen on adding extension points unless we had a solid use case for these.

PR #270, opened 2017-08-20
Rename codecov YAML configuration

robbieaverill:
Codecov.io docs say they support both filenames, but since it's failing to be detected their support have suggested renaming it.


robbieaverill:
Yay!

PR #271, opened 2017-08-25
NEW Add template accessors and extension helper methods for getting locale data

robbieaverill:


dhensby:
looks like this needs a rebase

robbieaverill:
Thanks, updated

src/Extension/FluentExtension.php
```diff
@@ -458,4 +458,40 @@ protected function getDataQueryLocale(DataQuery $dataQuery = null)
...
+            return $localeObj ? $localeObj->getBaseURL() : '';
+        }
+
+        // Mock locale and recalculate link
+        return FluentState::singleton()->withState(function ($newState) use ($locale) {
+            $newState->setLocale($locale);
+
+            $link = DataObject::get($this->owner->ClassName)->byID($this->owner->ID)->Link();
+
+            // Prefix with domain if available
```
tractorcow: I'll have a look at this logic; We may need to check if we are in domain mode, first.

Alternatively this should probable be baked into the updateLink() extension method.

src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -186,8 +186,12 @@ protected function getPersistKey()
     protected function getParamLocale(HTTPRequest $request)
     {
         $queryParam = FluentDirectorExtension::config()->get('query_param');
-        $locale = (string)$request->param($queryParam) ?: $request->requestVar($queryParam);
-        return $locale;
+        if (FluentDirectorExtension::config()->get('detect_locale')
```
tractorcow: detect_locale is meant to be a flag for "make the guess guess if we don't have an explicit locale". In this case a query param is explicit so we don't need to treat it as though we would an ambiguous check.

src/Model/Locale.php
```diff
@@ -296,4 +301,47 @@ public function getChain()
...
+     * Determine the base URL within the current locale
+     *
+     * @return string
+     */
+    public function getBaseURL()
+    {
+        $base = Director::baseURL();
+
+        // Build domain-specific base URL
+        if ($this->Domain() && $this->Domain()->exists()) {
```
tractorcow: We should only do this in domain mode. Even if a domain is configured (but not in domain mode) we should default to normal locale-prefix baseurl.

src/Extension/FluentExtension.php
```diff
@@ -511,4 +514,63 @@ protected function getManipulationRecordID($updates)
...
     }
+
+    /**
+     * Add the current locale's URL segment to the start of the URL
+     *
+     * @param  string &$base
+     * @param  string &$action
+     * @return null
+     */
+    public function updateRelativeLink(&$base, &$action)
```
tractorcow: We should move this to FluentSiteTreeExtension, since it's sitetree specific.

tractorcow: Also, we can use `updateLink` now, if you want to merge the domain with relativelink there. :)

tractorcow:
I'm going to merge and make the changes in another PR.

PR #277, opened 2017-09-15
Update templates and shift to using inline-editable

tractorcow:
Depends on https://github.com/symbiote/silverstripe-gridfieldextensions/pull/213 (merge gridfield PR first)


tractorcow:
Rebased.

PR #289, opened 2017-10-17
Update travis install steps

dhensby:
I wasn't sure which branch to do this against, but develop and master are both affected


robbieaverill:
Weird, I thought I'd deleted the develop branch =/ we should be using master from now on, will update this for you

robbieaverill:
Thanks @dhensby

PR #299, opened 2017-11-07
Add new config to determine frontend visibility of unpublished locales

chrispenny:
This addition makes the assumption that Fluent (by default) wants unpublished locales to inherit and display data based on its "Fallback", and for that data to be visible on the frontend (the same as current CMS/backend functionality). Unpublished locales are only hidden on the frontend when/if this config variable is set on the `owner` Object.

I would be more than happy to flip this assumption to hide on frontend by default if that is preferred.

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -110,6 +110,11 @@ class FluentExtension extends DataExtension
     protected $localisedFields = [];
 
     /**
+     * @var bool
+     */
+    private static $frontend_publish_required = true;
+
+    /**
      * Get list of fields that are localised
      *
      * @param string $class Class to get fields for (if parent)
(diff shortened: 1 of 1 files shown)


src/Extension/FluentExtension.php
```diff
@@ -282,7 +282,7 @@ public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null)
         }
 
         // On frontend only show if published in this specific locale
-        if (FluentState::singleton()->getIsFrontend()) {
+        if ($this->owner->config()->get('frontend-publish-required') && FluentState::singleton()->getIsFrontend()) {
```
tractorcow: Config can't have `-` in it, switch to underscore.

Set it to default true with a `private static $frontend_publish_required = true;` in this same extension.

chrispenny: Thanks, @tractorcow . I will update.

PR #300, opened 2017-11-07
Individual fields whitelisting option added.

mfendeksilverstripe:
* added an option to whitelist individual fields
* refactored code slightly
* added some basic tests for whitelisting

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -131,12 +131,6 @@ class FluentExtension extends DataExtension
             return $this->localisedFields[$class] = [];
         }
 
-        // Data and field filters
-        $fieldsInclude = Config::inst()->get($class, 'field_include');
-        $fieldsExclude = Config::inst()->get($class, 'field_exclude');
-        $dataInclude = Config::inst()->get($class, 'data_include');
-        $dataExclude = Config::inst()->get($class, 'data_exclude');
-
         // filter out DB
         foreach ($fields as $field => $type) {
             // If given an explicit field name filter, then remove non-presented fields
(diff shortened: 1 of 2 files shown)


src/Extension/FluentExtension.php
```diff
@@ -161,6 +151,52 @@ public function getLocalisedFields($class = null)
...
+
+        if ($fieldsInclude && in_array($field, $fieldsInclude)) {
+            return true;
+        }
+
+        if ($fieldsExclude && in_array($field, $fieldsExclude)) {
+            return false;
+        }
+
+        if ($fieldsInclude && !$this->anyMatch($field, $fieldsInclude)) {
```
tractorcow: Switch this from a falsey check to a truthy check.

src/Extension/FluentExtension.php
```diff
@@ -161,6 +151,52 @@ public function getLocalisedFields($class = null)
...
+
+        // Data and field filters
+        $fieldsInclude = Config::inst()->get($class, 'field_include');
+        $fieldsExclude = Config::inst()->get($class, 'field_exclude');
+
+        if ($fieldsInclude && in_array($field, $fieldsInclude)) {
+            return true;
+        }
+
+        if ($fieldsExclude && in_array($field, $fieldsExclude)) {
```
tractorcow: It'll be simpler to stick with the anyMatch() method rather than mixing comparisons. Remove these two in_array checks.

mfendeksilverstripe:
@tractorcow Code updated based on feedback. Please review.

src/Extension/FluentExtension.php
```diff
@@ -161,6 +151,44 @@ public function getLocalisedFields($class = null)
...
+
+        if ($dataInclude && !$this->anyMatch($type, $dataInclude)) {
+            return false;
+        }
+
+        if ($dataExclude && $this->anyMatch($type, $dataExclude)) {
+            return false;
+        }
+
+        return true;
```
tractorcow: I need to re-work this logic... I'm realising it's not working the way I'd invisioned.

Not sure about true as the final default.

src/Extension/FluentExtension.php
```diff
@@ -161,6 +151,44 @@ public function getLocalisedFields($class = null)
     }
 
     /**
+     * @param string $field
+     * @param string $type
+     * @param string|null $class
+     * @return bool
+     */
+    public function isFieldLocalised($field, $type, $class = null)
```
tractorcow: We probably should move the `$filter = Config::inst()->get($class, 'translate', Config::UNINHERITED);` check into here too.

src/Extension/FluentExtension.php
tractorcow: @mfendeksilverstripe should we just drop `data_exclude`, and have the three configs only? Anything else is excluded.

tractorcow: I think we should combine the final check;
return (in data_include) && !(in data_exclude)

tractorcow:
Cleaned up and replaced with https://github.com/tractorcow/silverstripe-fluent/pull/303

PR #301, opened 2017-11-08
Locale pages now display their published status relative to locale

chrispenny:
Related Issue: https://github.com/tractorcow/silverstripe-fluent/issues/297

## Purpose
To add new status flags, notices, and actions to the CMS to try and differentiate Pages that have been localised from Pages that are inheriting their content from another locale.

## New status flags
screen shot 2017-11-08 at 3 44 25 pm

## New actions
screen shot 2017-11-09 at 7 12 57 am

## New CMS notices
Displayed in the CMS when editing a Page.

### Locale is being inherited

> Content for this Page is being inherited from another Locale. If you wish you make an independent copy of this Page, please use one of the "Copy" actions provided.

### Locale has been drafted

> A draft has been created for this Locale, however, published content is still being inherited from another Locale. To publish this content for this Locale, use the "Save & publish" action provided.

I'm not sold on the wording that I've used (for flag names and notices), but it was the best I could think of. I'd also be more than happy to change any of the colours used. Originally I was going to make "inherited" a "greyed out" kinda of look, but that didn't meet AA accessibility standards.

I wasn't sure if build files should be committed or not, so I opted for *not*.


src/Extension/FluentSiteTreeExtension.php
```diff
@@ -95,4 +101,150 @@ public function updateLink(&$link, &$action, &$relativeLink)
...
+            }
+        }
+
+        // Make sure no other extensions have removed this field.
+        if ($actionPublish !== null) {
+            $actionPublish->addExtraClass('btn-primary font-icon-rocket');
+            $actionPublish->removeExtraClass('btn-outline-primary font-icon-tick');
+
+            if ($isDraftedInLocale) {
+                $actionPublish->setTitle(_t(SiteTree::class . '.BUTTONSAVEPUBLISH', 'Save & publish'));
```
robbieaverill: Unfortunately we can't use `::class` in translations yet, will need to be a full class string

chrispenny: Thanks, @robbieaverill. Just to confirm, you would be ok with me switching out all usages of `::class` with a string EG: `"SilverStripe\CMS\Model\SiteTree"`?

robbieaverill: In translation calls only, yeah :-)

tests/php/Extension/FluentSiteTreeExtensionTest.php
```diff
@@ -150,4 +151,97 @@ public function testFluentURLs($domain, $locale, $prefixDisabled, $pageName, $ur
...
+        /** @var Page|FluentSiteTreeExtension $page */
+        $page = $this->objFromFixture('Page', 'home');
+        $fields = new FieldList();
+
+        $page->updateCMSFields($fields);
+
+        $this->assertNotNull($fields->fieldByName('LocaleStatusMessage'));
+    }
+
+    public function testUpdateCMSActionsInherited()
```
chrispenny: `testUpdateCMSActionsInherited()` and `testUpdateCMSActionsDrafted()` are very similar, and I considered breaking a large portion of their shared logic out into a separate method, but, ultimately I felt it was better to have all of the assertions in the test methods themselves, rather than out in a shared method. I'd be happy to change this though.

robbieaverill:
@clarkepaul you might be interested in this

chrispenny:
I've finished adding Unit Tests. This is once again ready for review :)

tractorcow:
Testing this out now ;D

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -14,6 +16,8 @@
  */
 class FluentSiteTreeExtension extends FluentVersionedExtension
 {
+    private static $disable_locale_published_status_message = false;
```
tractorcow: I think we can flip this to truthy, given now that 4.0 lets us override truthy config with falsey. :)

PR #302, opened 2017-11-09
Reload the CMS when the locale is changed.

### LINKED ISSUE #298: Changing locales should reload the entire window
When changing locales via the locale switcher part of the window is reloaded with AJAX. This causes things like the pages tree list not to refresh with, for example, different titles between locales.

If we did a hard page reload this wouldn't be a problem.

chrispenny:
Related issue: https://github.com/tractorcow/silverstripe-fluent/issues/298


tractorcow:
Also related https://github.com/tractorcow/silverstripe-fluent/issues/76

PR #303, opened 2017-11-09
Individual fields whitelisting option added.

tractorcow:
Replaces #300

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -23,16 +23,26 @@ use TractorCow\Fluent\State\FluentState;
 /**
  * Basic fluent extension
  *
+ * When determining whether a field is localised, the following config options are checked in order:
+ * - translate (uninherited, for each class in the chain)
+ * - field_exclude
+ * - field_include
+ * - data_exclude
+ * - data_include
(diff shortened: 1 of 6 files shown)


src/Extension/FluentExtension.php
```diff
@@ -23,18 +23,28 @@
 /**
  * Basic fluent extension
  *
+ * When determining whether a field is localised, the following config options are checked in order:
+ * - translate (uninherited, for each class in the chain)
+ * - field_exclude
+ * - field_include
+ * - type_exclude
+ * - type_include
```
mfendeksilverstripe: I believe this should be called `data_exclude` and `data_include` to be consistent with the rest of the code.

mfendeksilverstripe:
Additionally, the tests are failing with PHP 5.6.

robbieaverill:
PHPCS violation fixed

src/Extension/FluentExtension.php
robbieaverill: Yep - seems like a typo. I've pushed a commit that fixes that.

tractorcow:
Ok fixed.

PR #304, opened 2017-11-10
Added new FluentFilteredExtension

chrispenny:
From a feature POV this is complete (if anyone could please give me some feedback), but I've unfortunately run out of time today to write the unit tests, so I will leave this as WIP until I can do that.

---

# Purpose

In combination with the new ability to disable`frontend_publish_required`, this `Extension` attempts to add an additional layer of customisation so that any given Page can be specifically shown or hidden on the front-end for a specific Locale.

# Example

I have the following Locales for my site:

* International
* Australia
* United States
* Japan

I create a new record (EG: a `Page`) which I want the International and Australian Locales to inherit the content for when displaying on the frontend. Within the "Filtered Locales" tab of the edit form I add these`Locales` to the `GridField`.

screen shot 2017-11-10 at 2 42 57 pm

By default, no `Locales` are automatically added to new records.

You should now be able to add new `Locales` to this `GridField` without having to save the record (EG: the `Page`), as saving the record would create a localised version, which is not what we want here.

# Improvements

1) Add some different CMS Status Flags for whether or not Pages are being displayed in the current Locale (a "greyed out" look, maybe).

2) Change the `GridField` to some sort of Checkbox set. I couldn't do this quickly initially because a `CheckboxSetField` requires you save the `DataObject` to save the relationships, and as mentioned above, we **do not** want to create a localised version of the `DataObject` when we add/remove `FilteredLocales`.

3) Add some extension points (or whatever) so that devs can easily add in some default `Locale` records that they might want automatically added to new records that are created.

# Issues

An issue has already been raised regarding the `GridField` search field flagging the form as "changed" when you start typing into it: https://github.com/silverstripe/silverstripe-admin/issues/350

### DIFF
--- a/src/Extension/FluentFilteredExtension.php
+++ b/src/Extension/FluentFilteredExtension.php
@@ -1,0 +1,103 @@
+<?php
+
+namespace TractorCow\Fluent\Extension;
+
+use SilverStripe\Forms\FieldList;
+use SilverStripe\Forms\GridField\GridField;
+use SilverStripe\Forms\GridField\GridFieldAddNewButton;
+use SilverStripe\Forms\GridField\GridFieldConfig_RelationEditor;
+use SilverStripe\Forms\Tab;
+use SilverStripe\ORM\DataExtension;
+use SilverStripe\ORM\DataList;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DataQuery;
+use SilverStripe\ORM\Queries\SQLSelect;
+use TractorCow\Fluent\Model\Locale;
+use TractorCow\Fluent\State\FluentState;
+
+/**
+ * Class FluentFilteredExtension
+ * @package TractorCow\Fluent\Extension
+ *
+ * @property FluentSiteTreeExtension|DataObject $owner
+ * @method DataList|Locale[] FilteredLocales()
+ */
+class FluentFilteredExtension extends DataExtension
+{
+    private static $many_many = [
+        'FilteredLocales' => Locale::class,
+    ];
+
+    public function updateCMSFields(FieldList $fields)
+    {
+        $locales = Locale::get();
+
+        // If there are no Locales, then we're not adding any fields.
+        if ($locales->count() === 0) {
+            return;
+        }
+
+        $fields->insertAfter(new Tab('Locales'), 'Main');
+
+        $config = GridFieldConfig_RelationEditor::create();
+        $config->removeComponentsByType(GridFieldAddNewButton::class);
+
+        // create a GridField to manage what Locales this Page can be displayed in.
+        $fields->addFieldToTab(
+            'Root.Locales',
+            GridField::create(
+                'FilteredLocales',
+                _t(__CLASS__.'.FILTERED_LOCALES', 'Display in the following Locales'),
+                $this->owner->FilteredLocales(),
+                $config
+            )
+        );
+    }
+
+    public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null)
+    {
+        // We don't want this logic applied in the CMS.
+        if (!FluentState::singleton()->getIsFrontend()) {
+            return;
+        }
+
+        $locale = $this->getDataQueryLocale($dataQuery);
+        if (!$locale) {
+            return;
+        }
+
+        $schema = DataObject::getSchema();
+        $table = $schema->baseDataTable(get_class($this->owner));
+        $filteredLocalesTable = $table . '_FilteredLocales';
+
+        $query->addInnerJoin(
+            $filteredLocalesTable,
+            "\"{$filteredLocalesTable}\".\"{$table}ID\" = \"{$table}\".\"ID\" AND \"{$filteredLocalesTable}\".\"{$locale->baseTable()}ID\" = ?",
+            null,
+            20,
+            [
+                $locale->ID,
+            ]
+        );
+    }
+
+    /**
+     * Get current locale from given dataquery
+     *
+     * @param DataQuery $dataQuery
+     * @return Locale|null
+     */
+    protected function getDataQueryLocale(DataQuery $dataQuery = null)
+    {
+        if (!$dataQuery) {
+            return null;
+        }
+
+        $localeCode = $dataQuery->getQueryParam('Fluent.Locale') ?: FluentState::singleton()->getLocale();
+        if ($localeCode) {
+            return Locale::getByLocale($localeCode);
+        }
+
+        return null;
+    }
+}
--- a/src/Extension/FluentFilteredExtension.php
+++ b/src/Extension/FluentFilteredExtension.php
@@ -16,10 +16,7 @@ use TractorCow\Fluent\Model\Locale;
 use TractorCow\Fluent\State\FluentState;
 
 /**
- * Class FluentFilteredExtension
- * @package TractorCow\Fluent\Extension
- *
- * @property FluentSiteTreeExtension|DataObject $owner
+ * @property FluentFilteredExtension|DataObject $owner
  * @method DataList|Locale[] FilteredLocales()
  */
 class FluentFilteredExtension extends DataExtension
@@ -37,7 +34,7 @@ class FluentFilteredExtension extends DataExtension
             return;
         }
 
-        $fields->insertAfter(new Tab('Locales'), 'Main');
+        $fields->insertAfter('Main', new Tab('Locales'));
 
         $config = GridFieldConfig_RelationEditor::create();
         $config->removeComponentsByType(GridFieldAddNewButton::class);
--- a/tests/php/Extension/FluentFilteredExtensionTest.php
+++ b/tests/php/Extension/FluentFilteredExtensionTest.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace TractorCow\Fluent\Tests\Extension;
+
+use Page;
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\Dev\FixtureFactory;
+use SilverStripe\Dev\SapphireTest;
+use TractorCow\Fluent\Extension\FluentFilteredExtension;
+use TractorCow\Fluent\Extension\FluentSiteTreeExtension;
+use TractorCow\Fluent\Model\Domain;
+use TractorCow\Fluent\Model\Locale;
+use TractorCow\Fluent\State\FluentState;
+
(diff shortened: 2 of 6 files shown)


src/Extension/FluentFilteredExtension.php
```diff
@@ -0,0 +1,99 @@
...
+        );
+    }
+
+    /**
+     * Get current locale from given dataquery
+     *
+     * @param DataQuery $dataQuery
+     * @return Locale|null
+     */
+    protected function getDataQueryLocale(DataQuery $dataQuery = null)
```
chrispenny: I was unsure what to do about this. This method is a direct copy of `FluentExtension:: getDataQueryLocale()`, but this extension doesn't extend that (which, it probably shouldn't anyway).

robbieaverill: It's not too bad in terms of duplication. You could put it into a shared trait, maybe? I'd be inclined to live with it existing twice personally

src/Extension/FluentFilteredExtension.php
```diff
@@ -0,0 +1,99 @@
...
+            return;
+        }
+
+        $schema = DataObject::getSchema();
+        $table = $schema->baseDataTable(get_class($this->owner));
+        $filteredLocalesTable = $table . '_FilteredLocales';
+
+        $query->addInnerJoin(
+            $filteredLocalesTable,
+            "\"{$table}ID\" = \"{$table}\".\"ID\" AND \"{$locale->baseTable()}ID\" = ?",
```
tractorcow: Locale base table is just `Locale` so you can hard code that on the right hand side.

Please make sure to prefix all conditions with the table in ANSI-92 SQL `"Table"."Field"`, even if using non-dataobject fields

chrispenny: @tractorcow my Locale base table looks to be `Fluent_Locale`, rather than `Locale`.

tests/php/Extension/FluentFilteredExtensionTest.php
```diff
@@ -0,0 +1,63 @@
...
+            ->setIsDomainMode(false);
+    }
+
+    public function testAugmentSQLFrontend()
+    {
+        FluentState::singleton()
+            ->setLocale('en_NZ')
+            ->setIsFrontend(true);
+
+        $this->assertEquals(1, SiteTree::get()->count());
```
chrispenny: @robbieaverill @tractorcow sorry to bug you both with this, but I'm really struggling with this test. It passes 100% of the time when I run it locally, but is failing in CI:
https://travis-ci.org/tractorcow/silverstripe-fluent/jobs/301652000

The (raw) SQL that gets added as an Inner Join in `augmentSQL` as part of this test is:
```sql
"SiteTree_FilteredLocales"."SiteTreeID" = "SiteTree"."ID" AND "SiteTree_FilteredLocales"."Fluent_LocaleID" = 1
```
Which, I think should match up fine with the fixture file that I've created? And, again, it passes locally - so, I'm really not sure where I'm going wrong.

Thank you for your help :)

src/Extension/FluentFilteredExtension.php
```diff
@@ -0,0 +1,105 @@
...
+ */
+class FluentFilteredExtension extends DataExtension
+{
+    private static $many_many = [
+        'FilteredLocales' => Locale::class,
+    ];
+
+    public function updateCMSFields(FieldList $fields)
+    {
+        parent::updateCMSFields($fields);
```
tractorcow: Parent doesn't do anything, so it's not necessary here.

src/Extension/FluentFilteredExtension.php
```diff
@@ -0,0 +1,105 @@
...
+
+        // If there are no Locales, then we're not adding any fields.
+        if ($locales->count() === 0) {
+            return;
+        }
+
+        $fields->insertAfter(new Tab('Locales'), 'Main');
+
+        $config = GridFieldConfig_RelationEditor::create();
+        $config->removeComponentsByType($config->getComponentByType(GridFieldAddNewButton::class));
```
tractorcow: removeComponentsByType takes a type string name not the component itself.

src/Extension/FluentFilteredExtension.php
```diff
@@ -0,0 +1,105 @@
...
+
+        // create a Checkbox group based on the array
```
tractorcow: Please update this comment if you modify this file again.

src/Extension/FluentFilteredExtension.php
```diff
@@ -0,0 +1,105 @@
...
+        $fields->addFieldToTab(
+            'Root.Locales',
+            GridField::create(
+                'FilteredLocales',
+                'Display in the following Locales',
```
tractorcow: Can you use a localised key for the field title please?


```php
_t(__CLASS__.'.FILTERED_LOCALES',  'Display in the following Locales')
```

src/Extension/FluentFilteredExtension.php
tractorcow: Yep a little duplication is fine; Otherwise you end up with tighter coupling.

tests/php/Extension/FluentFilteredExtensionTest.yml
```diff
@@ -0,0 +1,22 @@
+TractorCow\Fluent\Model\Locale:
+  nz:
+    ID: 1
+    Locale: en_NZ
+    Title: 'English (New Zealand)'
+    URLSegment: newzealand
+
+SiteTree:
+  home:
+    ID: 1
```
tractorcow: I would recommend highly against hard-coded IDs even in tests.

If you found it challenging to express your logic purely in yml, you'll need to put your logic into setUp() instead.

tests/php/Extension/FluentFilteredExtensionTest.php
tractorcow: It could be due to the fixture file not being setup correctly; Please see my comment on the yml file.

chrispenny: These are working now (using `FixtureFactory` rather than `yml`), however, I might still be doing wrong as between tests, the number of DB objects continues to increase (it looks like `SiteTree` objects aren't being removed).

I've got `$factory->clear()` running, but that didn't do the trick. I ended up defining an empty `yml` fixture which looked to clear the objects correctly.

I was following the docs on Fixtures (https://docs.silverstripe.org/en/4/developer_guides/testing/how_tos/fixturefactories/), but it looked outdated.

I'll be in the office tomorrow if it's easier to discuss in person.

src/Extension/FluentFilteredExtension.php
tractorcow: Oh right, sorry. :)

tests/php/Extension/FluentFilteredExtensionTest.php
```diff
@@ -0,0 +1,86 @@
...
+        FluentState::singleton()
+            ->setLocale('en_NZ')
+            ->setIsDomainMode(false);
+
+        /** @var FixtureFactory $factory */
+        $factory = $this->getFixtureFactory();
+
+        $factory->clear();
+
+        $factory->createObject('TractorCow\Fluent\Model\Locale', 'nz', [
```
tractorcow: I like this a lot better. :)

tests/php/Extension/FluentFilteredExtensionTest.php
tractorcow: Well it seems green now!

tests/php/Extension/FluentSiteTreeExtensionTest.yml
```diff
@@ -33,41 +33,34 @@ TractorCow\Fluent\Model\Domain:
 
 Page:
   nz-page:
-    ID: 1
     Title: A Page
     URLSegment: a-page
   home:
-    ID: 2
```
tractorcow: Tests are green, but there is duplication between this yml file and the setUp() in php. It's a little ambiguous where the content is coming from.

All that was necessary was to remove the hard-coded IDs... with that change only, do you need the PHP changes as well? Lookang at this YML file it seems to be correct now.

chrispenny: This is the fixture for `FluentSiteTreeExtensionTest`. This was just me going back and fixing some tests from my other feature.

chrispenny: I had a lot of issues with trying to put this into `FluentFilteredExtensionTest.yml`, I had it looking something like this (based on the docs):
```yml
TractorCow\Fluent\Model\Locale:
  nz:
    Locale: en_NZ
    # Title, URL Segment
  au:
    Locale: en_AU
    # Title, URL Segment

SiteTree:
  home:
    # Title, etc
    FilteredLocales:
      - =>TractorCow\Fluent\Model\Locale.nz
      - =>TractorCow\Fluent\Model\Locale.au
```

But I get an error due to `FixtureFactory::parseValue($value)` not allowing array values:
```
substr() expects parameter 1 to be string, array given

/var/www/html/vendor/silverstripe/framework/src/Dev/FixtureFactory.php:253
```

I then tried comma separated rather than list (which the docs said was ok), but I got a different error as it looked like the value wasn't be split.
```
SiteTree:
  home:
    Title: Home
    URLSegment: home
    FilteredLocales:
      =>TractorCow\Fluent\Model\Locale.nz,TractorCow\Fluent\Model\Locale.au
```

Error:
```
InvalidArgumentException: No fixture definitions found for "=>TractorCow\Fluent\Model\Locale.nz,=>TractorCow\Fluent\Model\Locale.au"
```

I started a discussion in the help channel, but didn't really get to the bottom of it, and whether it was a bug, or me doing something wrong. In any case, the PHP solution worked.

tractorcow: I guess I'll need to check it out and have a play with it sometime.

mfendeksilverstripe: @chrispenny @tractorcow I used this and it worked for me fine:
```
SilverStripe\CMS\Model\SiteTree:
  home:
    FilteredLocales: =>TractorCow\Fluent\Model\Locale.nz,=>TractorCow\Fluent\Model\Locale.au
```

tractorcow: I've updated the test fixtures; Should be ok now. ;)

PR #305, opened 2017-11-10
FIX Do not treat CMS preview requests as frontend

### LINKED ISSUE #296: On fresh install, no locales have any draft pages
Steps:

* SS4 vanilla install, require fluent
* Go to admin, create some locales
* Define a fallback chain for some of them
* Refresh page, switch to any locale (default for example)
* Go to Pages and choose a page

Expected result: shows draft version of a page

Actual result: "The requested page could not be found." in the preview panel

---

My fallback chain is en_AU > en_NZ > en_US (global default).

I suppose that with the fallback system, if the default locale doesn't have a record for a page it should fallback to reading from the global SiteTree tables? In this case en_US wouldn't have a home page for example, so would read from the global record instead.

robbieaverill:
Fixes #296

### DIFF
--- a/src/Middleware/InitStateMiddleware.php
+++ b/src/Middleware/InitStateMiddleware.php
@@ -70,6 +70,12 @@ class InitStateMiddleware implements HTTPMiddleware
                 return false;
             }
         }
+
+        // If using the CMS preview, do not treat the site as frontend
+        if ($request->getVar('CMSPreview')) {
+            return false;
+        }
+
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


src/Middleware/InitStateMiddleware.php
```diff
@@ -70,6 +70,12 @@ public function getIsFrontend(HTTPRequest $request)
                 return false;
             }
         }
+
+        // If using the CMS preview, do not treat the site as frontend
+        if ($request->getVar('CMSPreview')) {
```
tractorcow: Awesome thanks. :)

PR #309, opened 2017-11-19
Improve augmentSQL logic for FluentFilteredExtension. Fixes #307

### LINKED ISSUE #307: v4 FluentFilteredExtension: Issue with FilteredLocale logic on Pages with their own localisation
When viewing a page that has it's own (published) localisation, we shouldn't require that the same locale is also added into that Page's `FilteredLocales` records.

This is one for me to pick up (please feel free to assign it to me) - I'm just documenting it here so that I don't forget (as I might not be able to look at it today).

chrispenny:
Related issue: https://github.com/tractorcow/silverstripe-fluent/issues/307

When viewing a page that has it's own (published) localisation, we shouldn't require that the same locale is also added into that Page's FilteredLocales records.

I've updated the tests to represent the different options now available.

### DIFF
--- a/src/Extension/FluentFilteredExtension.php
+++ b/src/Extension/FluentFilteredExtension.php
@@ -67,7 +67,7 @@ class FluentFilteredExtension extends DataExtension
         $table = $schema->baseDataTable(get_class($this->owner));
         $filteredLocalesTable = $table . '_FilteredLocales';
 
-        $query->addInnerJoin(
+        $query->addLeftJoin(
             $filteredLocalesTable,
             "\"{$filteredLocalesTable}\".\"{$table}ID\" = \"{$table}\".\"ID\" AND \"{$filteredLocalesTable}\".\"{$locale->baseTable()}ID\" = ?",
             null,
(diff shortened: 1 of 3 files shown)


src/Extension/FluentFilteredExtension.php
```diff
@@ -97,4 +107,20 @@ protected function getDataQueryLocale(DataQuery $dataQuery = null)
...
     }
+
+    /**
+     * Get the localised table name with the localised suffix and optionally with a locale suffix for aliases
+     *
+     * @param string $tableName
+     * @param string $locale
+     * @return string
+     */
+    public function getLocalisedTable($tableName, $locale = '')
```
chrispenny: Some publication of code here from `FluentExtension`.

tractorcow:
>When viewing a page that has it's own (published) localisation, we shouldn't require that the same locale is also added into that Page's FilteredLocales records.

If you want to hide a page in a locale, it'll be forced to appear even if it's explicitly unchecked in the CMS. It feels like a contradiction of the purpose of this extension, which should override the page published state.

src/Extension/FluentFilteredExtension.php
```diff
@@ -76,6 +76,16 @@ public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null)
                 $locale->ID,
             ]
         );
+
+        $joinAlias = $this->getLocalisedTable($table, $locale->Locale);
+
+        // Only select records that either have their own Localisation, or have been added as a Filtered Locale.
```
tractorcow: Note, this also fails if using FluentFilteredExtension but not FluentExtension. ;)

chrispenny:
@tractorcow That's a fair comment. I will close this PR for now, and if we're finding it to be a burden then I'll revisit.

tractorcow:
Might I suggest an alternative that could suit your project... have an "onAfterPublish" event on your page that added it to the filtered locales. :D

chrispenny:
Thanks @tractorcow, that sounds like a perfect solution :)

PR #310, opened 2017-11-20
Add additional styles for FluentFilteredExtension SiteTree flags. Fixes #308

### LINKED ISSUE #308: v4 FluentFilteredExtension: Add styles for Site Tree
Another one for me to look at (please feel free to assign this to me).

Add some additional styles so that it is clear which Pages are and are not currently visible as you browse the CMS in any particular Locale.

Something like this (where the "About us" page isn't visible on this current Locale):
screen shot 2017-11-17 at 8 35 40 am

I'd also like to look into how easily I could add similar visibility to things like SiteTree dropdown fields, but that may end up being a further improvement later down the line.

chrispenny:
Related issue/improvement: https://github.com/tractorcow/silverstripe-fluent/issues/308


client/src/styles/fluent.scss
```diff
@@ -45,11 +48,21 @@ $fluent-color-inherited-text-badge: rgb(122, 69, 0);
...
+    }
+
     span {
       &.badge {
         &.status-fluentinherited {
           color: $fluent-color-inherited-text-badge;
         }
+
+        &.status-fluentfiltered {
+          display: none;
```
chrispenny: By default, I've hidden the flag text so that it doesn't conflict with the text from other flags (EG: "Draft", "Inherited", etc).

Potentially instead of adding a flag I could just append an extra class (somehow?), but I thought having the text there (in code) would mean other devs can just add their own style to display it (if they desire).

src/Extension/FluentFilteredExtension.php
```diff
@@ -63,9 +115,8 @@ public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null)
             return;
         }
 
-        $schema = DataObject::getSchema();
-        $table = $schema->baseDataTable(get_class($this->owner));
-        $filteredLocalesTable = $table . '_FilteredLocales';
+        $table = $this->getDataObjectBaseTable();
```
tractorcow: Could we just use `$this->owner->baseTable()` instead of a new method?

client/src/styles/fluent.scss
tractorcow: Could it be shown on hover or something instead?

What does the "conflict" look like? maybe we could make it work / fit.

chrispenny: @tractorcow I've changed the font colour just for the purposes of easily showing how it displays (otherwise, atm the default colour for this flag text is white):

- "About Us" is inheriting content and has not been added to "Filtered Locales".
- "Page not found" is drafted in the current Locale but has not been added to "Filtered Locales".
- "Server error" is published in the current Locale but has not been added to "Filtered Locales".

screen shot 2017-11-21 at 11 54 02 am

tractorcow:
Looks good but I'll need to test to make sure it looks ok.

PR #329, opened 2018-01-18
Added additional configuration options to allow custom routing rules.

mfendeksilverstripe:
Routing controller classes are currently hard coded which is not ideal. This patch allows these classes to be overrides via config API. This is very useful if custom routing functionality is necessary for a project that uses `Fluent`.


tractorcow:
I suggest https://github.com/tractorcow/silverstripe-fluent/pull/330 instead of additional configuration.

robbieaverill:
Replaced by #330 - feel free to reopen if there's more that needs to be done

PR #336, opened 2018-01-30
update so the correct sitemap template is being uses (with alternate languages)

guyvanbael:


dhensby:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. No action is needed at the moment, this is for your information.

In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request)

src/Control/FluentSitemapController.php
```diff
@@ -0,0 +1,22 @@
...
+     * @package fluent
+     * @author Damian Mooyman <damian.mooyman@gmail.com>
+     */
+    class FluentSitemapController extends GoogleSitemapController
+    {
+        public function init()
+        {
+            parent::init();
+            // Reset any session locale to use the default locale as the standard 'base' locale
+            //Fluent::set_persist_locale(Fluent::default_locale(true));
```
tractorcow: This class is redundant isn't it? Just use the default google sitemap controller.

src/Extension/FluentDirectorExtension.php
```diff
@@ -70,9 +72,9 @@ public function updateRules(&$rules)
 
         // @todo - Build SitemapController after GoogleSitemap module is migrated to ss4
         // If Google sitemap module is installed then replace default controller with custom controller
-        //if (class_exists(GoogleSitemapController::class)) {
-        //    $rules['sitemap.xml'] = SitemapController::class;
-        //}
+        if (class_exists(GoogleSitemapController::class)) {
+            $rules['sitemap.xml'] = FluentSitemapController::class;
```
tractorcow: If you use a fluent-namespaced controller you need the fluent-namespaced template also.

Just use the default googlesitemapcontroller class from will.r's module.

guyvanbael: If you use the the default Wilr's default googlesitemapcontroller, then the default GoogleSitemapController_sitemap.ss gets rendered and that one doesn't contain the alternate languages. Or is there a way to route the default controller to the FluentSitemapController.ss?

tractorcow:
Replaced with https://github.com/tractorcow/silverstripe-fluent/pull/352

Issue #328, opened 2018-01-18
Provide migration guide from Translatable to Fluent 4.0


robbieaverill:
Related: https://github.com/tractorcow/silverstripe-fluent/issues/278

As an important part of the migration process for CWP 1.x users to using CWP 2.x, they'll be switching from using Translatable to Fluent. We need to ensure that we have a comprehensive documentation guide for users that outlines the architectural differences between the two modules, as well as the changes for common public API methods (like getting the current language/locale, and the list of what's available).

We should also ensure that the BuildTask to assist with migrating from Translatable to Fluent (in this module) is working as expected (issue: https://github.com/tractorcow/silverstripe-fluent/issues/278). We can assume at this point that Translatable will be SilverStripe 3 only, and Fluent would be SilverStripe 4, that the Translatable code is most likely gone but the data in the DB is still available (?).

robbieaverill:
Finished with #339

PR #339, opened 2018-02-01
NEW Add task to convert SS3 Translatable data to SS4 Fluent

robbieaverill:
To run this task, ensure you have updated your system to SS4, run a dev/build
and added the locales you need to use in the CMS Locale interface. Next run
the build task, which will write and/or publish your content via the Fluent
ORM, then clean up the old Translatable tables and extra columns.

This task is based on the existing task from SS3.

Part of #328

### DIFF
--- a/docs/en/domain-configuration.md
+++ b/docs/en/domain-configuration.md
@@ -47,7 +47,7 @@ Either one of:
  * Ensure that all domains configured are the only domains that the site can be accessed under, or
  * Add `SS_FLUENT_FORCE_DOMAIN=true` to your `.env` file, or
  * Set the `TractorCow\Fluent\Extension\FluentDirectorExtension.force_domain` config to true.
-   (this will effect your development environment).
+   (this will affect your development environment).
 
 Outside of these conditions, the domains configuration property will be entirely ignored, meaning you will not normally
 need to alter your SilverStripe configuration between environments.
--- a/src/Task/ConvertTranslatableTask.php
+++ b/src/Task/ConvertTranslatableTask.php
@@ -181,7 +181,7 @@ class ConvertTranslatableTask extends BuildTask
 
                         // Drop the "Locale" column from the base table
                         Debug::message('Dropping "Locale" column from ' . $baseTable, false);
-                        DB::query(sprintf('ALTER TABLE "%s" DROP COLUMN "Locale"', Convert::raw2sql($baseTable)));
+                        DB::query(sprintf('ALTER TABLE "%s" DROP COLUMN "Locale"', $baseTable));
 
                         // Drop the "_translationgroups" translatable table
                         Debug::message('Deleting Translatable table ' . $groupTable, false);
--- a/src/Task/ConvertTranslatableTask.php
+++ b/src/Task/ConvertTranslatableTask.php
@@ -85,9 +85,6 @@ class ConvertTranslatableTask extends BuildTask
 
     public function run($request)
     {
-        // Extend time limit
-        set_time_limit(100000);
-
         $this->checkInstalled();
 
         // we may need some privileges for this to work
@@ -152,8 +149,10 @@ class ConvertTranslatableTask extends BuildTask
 
                             // Check for obsolete classes that don't need to be handled any more
                             if ($instance->ObsoleteClassName) {
-                                Debug::message("Skipping {$instance->ClassName} with ID {$instanceID} because it from an obsolete class",
-                                    false);
+                                Debug::message(
+                                    "Skipping {$instance->ClassName} with ID {$instanceID} because it from an obsolete class",
+                                    false
+                                );
                                 continue;
                             }
 
--- a/src/Task/ConvertTranslatableTask.php
+++ b/src/Task/ConvertTranslatableTask.php
@@ -125,7 +125,7 @@ class ConvertTranslatableTask extends BuildTask
     /**
      * Gets all classes with FluentExtension
      *
-     * @return array Aray of classes to migrate
+     * @return array Array of classes to migrate
      */
     public function fluentClasses()
     {
--- a/src/Task/ConvertTranslatableTask.php
+++ b/src/Task/ConvertTranslatableTask.php
@@ -1,0 +1,268 @@
+<?php
+
+namespace TractorCow\Fluent\Task;
+
+use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Convert;
+use SilverStripe\Dev\BuildTask;
+use SilverStripe\Dev\Debug;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DB;
+use SilverStripe\ORM\Queries\SQLSelect;
+use SilverStripe\Security\DefaultAdminService;
+use SilverStripe\Security\Security;
+use SilverStripe\Versioned\Versioned;
+use TractorCow\Fluent\Extension\FluentExtension;
+use TractorCow\Fluent\Extension\FluentFilteredExtension;
+use TractorCow\Fluent\Model\Locale;
+use TractorCow\Fluent\State\FluentState;
+use TractorCow\Fluent\Task\ConvertTranslatableTask\Exception;
+
+/**
+ * Provides migration from the Translatable module in a SilverStripe 3 website to the Fluent format for SilverStripe 4.
+ * This task assumes that you have upgraded your website to run on SilverStripe 4 already, and you want to migrate the
+ * existing data from your project into a format that is compatible with Fluent.
+ *
+ * Don't forget to:
+ *
+ * 1. Back up your DB
+ * 2. dev/build
+ * 3. Log into the CMS and set up the locales you want to use
+ * 4. Back up your DB again
+ * 5. Log into the CMS and check everything
+ */
+class ConvertTranslatableTask extends BuildTask
+{
+    protected $title = "Convert Translatable > Fluent Task";
+
+    protected $description = "Migrates site DB from SS3 Translatable DB format to SS4 Fluent.";
+
+    private static $segment = 'ConvertTranslatableTask';
+
+    /**
+     * Fields that are localised per class (key = class name)
+     *
+     * @var array[]
+     */
+    protected $localisedFields = [];
+
+    /**
+     * Checks that fluent is configured correctly
+     *
+     * @throws ConvertTranslatableTask\Exception
+     */
+    protected function checkInstalled()
+    {
+        // Assert that fluent is configured
+        $locales = Locale::getLocales();
+        if (empty($locales)) {
+            throw new Exception("Please configure Fluent locales (in the CMS) prior to migrating from translatable");
+        }
+
+        $defaultLocale = Locale::getDefault();
+        if (empty($defaultLocale)) {
+            throw new Exception(
+                "Please configure a Fluent default locale (in the CMS) prior to migrating from translatable"
+            );
+        }
+    }
+
+    /**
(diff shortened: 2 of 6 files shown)


docs/en/domain-configuration.md
```diff
@@ -47,7 +47,7 @@ Either one of:
  * Ensure that all domains configured are the only domains that the site can be accessed under, or
  * Add `SS_FLUENT_FORCE_DOMAIN=true` to your `.env` file, or
  * Set the `TractorCow\Fluent\Extension\FluentDirectorExtension.force_domain` config to true.
-   (this will effect your development environment).
+   (this will affect your development environment).
```
NightJar: 👍

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+                }
+            }
+        }
+        return array_unique($classes);
+    }
+
+    public function run($request)
+    {
+        // Extend time limit
+        set_time_limit(100000);
```
NightJar: This isn't a wise idea IMO - PHP's CLI SAPI has a timelimit of `0` (inifinite) _by default_ - unless this is set at run time during SS's boot (which would be a bit foolish IMO) we shouldn't 'increase' the limit from `infinity` to `some_large_but_still_finite_number`.

I've been burnt by jobs doing this in the past, and it is certainly both tiresome to debug (looking for errors as opposed to timeouts), and 4x as unpleasant to clean up after (albeit you're using transactions here so hopefully it wouldn't' be so bad). When execution stops, it simply stops - with no cleanup IIRC.

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+                    && $class::has_extension(FluentFilteredExtension::class)
+                ) {
+                    $class::remove_extension(FluentFilteredExtension::class);
+                }
+
+                // Select all instances of this class in the base table, where the Locale field is not null.
+                // Translatable has a Locale column on the base table in SS3, but Fluent doesn't use it. Newly
+                // created records via SS4 Fluent will not set this column, but will set it in {$baseTable}_Localised
+                $instances = DataObject::get($class, sprintf(
+                    '"%s"."Locale" IS NOT NULL',
```
NightJar: Although it's an _extreme_ edge case, assuming ANSI standard SQL isn't always a good thing (but it _should_ be!) - SQLServer uses `[` and `]` to delimit objects (not `"`). I believe there's a method or member field on the Database Adapter class that houses the more ideal value to use - perhaps we should leverage this as an example of good practice?

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+                        });
+                }
+
+                // Drop the "Locale" column from the base table
+                Debug::message('Dropping "Locale" column from ' . $baseTable, false);
+                DB::query(sprintf('ALTER TABLE "%s" DROP COLUMN "Locale"', Convert::raw2sql($baseTable)));
+
+                // Drop the "_translationgroups" translatable table
+                Debug::message('Deleting Translatable table ' . $groupTable, false);
+                DB::query(sprintf('DROP TABLE IF EXISTS "%s"', $groupTable));
```
NightJar: Again with the extreme edge cases, but `IF EXISTS` isn't always a thing either: https://blogs.msdn.microsoft.com/sqlserverstorageengine/2015/11/03/drop-if-exists-new-thing-in-sql-server-2016/

src/Task/ConvertTranslatableTask.php
robbieaverill: Fair comment. This is here from the original SS3 task's code

src/Task/ConvertTranslatableTask.php
robbieaverill: We use ANSI as standard so that the DB connectors can convert as necessary into the appropriate syntax for the various DB drivers we support - at the moment SQL Server isn't one of those. This is mainly to provide compatibility for MySQL and PostgreSQL

src/Task/ConvertTranslatableTask.php
robbieaverill: See above, we don't support SQL Server

src/Task/ConvertTranslatableTask.php
NightJar: cool, fair enough :)

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+        }
+    }
+
+    /**
+     * Do something inside a DB transaction
+     *
+     * @param callable $callback
+     * @throws \Exception
+     */
+    protected function withTransaction($callback)
```
tractorcow: Use `DB::get_conn()->withTransaction` instead.

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+    public function getTranslatedFields($class)
+    {
+        if (isset($this->localisedFields[$class])) {
+            return $this->localisedFields[$class];
+        }
+        $fields = [];
+        $hierarchy = ClassInfo::ancestry($class);
+        foreach ($hierarchy as $class) {
+            // Skip classes without tables
+            if (!DataObject::getSchema()->classHasTable($class)) {
```
tractorcow: You can filter this with `ClassInfo::ancestry($class, true)` instead of manually.

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    public function run($request)
+    {
+        // Extend time limit
+        set_time_limit(100000);
+
+        // we may need some privileges for this to work
+        // without this, running under sake is a problem
+        // maybe sake could take care of it ...
```
tractorcow: Can you use `Member::actAs()` API instead?

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+        Security::setCurrentUser(
+            DefaultAdminService::singleton()->findOrCreateDefaultAdmin()
```
tractorcow: Will fail if default admin isn't enabled. Will need a work-around.

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+                ));
+
+                foreach ($instances as $instance) {
+                    /** @var DataObject $instance */
+
+                    // Get the Locale column directly from the base table, since the SS ORM will not include it
+                    $instanceLocale = SQLSelect::create()
+                        ->setFrom($baseTable)
+                        ->setSelect('Locale')
+                        ->setWhere(['ID' => $instance->ID])
```
tractorcow: `'"ID"'` vs `'ID'` since you need to SQL escape this fragment. Postgres will complain. I'd even add basetable to be consistent.

```
->setWhere(["\"{$baseTable}\".\"ID\"" => $instance->ID])
```

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+                    '"%s"."Locale" IS NOT NULL',
+                    $baseTable
+                ));
+
+                foreach ($instances as $instance) {
+                    /** @var DataObject $instance */
+
+                    // Get the Locale column directly from the base table, since the SS ORM will not include it
+                    $instanceLocale = SQLSelect::create()
+                        ->setFrom($baseTable)
```
tractorcow: From needs to be ansi quoted.

```
->setFrom("\"{$baseTable}\"")
```

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+                        ->setSelect('Locale')
```
tractorcow: Same as above, ansi quoted.

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,268 @@
...
+                    $instanceLocale = $instanceLocale['Locale'];
+
+                    // Check for obsolete classes that don't need to be handled any more
+                    if ($instance->ObsoleteClassName) {
+                        Debug::message("Skipping {$instance->ClassName} with ID {$instanceID} because it from an obsolete class", false);
+                        continue;
+                    }
+
+                    Debug::message(
+                        "Updating {$instance->ClassName} {$instance->MenuTitle} ({$instance->ID}) with locale {$instanceLocale}",
```
tractorcow: Some objects don't have MenuTitle, just use Title.

src/Task/ConvertTranslatableTask.php
robbieaverill: We could just push a message and return early in this case

tractorcow: findOrCreateAdmin() works as a failover ;P

src/Task/ConvertTranslatableTask.php
robbieaverill: Do you have a suggestion for how to implement this in a way that will work for the CLI and via web as well? I realise servers like Apache may have their own timeouts, but the process will still continue to run in the background. I don't like the idea of creating a script that can run infinitely in the background (if initiated from the web)

src/Task/ConvertTranslatableTask.php
robbieaverill: We actually don't need this method any more. The Fluent ORM handles it internally

src/Task/ConvertTranslatableTask.php
NightJar: By default; simply do not attempt to address this (non) issue.

Each PHP SAPI has it's own configuration, and by default this is set to a positive integer (which I cannot recall) for Apache (and other non-CLI SAPIs afaik). So unless a system administrator has made a less than advisable adjustment to the php.ini for that sapi, this should be fine.

src/Task/ConvertTranslatableTask.php
robbieaverill: Lost you - leave it as it?

robbieaverill:
All comments updated, except [this one](https://github.com/tractorcow/silverstripe-fluent/pull/339#discussion_r165860039) - rested =)

src/Task/ConvertTranslatableTask.php
robbieaverill: Looking back through the git history there doesn't seem to be an explicit reason for this to be here, so unless @tractorcow disagrees I'm removing it now

src/Task/ConvertTranslatableTask.php
```diff
@@ -0,0 +1,209 @@
...
+                                        throw new Exception("Failed to publish");
+                                    } else {
+                                        Debug::message("  --  Published", false);
+                                    }
+                                });
+                        }
+
+                        // Drop the "Locale" column from the base table
+                        Debug::message('Dropping "Locale" column from ' . $baseTable, false);
+                        DB::query(sprintf('ALTER TABLE "%s" DROP COLUMN "Locale"', Convert::raw2sql($baseTable)));
```
tractorcow: Don't raw2sql on table names.

tractorcow: It doesn't work. You can't put quotes / slashes etc into tables. raw2sql() only works on values not identifiers. In other words, it's the wrong escape type.

NightJar: Right, designed for values not identifiers. That makes sense. Thank you :)

Issue #311, opened 2017-11-29
no visual indication in which locale user is editing content


guyvanbael:
in version 3.8 users could easily see in which locale they were editing contect.
Version 4 has no visual aid.
And also... the locale isn't added to the urlsegment, so the link is wrong.
<img width="488" alt="schermafbeelding 2017-11-29 om 14 39 18" src="https://user-images.githubusercontent.com/3098809/33378006-b57d931c-d513-11e7-87f7-f219e87122c1.png">

tractorcow:
Thanks we can fix the preview link without too much trouble, I believe.

The "current locale" message is still being migrated I believe from the old fluent. We aren't 100% certain it'll be the same as it used to be.

robbieaverill:
Note: Use Bootstrap classes (`.alert.alert-warning`/`.alert-info`) instead of `.message.warning` for example, since they're being removed

tractorcow:
Sorry this should have been on the roadmap for upgrade. I've added this now https://github.com/tractorcow/silverstripe-fluent/issues/261

raissanorth:
The issue regarding the current local message is addressed in https://github.com/tractorcow/silverstripe-fluent/pull/343

robbieaverill:
Nice @raissanorth! I've logged the other issue at https://github.com/tractorcow/silverstripe-fluent/issues/344 and will tackle it now

robbieaverill:
@tractorcow - @clarkepaul came up with some cool designs today for a more accurate way of communicating the locale the content exists in, which is potentially more relevant than the Fluent 3 alert message - since we now have inherited pages that are visible on the frontend (if using fallback locales) and invisible pages that can still be copied through to make them owned by and visible in the current locale (if using exclusive locales).

We had a couple of questions for you though:

* If I have en_GB as my default locale, en_NZ as another that falls back to en_GB and de_DE as an exclusive locale that doesn't fallback, I expect to see en_GB owned pages in the sitetree (greyed out) when I'm viewing it in the en_NZ locale. I should also expect to see en_GB pages in the sitetree when editing in de_DE, but should I expect to see en_NZ pages when editing in de_DE?

* How do non-translated fields fall back? E.g. if I translate all page fields except for Content (for example), does the content value get read from the default locale always? Can I modify that default value when changing the Content in a non-default locale too, like in Fluent 3?

tractorcow:
> but should I expect to see en_NZ pages when editing in de_DE?

[...]

PR #343, opened 2018-02-08
FIX Visual indicator of editing locale

raissanorth:
Reintroduces label indications of current editing locale on translated CMS fields. Addresses one of the two issues raised in #311. 

TODO: Migrate the functionality to detect modified fields.

Editing in default locale:
image

Editing in non-default locale:
image

cc @sachajudd @clarkepaul


tractorcow:
Yay

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,118 @@ public function cacheKeyComponent()
...
+                    $currentLocale = Locale::getCurrentLocale();
+                    $locale = $currentLocale->Locale;
+                    $language = $currentLocale ? strtok($locale, '_') : '';
+                    $title = $field->Title();
+
+                    $titleClasses = 'fluent-locale-label';
+                    $modifiedTitle = 'Using default locale value';
+
+                    // Add a visual indicator for whether the value has been changed from the default locale
+//                    if (Fluent::isFieldModified($this->owner, $field, $locale)) {
```
tractorcow: Can we please remove this instead of commenting it out?

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,118 @@ public function cacheKeyComponent()
...
+                // E.g. UploadField()
+                $field = $fields->dataFieldByName($translatedField);
+                if (!$field && preg_match('/^(?<field>\w+)ID$/', $translatedField, $matches)) {
+                    $field = $fields->dataFieldByName($matches['field']);
+                }
+
+                // Highlight any translatable field
+                if ($field && !$field->hasClass('LocalisedField')) {
+                    // Add a language indicator next to the fluent icon
+                    $currentLocale = Locale::getCurrentLocale();
```
tractorcow: Shift this logic up above the foreach so it doesn't need rec-calculating for each field.

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,118 @@ public function cacheKeyComponent()
...
+
+        $localeNames     = Locale::get()->map('Locale', 'Title')->toArray();
+        $defaultLocale   = Locale::getDefault();
+        $isDefaultLocale = ($defaultLocale === Locale::getCurrentLocale());
+        $defaultLocaleKey = $defaultLocale ? $defaultLocale->Locale : '';
+        $currentLocale = Locale::getCurrentLocale();
+        $currentLocaleKey = $currentLocale ? $currentLocale->Locale : '';
+        $messageClass    = ($isDefaultLocale) ? 'good' : 'notice';
+        $message         = ($isDefaultLocale)
+            ? _t('Fluent.DefaultLocale', 'This is the default locale')
```
tractorcow: `__CLASS__` instead of old `Fluent.` please for i18n keys.

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,118 @@ public function cacheKeyComponent()
...
+            : _t('Fluent.DefaultLocaleIs', 'The default locale is') . ' ' . $localeNames[$defaultLocaleKey];
```
tractorcow: Let's update to use parameters.

```php
_t(
	__CLASS__.'.DefaultLocaleIs',
	'The default locale is {locale}',
	['locale' => $localeNames[$defaultLocaleKey]]
);
```

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,118 @@ public function cacheKeyComponent()
...
+
+        $fields->unshift(
+            LiteralField::create(
+                'CurrentLocaleMessage',
+                sprintf(
+                    '<p class="message %s">' . _t(
+                        'Fluent.EditingIn',
+                        'Please note: You are editing in'
+                    ) . ' %s. %s.</p>',
```
tractorcow: Same as above, can we use named parameters for the message?

```php
_t(
	__CLASS__.'.EditingIn',
	'Please note: You are editing in {locale}',
	['locale' => $localeNames[$defaultLocaleKey]]
);
```

tractorcow:
Good job @raissanorth just a few i18n good-to-haves :)

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,118 @@ public function cacheKeyComponent()
     {
         return 'fluentlocale-' . FluentState::singleton()->getLocale();
     }
+
+// <editor-fold defaultstate="collapsed" desc="CMS Field Augmentation">
```
robbieaverill: Can we remove these IDE tags?

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,118 @@ public function cacheKeyComponent()
...
+        if (Config::inst()->get(__CLASS__, 'disable_current_locale_message')) {
+            return $this;
+        }
+
+        // If the field is already present, don't add it a second time
+        if ($fields->fieldByName('CurrentLocaleMessage')) {
+            return $this;
+        }
+
+        $localeNames     = Locale::get()->map('Locale', 'Title')->toArray();
```
robbieaverill: If you use `Locale::getCached()` you'll get a minor performance benefit, but follow the same API use as the rest of the module

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,118 @@ public function cacheKeyComponent()
...
+        $defaultLocale   = Locale::getDefault();
+        $isDefaultLocale = ($defaultLocale === Locale::getCurrentLocale());
+        $defaultLocaleKey = $defaultLocale ? $defaultLocale->Locale : '';
+        $currentLocale = Locale::getCurrentLocale();
+        $currentLocaleKey = $currentLocale ? $currentLocale->Locale : '';
+        $messageClass    = ($isDefaultLocale) ? 'good' : 'notice';
```
robbieaverill: Let's use Bootstrap classes here: `alert alert-success` or `alert alert-info`

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,118 @@ public function cacheKeyComponent()
...
+        $message         = ($isDefaultLocale)
+            ? _t('Fluent.DefaultLocale', 'This is the default locale')
+            : _t('Fluent.DefaultLocaleIs', 'The default locale is') . ' ' . $localeNames[$defaultLocaleKey];
+
+        $fields->unshift(
+            LiteralField::create(
+                'CurrentLocaleMessage',
+                sprintf(
+                    '<p class="message %s">' . _t(
```
robbieaverill: Remove `message` and let it be pushed in by `$messageClass` (see previous comment)

raissanorth:
Thanks for your feedback @tractorcow. I made the requested changes. Thanks @NightJar for your help with this migration.

src/Extension/FluentExtension.php
raissanorth: Of course we can.

src/Extension/FluentExtension.php
NightJar: I vote `alert alert-info`, since that's what the messages are doing :)

robbieaverill: It's doing both, depends on which locale you're in

raissanorth: Sweet, looking good.

robbieaverill:
I'm not sure what I think about bringing across the individual field indicators. In Fluent 3 they were useful because Fluent localised on a field by field basis, but in Fluent 4 it localises a whole row for a locale, so I'm not sure it'll provide any benefit really (at a field level).

What do you think @tractorcow?

tractorcow:
@raissanorth you're right, it doesn't make much sense to distinguished fields that have customised content. All that's really important is indicating which fields CAN be localised.

Also, can you please remove the other commented out code blocks?

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,121 @@ public function cacheKeyComponent()
...
+//                            'data-default-locale-value',
+//                            $this->owner->{Fluent::db_field_for_locale($field->getName(), Locale::getDefault())}
+//                        );
+//                    }
+
+                    $field->addExtraClass('LocalisedField');
+                }
+
+                // Remove translation DBField from automatic scaffolded fields
+//                foreach (Fluent::locales() as $locale) {
```
tractorcow: Sorry, I ment to indicate that all commented blocks should be removed.

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,121 @@ public function cacheKeyComponent()
...
+                                $titleClasses,
+                                $modifiedTitle,
+                                $language,
+                                $title
+                            )
+                        )
+                    );
+
+                    // Set the default value to the element so we can compare it with JavaScript
+//                    if (Locale::getDefault() !== $locale) {
```
tractorcow: And this please :)

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,121 @@ public function cacheKeyComponent()
...
+                // Find field matching this translated field
+                // If the translated field has an ID suffix also check for the non-suffixed version
+                // E.g. UploadField()
+                $field = $fields->dataFieldByName($translatedField);
+                if (!$field && preg_match('/^(?<field>\w+)ID$/', $translatedField, $matches)) {
+                    $field = $fields->dataFieldByName($matches['field']);
+                }
+
+                // Highlight any translatable field
+                if ($field && !$field->hasClass('LocalisedField')) {
```
tractorcow: you can reduce nesting by inverting this.


```php
if (!$field || $field->hasClass('LocalisedField')) {
	continue;
}
```

raissanorth:
I removed the commented out code, because the changed fields are not relevant. Rather, the visual icons highlight the localised fields. In this updated version changes are made per row and not per field.

src/Extension/FluentExtension.php
raissanorth: If I were to make this change then the hasClass() function may be called on null.

lang/en.yml
```diff
@@ -1,6 +1,25 @@
 en:
+  FluentExtension:
+    php.TractorCow\Fluent\Extension\FluentExtensionDefaultLocale: 'This is the default locale'
+    php.TractorCow\Fluent\Extension\FluentExtensionDefaultLocaleIs: 'The default locale is {locale}'
+  FluentSiteTreeExtension:
+    php.TractorCow\Fluent\Extension\FluentSiteTreeExtensionLOCALESTATUSDRAFT: "A draft has been created for this locale, however, published content is still being inherited from\n                another locale. To publish this content for this locale, use the \"Save & publish\" action provided."
+    php.TractorCow\Fluent\Extension\FluentSiteTreeExtensionLOCALESTATUSINHERITED: "Content for this page is being inherited from another locale. If you wish you make an independent copy\n                of this page, please use one of the \"Copy\" actions provided."
```
robbieaverill: There's still some dodgy translation keys coming through here. Have a hunt for lines like `_t('FluentExtension....` which need to be switched to `_t(__CLASS__ ...`. I'm not sure where the `php.' . __CLASS__` keys are coming in but try and find them. This example here `php.TractorCow\Fluent\Extension\FluentSiteTreeExtensionLOCALESTATUSINHERITED` shouldn't have the `php.` prefix

src/Extension/FluentExtension.php
robbieaverill: If `!$field` evaluates to true, which it would if `$field` is null, then you wouldn't have that happen

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,106 @@ public function cacheKeyComponent()
...
+                    // Add a language indicator next to the fluent icon
+                    $title = $field->Title();
+
+                    $titleClasses = 'fluent-locale-label';
+                    $modifiedTitle = 'Using default locale value';
```
robbieaverill: Can we translate this please

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,106 @@ public function cacheKeyComponent()
...
+        }
+
+        // If the field is already present, don't add it a second time
+        if ($fields->fieldByName('CurrentLocaleMessage')) {
+            return $this;
+        }
+
+        $localeNames     = Locale::getCached()->map('Locale', 'Title')->toArray();
+        $defaultLocale   = Locale::getDefault();
+        $isDefaultLocale = ($defaultLocale === Locale::getCurrentLocale());
```
robbieaverill: You can use `$isDefaultLocale = $currentLocale->getIsDefault()` rather than performing this logic again

src/Extension/FluentExtension.php
```diff
@@ -835,4 +846,106 @@ public function cacheKeyComponent()
...
+        $defaultLocaleKey = $defaultLocale ? $defaultLocale->Locale : '';
+        $currentLocale = Locale::getCurrentLocale();
+        $currentLocaleKey = $currentLocale ? $currentLocale->Locale : '';
+        $messageClass    = ($isDefaultLocale) ? 'alert-success' : 'alert-info';
+        $message         = ($isDefaultLocale)
+            ? _t(__CLASS__.'DefaultLocale', 'This is the default locale')
+            : _t(
+                __CLASS__.'DefaultLocaleIs',
```
robbieaverill: Missing `.` before DefaultLocaleIs - this will evaluate to `TractorCow\Fluent\.....\...ExtensionDefaultLocale` - could be why you're seeing weird php. prefixes in the en.yml file

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -155,7 +155,7 @@ public function updateCMSFields(FieldList $fields)
         if (!FluentState::singleton()->getLocale()) {
             return;
         }
-
+        parent::updateCMSFields($fields);
```
robbieaverill: To some extent this makes the API a little less predictable, because the parent's method will not always be run. Maybe we should move the `FluentState::singleton()->getLocale` check into the parent class and just call `parent::updateCMSFields($fields)` in this class - what do you think?

robbieaverill: Scratch that - you'd need to return something more explicit from the parent method and check for it here. Just leave it =)

robbieaverill:
I've tested this out locally and here's my thoughts:

**1:** When you hover over the green localised icon next to a form field it says "Using default locale value", which was true in Fluent 3 but in Fluent 4 we'll be using this to indicate that the field is translatable - perhaps the message should say "Field is able to be translated"?

**2:** We're adding this indicator to FluentExtension so it should be applied to any DataObject that is "fluent" - this is working nicely for pages (e.g in your screenshot above), and works for DataObjects like SiteConfig pictured below:
image

But doesn't work for files and folders, which are essentially representing a DataObject that is rendered via the new React + form builder API which will slowly become increasingly more common. To make this approach work for that as well we'll need to add a `updateFormFields(FieldList $fields)` method where you do the same thing as `updateCMSFields()`.

If we want to re-use this logic for instances of FormFactory we may need to move the logic from FluentExtension that is concerned with presenting UI indicators in the CMS into a specific extension which only does that - I tried applying this to AssetFormFactory (parent of File and FolderFormFactory) but the fact that FluentExtension is trying to modify SQL queries and everything too is breaking it.

@tractorcow - how important do you think this is for the stable release? We could maybe skip this point (no. 2) for now, merge with everything else fixed up and refactor a little to make this possible afterwards?

src/Extension/FluentExtension.php
raissanorth: Of course, you are correct. That is exactly the reason why.

raissanorth:
@robbieaverill I like your first suggestion. Thanks for being so thorough. I'd be happy to change it to 'Field is able to be translated'

tractorcow:
> But doesn't work for files and folders, which are essentially representing a DataObject that is rendered via the new React + form builder API which will slowly become increasingly more common. To make this approach work for that as well we'll need to add a updateFormFields(FieldList $fields) method where you do the same thing as updateCMSFields().


If we implemented this feature using entirely CSS, could it work for those sections in asset-admin? It should use the same bootstrap styles.

robbieaverill:
Yeah that's right. The key considerations are that we need to push the LocalisedField (?) class into the form fields via the new API, and also ensure the fluent CSS is required

tractorcow:
I think the fluent indicator icon, at the very least, could be applied globally as a CSS only rule.

The '(en)' key could be added on forms that supported HTML labels.

tractorcow:
@robbieaverill that sounds like a really good solution. It can work with asset-admin too.

tractorcow:
> We don't really need to add the (en) parts to individual fields since we localise whole records now. I think you're right and we can just add a CSS class that adds the translation icon.

Not always whole records, but a subset of fields yes.

clarkepaul:
I've come up with some concepts which hopefully is going in the right direction (might need more work, let me know). Commenting on the mockups should also work if you like.
https://invis.io/ZBFVZJCKU38#/279729292_X_Fluent

robbieaverill:
There's some good feedback on this pull request, but I think it's going to be easier to close this and open new ones like #358 to address the updates in smaller increments

PR #348, opened 2018-02-09
FIX Set locale domain or base URL, and URL segment for locale as a page's URL segment

### LINKED ISSUE #344: URLSegment in CMS edit screen doesn't show localised URL, so its wrong
From #311 

cc @guyvanbael

>And also... the locale isn't added to the urlsegment, so the link is wrong.

schermafbeelding 2017-11-29 om 14 39 18

robbieaverill:
Resolves #344

### DIFF
--- a/src/Extension/FluentSiteTreeExtension.php
+++ b/src/Extension/FluentSiteTreeExtension.php
@@ -3,11 +3,15 @@
 namespace TractorCow\Fluent\Extension;
 
 use SilverStripe\CMS\Controllers\RootURLController;
+use SilverStripe\CMS\Forms\SiteTreeURLSegmentField;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
 use SilverStripe\Forms\CompositeField;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\LiteralField;
+use TractorCow\Fluent\Model\Domain;
+use TractorCow\Fluent\Model\Locale;
 use TractorCow\Fluent\State\FluentState;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Extension/FluentSiteTreeExtension.php
```diff
@@ -219,6 +224,50 @@ protected function addLocaleStatusMessage(FieldList $fields)
         );
     }
 
+    /**
+     * Add the locale's URLSegment to the URL prefix for a page's URL segment field
+     *
+     * @param FieldList $fields
+     * @return $this
+     */
+    protected function addLocalePrefixToUrlSegment(FieldList $fields)
```
tractorcow: Yeah this code must have been a pain to write!

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -219,6 +224,50 @@ protected function addLocaleStatusMessage(FieldList $fields)
...
+            ? $this->owner->Parent()->RelativeLink(true)
+            : null;
+
+        // $parentSegment could contain the locale's URL segment now. If we're in domain mode then remove it
+        if ($domain && substr($parentSegment, 0, strlen($localeUrlSegment)) === $localeUrlSegment) {
+            $parentSegment = ltrim(substr($parentSegment, strlen($localeUrlSegment)), '/');
+        }
+
+        // Join the base URL component (domain, or absolute URL with locale's URL segment) to any parent page segemnts
+        $baseUrl = Controller::join_links($baseUrl, $localeUrlSegment, $parentSegment);
```
tractorcow: We should join only either $baseURL and $localeURLSegment (if there is no parent), or $baseUrL and $parentSegment (if there is a parent)

tractorcow: See https://github.com/tractorcow/silverstripe-fluent/issues/367#issuecomment-368662704

PR #352, opened 2018-02-14
FIX Re-enable support for adding other locale links to Google sitemaps

### LINKED ISSUE #332: alternate language not working in ss4.0.1
Using ss4.0.1 the template which adds the alternate language isn't used. The default from googlesitemaps module is used. Copying the code over to the googlesitemaps template is a workaround.

robbieaverill:
Resolves #332 and replaces #336

Note: we may still need to add a rule somewhere to set the default locale for the sitemap.xml route to prevent the results being filtered by the current locale.

In that case you'd still not get non-default locale only pages in the default locale, but you can fix that by using domains.


templates/Wilr/GoogleSitemaps/Control/GoogleSitemapController_sitemap.ss
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type='text/xsl' href='$StylesheetPath'?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml">
+	<% loop $Items %>
+        <url>
+            <loc>$AbsoluteLink</loc>
+            <% if $LastEdited %><lastmod>$LastEdited.Format(y-MM-dd'T'HH:mm:ss)</lastmod><% end_if %>
```
tractorcow: Incorrect quote type. Template strings should be quoted.

PR #355, opened 2018-02-16
Fix bug: removing the entire table.

### LINKED ISSUE #354: Deleting a record removes the entire table.
https://github.com/tractorcow/silverstripe-fluent/blob/8e863f0de9d455bdfea5ce71471a41362940f7c8/src/Extension/FluentExtension.php#L433

This seems deletes the whole table while it's supposed to delete only one record?

ailionx:
Fixed #354

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -426,7 +426,8 @@ class FluentExtension extends DataExtension
 
             // Create root table delete
             $rootTable = $this->getDeleteTableTarget($table);
-            $rootDelete = SQLDelete::create("\"{$rootTable}\"");
+            $rootDelete = SQLDelete::create("\"{$rootTable}\"")
+                ->addWhere(["\"{$rootTable}\".\"ID\"" => $this->owner->ID]);
 
             // If table isn't localised, simple delete
             if (!isset($localisedTables[$table])) {


src/Extension/FluentExtension.php
```diff
@@ -430,7 +430,7 @@ public function updateDeleteTables(&$queriedTables)
 
             // If table isn't localised, simple delete
             if (!isset($localisedTables[$table])) {
-                $rootDelete->execute();
+                $rootDelete->addWhere(['"ID"' => $this->owner->ID])->execute();
```
tractorcow: Can you move the addWhere() up to outside of the loop? We use `$rootDelete` further down, and that also needs the ID filter. Please qualify the ID column also with the table name. :)

tractorcow:
```php
$rootDelete = SQLDelete::create("\"{$rootTable}\"")
	->addWhere(["\"{$rootTable}\".\"ID\"" => $this->owner->ID]);
```

NightJar: If possible, could you also please write a test for this fix?
That is, at test that deleting a non-localised record works as expected (as opposed to removing _all_ records)?

tractorcow:
Nope @robbieaverill current master fluent / 4.0 framework provide the right SQL for me.

```
FROM "SiteTree_Versions" LEFT JOIN "SiteTree_Localised_Versions" AS "SiteTree_Localised_en_US" 
```

tractorcow:
I think reading is ok after all... it's writing that's incorrect.

It's writing records to `SiteTree_Localised_Versions` but not to `SiteTree_Localised` or `SiteTree` on rollback.

src/Extension/FluentExtension.php
tractorcow: It's a bit of a problem actually... you need to test the record is deleted at the right time. If you delete this row, you need to delete the parent row too. However, THAT table could still be needed for other locales, so you can't just delete it then either.

The real fix is a pain. I'll write up something tomorrow.

tractorcow: Test:
 - create record in Locale A
 - create version in locale B
 - delete in locale B - base records exist, but _localised rows deleted
 - delete in locale A - both base and _localised records deleted (since it's the last locale).

ailionx: Thanks, guys. For now, I just updated the commit to lift the fix out of the loop as @tractorcow suggested.

tractorcow:
Other enhancements I can make later. :)

ailionx:
Thanks, @tractorcow for the help and merging.

PR #359, opened 2018-02-19
ENHANCEMENT Support fulltext search per locale

### LINKED ISSUE #335: fluent breaks fulltextsearch
configuring basic site search (by enabling fulltextsearchable) doesn't work when fluent module is 
installed. Here's the stack strace.
https://gist.github.com/guyvanbael/d23c189dbcc181bc50c377d6a38aac7f

NightJar:
In order to be able to search content, we must first be able to add it
to the index. To add the content in each locale (which can vary) we must
support a Variant, much like that of the `Versioned` module. Fluent
itself takes care of modifying the query to select content, we just must
allow the Reindex task to attain and adjust it's settings.

Resolves #335

### DIFF
--- a/src/Search/FluentSearchVariant.php
+++ b/src/Search/FluentSearchVariant.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace TractorCow\Fluent\Search;
+
+use TractorCow\Fluent\Extension\FluentExtension;
+use TractorCow\Fluent\State\FluentState;
+use TractorCow\Fluent\Model\Locale;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\FullTextSearch\Search\Variants\SearchVariant;
+use SilverStripe\FullTextSearch\Search\SearchIntrospection;
+use SilverStripe\FullTextSearch\Search\Queries\SearchQuery;
+
+if (!class_exists(SearchVariant::class)) {
+    return;
+}
+
+/**
+ * Allow FulltextSearch to index content from all configured locales, and support
+ * users searching to be restricted to their active locale. Works in a very similar
+ * way to SearchVariantVersioned from the FulltextSearch module (where that variant
+ * restricts searches to live pages, etc.)
+ */
+class FluentSearchVariant extends SearchVariant
+{
+    public function appliesTo($class, $includeSubclasses)
+    {
+        return $this->appliesToEnvironment() &&
+            SearchIntrospection::has_extension($class, FluentExtension::class, $includeSubclasses) &&
+            Locale::getCached()->count();
+    }
+
+    public function currentState()
+    {
+        return FluentState::singleton()->getLocale();
+    }
+
+    public function reindexStates()
+    {
+        return Locale::getCached()->column('Locale');
+    }
+
+    public function activateState($state)
+    {
+        $fluentState = FluentState::singleton()->setLocale($state);
+    }
+
+    public function alterQuery($query, $index)
+    {
+        if (FluentState::singleton()->getIsFrontend()) {
+            $query->filter('_locale', [
+                $this->currentState(),
+                SearchQuery::$missing
+            ]);
+        }
+    }
+
+    public function alterDefinition($class, $index)
+    {
+        $this->addFilterField($index, '_locale', [
+            'name' => '_locale',
+            'field' => '_locale',
+            'fullfield' => '_locale',
+            'base' => DataObject::getSchema()->baseDataClass($class),
+            'origin' => $class,
+            'type' => 'String',
+            'lookup_chain' => [
+                [
+                    'call' => 'variant',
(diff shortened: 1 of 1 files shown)


src/Search/FluentSearchVariant.php
```diff
@@ -0,0 +1,71 @@
...
+use SilverStripe\FullTextSearch\Search\SearchIntrospection;
+use SilverStripe\FullTextSearch\Search\Queries\SearchQuery;
+
+class FluentSearchVariant extends SearchVariant
+{
+
+    public function appliesToEnvironment()
+    {
+        return $this->config()->get('enabled');
+    }
```
robbieaverill: You could remove this, since `SearchVariant::appliesToEnvironment` is doing the same thing

src/Search/FluentSearchVariant.php
```diff
@@ -0,0 +1,71 @@
...
+
+use TractorCow\Fluent\Extension\FluentExtension;
+use TractorCow\Fluent\State\FluentState;
+use TractorCow\Fluent\Model\Locale;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\FullTextSearch\Search\Variants\SearchVariant;
+use SilverStripe\FullTextSearch\Search\SearchIntrospection;
+use SilverStripe\FullTextSearch\Search\Queries\SearchQuery;
+
+class FluentSearchVariant extends SearchVariant
```
robbieaverill: We need a check here to see if `SearchVariant` exists before declaring the class. It's not ideal but this will break without FTS installed otherwise

src/Search/FluentSearchVariant.php
```diff
@@ -0,0 +1,71 @@
...
+    public function appliesToEnvironment()
+    {
+        return $this->config()->get('enabled');
+    }
+
+    public function appliesTo($class, $includeSubclasses)
+    {
+        return $this->appliesToEnvironment() &&
+            SearchIntrospection::has_extension($class, FluentExtension::class, $includeSubclasses) &&
+            Locale::get()->count();
```
robbieaverill: If you use `Locale::getCached()->count()` it will minimise ORM calls

src/Search/FluentSearchVariant.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public function currentState()
+    {
+        return FluentState::singleton()->getLocale();
+    }
+
+    public function reindexStates()
+    {
+        return Locale::get()->column('Locale');
```
robbieaverill: Use `::getCached()`

src/Search/FluentSolrIndex.php
```diff
@@ -0,0 +1,18 @@
...
+
+namespace TractorCow\Fluent\Search\Solr;
+
+use SilverStripe\FullTextSearch\Solr\SolrIndex;
+use SilverStripe\CMS\Model\SiteTree;
+
+/**
+ * Default search index
+ */
+class FluentSolrIndex extends SolrIndex
```
robbieaverill: We don't really need this, since it is essentially a standard Solr index and is covered in the FTS documentation already

_config.php
```diff
@@ -1,2 +1,27 @@
...
+            // Allow storing the solr index and config data in an arbitrary location,
+            // e.g. outside of the webroot
+            'path' => Environment::getEnv('SOLR_INDEXSTORE_PATH')
+                ? Environment::getEnv('SOLR_INDEXSTORE_PATH')
+                : BASE_PATH . '/.solr',
+            'remotepath' => Environment::getEnv('SOLR_REMOTE_PATH')
+                ? Environment::getEnv('SOLR_REMOTE_PATH')
+                : null
+        ]
+    ]);
```
robbieaverill: As discussed in standup - remove this autoconfig logic

src/Search/FluentSearchVariant.php
```diff
@@ -0,0 +1,69 @@
...
+use SilverStripe\ORM\DataObject;
+use SilverStripe\FullTextSearch\Search\Variants\SearchVariant;
+use SilverStripe\FullTextSearch\Search\SearchIntrospection;
+use SilverStripe\FullTextSearch\Search\Queries\SearchQuery;
+
+if (!class_exists(SearchVariant::class)) {
+    return;
+}
+
+class FluentSearchVariant extends SearchVariant
```
robbieaverill: Maybe adding a doc block might be useful to explain what it does at a high level?

NightJar: Sounds ideal.

NightJar: Done :)

robbieaverill:
mog

NightJar:
54zdierl g3w
moggie, emerging green.

PR #361, opened 2018-02-19
ENHANCEMENT Update bootstrap alert 

raissanorth:
Changes were made to indicate that locale not visible until it is published, as well as the translations updated via the TextCollector task:

image

An 'invisible eye' icon still needs to be added at some stage.

Related to #311 

Reference: https://projects.invisionapp.com/share/ZBFVZJCKU38#/screens/279729292_X_Fluent

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -1,6 +1,14 @@
 en:
   TractorCow\Fluent\Control\LocaleAdmin:
     MENUTITLE: Locales
+  TractorCow\Fluent\Extension\FluentFilteredExtension:
+    FILTERED_LOCALES: 'Display in the following locales'
+    LOCALEFILTEREDHELP: 'This page is not visible in this locale'
+    LOCALEFILTEREDSHORT: Filtered
+  TractorCow\Fluent\Extension\FluentSiteTreeExtension:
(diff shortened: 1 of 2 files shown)


raissanorth:
I amended one of the unit tests, as the logic has changed.

chrispenny:
Hi team,

Why was the `LOCALESTATUSDRAFT` removed? If an object has implemented `$frontend_publish_required = false`, then that message is a legitimate possibility where the frontend content is being inherited from another Locale, but the CMS content has been drafted in the current Locale.

robbieaverill:
Hey @chrispenny - we made a UX decision to remove any additional SiteTree flags. In the situation you're describing I think it'd probably better to indicate this via some alert messages in the CMS edit form - what do you think?

chrispenny:
Yea, that's all good - this is the message that gets displayed at the Top of the CMS fields though (separate to the flags).

robbieaverill:
Ah I see. In that case it was probably unintentional

chrispenny:
No worries. If nobody beats me to it then I'll add it back in the next time I'm working in that space (not 100% sure when that will be though).

PR #362, opened 2018-02-20
NEW Add Fluent locale states as badges in CMS breadcrumbs

robbieaverill:
Part of the fluent UX updates from @clarkepaul and @sachajudd

This change requires SilverStripe 4.1.0, but Fluent will continue to work without this feature in earlier versions. Note that we may increase the minimum SilverStripe core requirement to 4.1.0 in the future if necessary.

### New page in default locale:

image

### New page in non-default locale:

image

### Localise existing page in another locale:

image

then:

image

#### Default locale:

image

#### Another locale without a localisation:

image

#### Another locale with a localisation:
image


NightJar:
Looks like that test might need updating, or being told to not care about case :>

robbieaverill:
Damn. Updated.

Issue #350, opened 2018-02-14
Installing fluent causes the default locale/site to disappear.


NightJar:
After installing fluent and setting up 2 locales:
- en_NZ (default: true)
- de_DE

the main site (`localhost/`) has disappeared. Furthermore I cannot get `localhost/about-us` nor `localhost/en_NZ/about-us` to load.
`localhost/de_DE` however loads, although it shows as the only page available (no other translations have been made for other pages, so I assume this is correct).

I may have something misconfigured - in which case the issue is less severe, but could end up being a common pitfall.

`tractorcow/silverstripe-fluent: 4.0.x-dev`
`commit ddf8c127328bb5d1f506a5e4b5b1bfdb99ebd1e9 (HEAD -> master, origin/master, origin/HEAD, composer/master)`

robbieaverill:
Yeah you need to publish those pages in each locale you want them to be visible in. We should make it easier if we can, to ensure the default 5 pages that you get when you install SilverStripe are available in the default locale at least

NightJar:
So this is a misunderstanding on my behalf. Fluent currently requires _all_ locales to be set up and published - the site as it was before installation is effectively a non-thing once a locale is created.

I'll expand on this description later :>

robbieaverill:
We could maybe also add a task that would publish all pages in the default locale

PR #370, opened 2018-03-07
Update documentation

### LINKED ISSUE #349: Editing a non-default local affects the default locale
While editing the Home page of my site in the "German" `locale` (where the default is "English (New Zealand)") I changed the title to "Home, lang:de", then clicked "Save and Publish".

This correctly updated `SiteTree_localised_live` to reflect the change `where Locale = 'de'`, however to my surprise then viewing the `SiteTree_live` table also revealed that `Title = 'Home, lang:de'`, which is somewhat both an erroneous state, and technically false.

This issue however did _not_ occur when editing the Content of the same page.

`tractorcow/silverstripe-fluent: 4.0.x-dev`
`commit ddf8c127328bb5d1f506a5e4b5b1bfdb99ebd1e9 (HEAD -> master, origin/master, origin/HEAD, composer/master)`

raissanorth:
The documentation changes alert users that editing non-default locales affects the SiteTree table and emphasise that localised content needs to be published in each locale one wants it to be visible. Fixes #349 and #350.

### DIFF
--- a/docs/en/configuration.md
+++ b/docs/en/configuration.md
@@ -24,10 +24,10 @@ Navigate to the `Fallbacks` tab, which allows you to specify one or more fallbac
 
 Once you add at least two locales to your site, you can begin localising your content. 
 
-***Important:** Pages in locales that fall back must be added and published in each locale you want them to be visible 
+_**Important:** Pages in locales that fall back must be added and published in each locale you want them to be visible 
 in - including the default locale. This essentially requires the re-publication of content in each locale, once content 
 is localised. Be aware that the site will not appear as it did before the creation of Fluent locales until this step is 
-completed.*
+completed._
 
 If desired, Fluent can be enabled on a field by field basis. Note that non-translated fields on any page will be 
 displayed in the default locale.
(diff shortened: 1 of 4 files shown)


docs/en/configuration.md
```diff
@@ -7,17 +7,30 @@ will interfere with locale bootstrapping in certain situations (such as `Securit
...
+**Note:** If using domains, you can additionally assign per-domain defaults as well.
+
  - `Domain`: Dropdown to assign this locale to a domain.
- - `Fallbacks`: Allows you to specify one or more fallback locales for this locale.
+ 
+Navigate to the `Fallbacks` tab, which allows you to specify one or more fallback locales for this locale.
+
+Once you add at least two locales to your site, you can begin localising your content. 
+
+***Important:** Pages in locales that fall back must be added and published in each locale you want them to be visible
```
robbieaverill:
```diff
- ***Important:**
+ **Important:**
```

raissanorth: As this point is rather important see #350, I intended to display this block in italics and bold. I am happy though to change it to something better suited - any ideas - to draw the necessary attention to it.

robbieaverill: Ah, I see - in that case you need to close it correctly (notice that the bold carries through). Maybe use an underscore for italic instead

docs/en/configuration.md
```diff
@@ -140,6 +153,9 @@ class FormPage extends Page
 }
 ```
 
+**Note:** Editing a non-default locale affects the `SiteTree(_live)` table. In contrast to SilverStripe 3, the
```
tractorcow: It doesn't matter if the locale is default or not. In fluent 4 vs 3, the "default" locale matters a lot less than it once did as far as content authoring is going.

Now all it means is that the default will be the first locale a user sees on the frontend.

raissanorth: Okay, I am going to change the phrasing slighlty. Let me know if that sounds better.

tractorcow:
Thanks @raissanorth great job :D

PR #374, opened 2018-03-18
Allow cookie domain and path to be configurable

ninty9notout:
We have two components to our websites: the marketing website which runs on [www.](https://www.royalalberthall.com/tickets/events/2018/nine-inch-nails/) and the purchase path which runs on [tickets.](https://tickets.royalalberthall.com/booking/production/bestavailable/60640)

The marketing site needs to set the cookie so that the tickets. domain has access to it, which is a problem because the module will only set the persist cookie on the current domain.

Two options to get around this is set cookie to be set on domain without the subdomain i.e. `.example.com` but that's not really required if we just allow the cookie domain to be set, which is option two.

The PR exposes two variables which will let the user specify the path using `persist_cookie_path` and domain using `persist_cookie_domain` using the config, falling back to `null` if not set (which is basically what the module currently does).

```yaml
TractorCow\Fluent\Middleware\DetectLocaleMiddleware:
  persist_cookie_domain: .wmc.site
```

image

### DIFF
--- a/src/Middleware/DetectLocaleMiddleware.php
+++ b/src/Middleware/DetectLocaleMiddleware.php
@@ -46,7 +46,7 @@ class DetectLocaleMiddleware implements HTTPMiddleware
      * Use this path when setting the locale cookie
      *
      * @config
-     * @var int
+     * @var string
      */
     private static $persist_cookie_path = null;
 
@@ -54,7 +54,7 @@ class DetectLocaleMiddleware implements HTTPMiddleware
      * Use this domain when setting the locale cookie
      *
      * @config
-     * @var int
+     * @var string
      */
     private static $persist_cookie_domain = null;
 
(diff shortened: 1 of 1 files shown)


src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -42,6 +42,22 @@ class DetectLocaleMiddleware implements HTTPMiddleware
      */
     private static $persist_cookie_expiry = 90;
 
+    /**
+     * Use this path when setting the locale cookie
+     *
+     * @config
+     * @var int
```
robbieaverill: Could you please use a doc block hint that describes what the property would be? E.g. here it's probably `@var string`

tractorcow: Yeah it's a string not an int isn't it?

ninty9notout:
@robbieaverill @tractorcow Classic case of fail copy/paste. Amended.

PR #381, opened 2018-03-26
FIX Cast default locale title as a DBField to prevent search field scaffolding error

### LINKED ISSUE #380: Uncaught Error: Call to a member function scaffoldSearchField() on string
Hi,

I just installed your plugin and went into locales section and after click on domain tab I got Error
      
    [Emergency] Uncaught Error: Call to a member function scaffoldSearchField() on string
    
    GET /admin/locales/TractorCow-Fluent-Model-Domain?flushtoken=13ea0aa5a90883c8efceb65e6aea60d5&flush=

    Line 2227 in C:\wamp64\www\xnet\vendor\silverstripe\framework\src\ORM\DataObject.php

Is it a bug or I miss something in installation steps ?

I am using SS v4.1.0

robbieaverill:
On SS 4.1.0 returning a string in this GridField summary field throws an exception. Instead, we should cast it as a DBField.

Resolves #380

### DIFF
--- a/src/Model/Domain.php
+++ b/src/Model/Domain.php
@@ -14,6 +14,7 @@ use SilverStripe\Forms\TabSet;
 use SilverStripe\Forms\TextField;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\HasManyList;
 use TractorCow\Fluent\State\FluentState;
 
@@ -139,13 +140,13 @@ class Domain extends DataObject
     }
 
     /**
-     * @return string
+     * @return DBField|null
      */
     public function DefaultLocaleTitle()
     {
         $locale = $this->DefaultLocale();
         if ($locale) {
-            return $locale->getTitle();
+            return DBField::create_field('Text', $locale->getTitle());
         }
         return null;
     }


tractorcow:
Ok, what I see is a change to `relObject` in DataObject 4.1 which has better support for "nulls along the way". If `DefaultLocale` is null, 4.0 will break on `null->$Title` if we use dot notation.

This fix isn't in 4.0, meaning that the old code needed this horrible "shiv" instead to support it.

What we should do is (4.1 only) use the preferred summary_fields dot-notation.

```
private static $summary_fields = [
    'Domain' => 'Domain',
    'DefaultLocale.Title' => 'Default Locale',
    'LocaleNames' => 'Locales',
];
```

Since we are supporting 4.0 for now, I approve of this fix since it does the job necessary. Just adding some background support for what went wrong. :)

src/Model/Domain.php
```diff
@@ -139,13 +140,13 @@ public function DefaultLocale()
...
     /**
-     * @return string
+     * @return DBField|null
      */
     public function DefaultLocaleTitle()
     {
         $locale = $this->DefaultLocale();
         if ($locale) {
-            return $locale->getTitle();
+            return DBField::create_field('Text', $locale->getTitle());
```
tractorcow: Could have also used `$locale->obj('Title')` but this is acceptable too. :)

PR #382, opened 2018-03-28
Add new `FluentBlackListFields` Extension

oilee80:
Used to stop the field value from being copied from one translation to another

Also updated readme.md with example code

### DIFF
--- a/docs/en/dataobjects.md
+++ b/docs/en/dataobjects.md
@@ -103,3 +103,21 @@ SilverStripe\Assets\File:
   translate:
     - Title
 ```
+
+### Fluent `has_one`s
+
+By default when you specify a related field to be `fluent`ed the same ID is used for both versions this can cause problems where you update the object and it is updated for all languages. To get around this you can use the `TractorCow\Fluent\Extension\FluentBlackListFields` Extension to specify explicit fields to not copy the value of when creating the new translation.
+An example configuration where you want separate MessageObjects for different languages may look like this:
+
+```php
+class MyObject extends DataObject
+{
+    private static $has_one = [
+        'Message' => 'MessageObject'
+    ];
+
+    private static $extensions = [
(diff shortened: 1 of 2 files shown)


oilee80:
Hi @robbieaverill,
Yes when it does the copy it copies the ID of the object as well which means all languages are pointing to the same object.

It makes more sense to make the relationship fluent than to have ALL the fields of the related object as being fluent (especially as the related object could have relationships as well)

Thanks for the pointer to the splat operator, not used it before

tractorcow:
The code does solve the problem, but it's a very specific problem that is hard to express as a generic feature. I feel like this belongs in user code, not as a part of the core module.

It's not really a "blacklist", what it is is a list of "fields that don't inherit from the base table", so it doesn't feel correctly labelled to me. It makes it harder to explain and justify it as a core feature.

Let me give you an example of a similar solution that feels more core-worthy.

Maybe you have `translate_defaults` config on each dataobject, and `FluentExtension::localiseSelect` will check the `translate_defaults` for a sql fragment / expression before falling back to `"\"{$table}\".\"{$field}\""`.

E.g.


```php
// Note: In CMS only we fall back to value in root table (in case not yet migrated)
        // On the frontend this row would have been filtered already (see augmentSQL logic)
        $sqlDefaults = $this->owner->config()->get('translate_defaults');
        $sqlDefault = isset($sqlDefaults[$field]) ? $sqlDefaults[$field] : "\"{$table}\".\"{$field}\"";
        $this->owner->invokeWithExtensions('updateLocaliseSelectDefault', $sqlDefault, $table, $field, $locale);
        $query .= "\tELSE $sqlDefault END\n";
```

And in your dataobject have something like

```
private static $translate_defaults = [
	'MessageID' => 'NULL',
];
```

On the other hand, I would probably still say you should be doing this logic in a user-code extension. :) You could implement the config there without too much complexity.

PR #383, opened 2018-03-28
Adds an Extension that will add a custom Default value when copying

oilee80:
Allows for a default value to be specified for fields
  also enables the Language Title to be passed in


tractorcow:
This is getting a bit closer to what I had suggested over at https://github.com/tractorcow/silverstripe-fluent/pull/382#issuecomment-377102653, but I will give the same answer that I'm not convinced this should be a core feature in it's current state.

I will however, help you if you get stuck with getting this implemented in your project.

Thanks for your work and taking the time to contribute. It's appreciated even if it didn't make it in. :)

PR #393, opened 2018-04-10
Update composer requirement constraints

NightJar:
Although inconsequential, the current constraints are all marked as dev - an overhang from when SilverStripe 4 was in pre-release stages. This is no longer necessary, so we can remove them in order to keep the constraints more in line with other modules (such as the requirements here) and easier to read.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "silverstripe/vendor-plugin": "^1",
         "silverstripe/framework": "^4",
         "silverstripe/cms": "^4",
-        "symbiote/silverstripe-gridfieldextensions": "^3"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -21,10 +21,10 @@
         "issues": "http://github.com/tractorcow/silverstripe-fluent/issues"
     },
     "require": {
-        "silverstripe/vendor-plugin": "^1.0",
-        "silverstripe/framework": "^4@dev",
-        "silverstripe/cms": "^4@dev",
-        "symbiote/silverstripe-gridfieldextensions": "^3.0.3@dev"
```
robbieaverill: This was a specific constraint, we should leave it there

PR #396, opened 2018-04-12
Change getLocalisedTables method to public.

chrispenny:
I have found that the `getLocalisedTables()` method is incredibly useful for some of the project specific *oddities* that we've been adding. Can we please make it public so that I don't have to duplicate the logic elsewhere?


tractorcow:
I guess! Will need to be a new minor version though. (4.1).

Need to set up some more branches before we merge this.

robbieaverill:
We have a 4.0 branch already - good to merge?

PR #408, opened 2018-04-26
ENHANCEMENT Shift extension default to yml file to promote better extensibility

### LINKED ISSUE #406: frontend_publish_required not working as expected
in my fluentconfig.yml I've set
```
Derralf\App\Model\TeaserBox:
  frontend_publish_required: false
```
TeaserBox is not versioned. 
Global default locale is en_US
SS 4.1 with Fluent ^4

If I understand correctly, with `frontend_publish_required` set to `false` my english (not yet translated) TeaserBoxes should display on my german and italian Pages as well with fallback contents.
But it will only display after localising it (open & save on german/italian page or modeladmin).

Only setting `frontend_publish_required` to false globally:
```
TractorCow\Fluent\Extension\FluentExtension:
  frontend_publish_required: false
```
...will display my non-localised teasers in the frontend, but will also affect Sitetree and all other fluent-extended Objects as well.

What am I doing wrong?
What did I miss?

tractorcow:
Fixes #406

Replacement solution for https://github.com/tractorcow/silverstripe-fluent/pull/407

Instead of adding the config via extension (which cannot be easily overridden on a model by model basis via yml) add the default config via yml.

@dhensby I think this might be a better convention going forward; Promote use of yml for configs, rather than adding private statics to extension classes. What is your feeling on this?

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -907,14 +907,6 @@ class FluentExtension extends DataExtension
             return false;
         }
 
-        // Check legacy (deprecated) config
-        $frontendPublishRequired = $this->owner->config()->get('frontend_publish_required');
-        if (isset($frontendPublishRequired)) {
-            Deprecation::notification_version('5.0.0', 'Use ignore_frontend_publish=true instead');
-            return (bool)$frontendPublishRequired;
-        }
-
-        // Require saved unless ignoring frontend published
-        return !$this->owner->config()->get('ignore_frontend_publish');
+        return $this->owner->config()->get('frontend_publish_required');
     }
 }
--- a/_config/model.yml
+++ b/_config/model.yml
@@ -1,0 +1,6 @@
+---
+Name: fluentmodel
+---
+SilverStripe\ORM\DataObject:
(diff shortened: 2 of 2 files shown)


src/Extension/FluentExtension.php
```diff
@@ -907,14 +907,6 @@ protected function requireSavedInLocale()
             return false;
         }
 
-        // Check legacy (deprecated) config
-        $frontendPublishRequired = $this->owner->config()->get('frontend_publish_required');
-        if (isset($frontendPublishRequired)) {
-            Deprecation::notification_version('5.0.0', 'Use ignore_frontend_publish=true instead');
```
tractorcow: I'm un-deprecating this since we haven't had a release yet. :)

robbieaverill:
I like this new approach. It'll remove some PHP code (private statics for config) in a way that might bring SilverStripe further towards being modern :)

dhensby:
> @dhensby I think this might be a better convention going forward; Promote use of yml for configs, rather than adding private statics to extension classes. What is your feeling on this?

agreed

PR #415, opened 2018-05-18
WIP: FIX Implement localisable order by clause to support translatable sorting

### LINKED ISSUE #208: sort() is not working on translated fields
Sorting of translated fields is not working:

`$this->Tags()->sort('Title', 'ASC');`
Does only work on default locale, because sort is ALWAYS sorting Tags on default locale instead of current locale.

```
$Loc_TitleField = Fluent::db_field_for_locale('Title', $this->CurrentLocale());
$this->Tags()->sort($Loc_TitleField, 'ASC');
```

Is not working, because Tags that have no textual difference in translation, are `null` in the database and therefore mess up the sort.

robbieaverill:
Resolves #208

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -967,7 +967,7 @@ class FluentExtension extends DataExtension
 
             // Check if this field is in any of the tables, and just pick any that match
             foreach ($tables as $table => $fields) {
-                if (in_array($field, $fields)){
+                if (in_array($field, $fields)) {
                     return [$table, $field, false];
                 }
             }
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -7,9 +7,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
-use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\FieldList;
-use SilverStripe\Forms\LiteralField;
 use SilverStripe\i18n\i18n;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataExtension;
@@ -333,7 +331,7 @@ class FluentExtension extends DataExtension
                     "\"{$table}\".\"ID\" = \"{$joinAlias}\".\"RecordID\" AND \"{$joinAlias}\".\"Locale\" = ?",
                     $joinAlias,
                     20,
-                    [ $joinLocale->Locale ]
+                    [$joinLocale->Locale]
                 );
             }
         }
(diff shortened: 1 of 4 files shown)


robbieaverill:
I've implemented a rough looking fix to allow localisable sort conditions in the same way that localisable where conditions work at the moment. There's a test failure in `testWritesToCurrentLocale` which looks to be pre-existing (tested locally on 4.1, re-running the last 4.1 build on Travis now).

@tractorcow would be good to get your eyes over this when you're back from your holiday =)

robbieaverill:
Ok so the tests I added in the first commit broke the `testWritesToCurrentLocale` test - I'll work out why and push an update

src/Extension/FluentExtension.php
```diff
@@ -389,9 +389,59 @@ public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null)
...
+            // Look for table context
+            if (preg_match('/^"(?<table>[\w\\\\]+)"\."(?<field>\w+)"$/i', $column, $matches)) {
+                $field = $matches['field'];
+            } else {
+                // If no table context is given, see if we can find one in the select
+                // e.g. "Title" -> "SomeTable"."Title"
+                $foundInConditionSearch = false;
+                foreach ($conditionSearch as $conditionSearchOption) {
+                    if (preg_match(
+                        '/^"(?<table>[\w\\\\]+)"\."(?<field>' . trim($column, '"') . ')"$/i',
```
tractorcow: If $column fails the above regexp, I think assuming that it's a basic column name is a bit risky. It could be a non-trivial sub-query, or it could be a composite sort. The `else` should really be another regexp that is similar but doesn't require the table prefix.

robbieaverill: Ok, do you have a couple of examples that I could add to the tests?

tractorcow: like `$list->sort('"Title", "Content"')`

robbieaverill: Sweet. We can't really make subqueries fluent in this case though right...?

tractorcow: Right so you should skip them.

robbieaverill: I've added a bunch more tests, the last commit contains some tests that I've got no idea how to write logic for to make them pass 😆 

Also I'm a bit iffy about some of the tests in the "Add more sorting tests" commit because they're relying on the DB to return natural sort orders when the sorting isn't specific enough to always return a definite order - the fixtures have common values in them (deliberately) but maybe we either need to remove the common values to ensure that there's a predictable sort order at all times.

tractorcow:
Pushed up some fixes @robbieaverill

tractorcow:
Pushed up linting fixes @robbieaverill but there are still some small issues to fix.

chrispenny:
I'm late to the party on this one, but please accept my delayed:
Thank you! I appreciate your work (as always)!

lerni:
If you experience this issue with SS 3.x theres a composer installable patch https://github.com/lerni/silverstripe3-mysql57-fluent

PR #416, opened 2018-05-22
Reversing the position of haystack and needle

ninty9notout:
**Issue 1**

This fixes the issue where the browser may send a generic language header such as `en` or `cy` as opposed to the whole `en-GB` or `cy-GB`.

In that case we end up looking for `cy-GB` inside `cy` which `stripos` doesn't like. Flipping params fixes this issue.

**Issue 2**

Due to [legal reasons](https://en.wikipedia.org/wiki/Welsh_Language_Act_1993), no one language can be considered default. Fluent will currently return the first locale in the list regardless of the default flag being set or not. Introducing the `disable_implicit_default` flag to not have it do this allows some flexibility around this. By default, Fluent will behave as it currently does unless the flag is explicitly set to `true`.

**Issue 3**

Once the correct domain has been determined, there is no need to prefix the locale to URLs as Fluent will serve the page in the correct locale thereafter. Introducing the `prefix_locale_in_link` flag will allow us to disable this globally. Also it's really useful for outputting the canonical URL without the prefix. Again, by default Fluent will continue to behave as it currently does unless the flag is explicitly set to `false`.

### DIFF
--- a/src/Extension/FluentBadgeExtension.php
+++ b/src/Extension/FluentBadgeExtension.php
@@ -8,9 +8,18 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\FieldType\DBField;
 use TractorCow\Fluent\Model\Locale;
 use TractorCow\Fluent\State\FluentState;
+use SilverStripe\Core\Config\Configurable;
 
 class FluentBadgeExtension extends Extension
 {
+    use Configurable;
+
+    /**
+     * @config
+     * @var boolean
+     */
+    private static $show_full_locale_badge = false;
+
     /**
      * Push a badge to indicate the language that owns the current item
      *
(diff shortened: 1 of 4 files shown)


ninty9notout:
@robbieaverill @tractorcow For your consideration.

tractorcow:
Am in China will let robbie handle this 😁

robbieaverill:
Are you able to get the tests passing to start with?

ninty9notout:
@robbieaverill There's an issue for this which I'm currently fixing. Please hold off from reviewing for a bit.

ninty9notout:
Okay, that has been issue amended, and a new feature has been added.

**Issue with `FluentExtension::getSourceLocale`**

Doesn't seem like `SourceLocale` was being populated correctly when fetching a record from the database which existed in one locale but not in the current locale. We're adding a fallback to use `Locale` instead, and then finally falling back to using the default locale.

**Feature to show full locale title in the badge**

`FluentBadgeExtension::getBadge` will only output the suffix of a domain which is fine in most cases, however, there are locales where the suffix is the same. Staying without our Welsh example, both `cy_GB` and `en_GB` will show "GB" as the badge. Pretty bad UX you can imagine.

The best way around this is to provide support to show the full title instead of just the suffix. `show_full_locale_badge` has been introduce to do just this:

image

As with the other flags, Fluent will by default show only the suffix unless the flag is explicitly set to `true`.

clarkepaul:
@robbieaverill so do these changes keep the current functionality but provides devs the ability to set it to show a full language eg. "en_GB"? If so, I've got no problems with this.

robbieaverill:
Closing due to lack of activity. Feel free to reopen if you'd like to revive this pull request @ninty9notout =) Thanks!

PR #417, opened 2018-05-23
Quote table and fields in isLocalisedInStage for postgres compatibility

blasic:
Request to /admin/pages/treeview fails with "sitetree_localised" relation not found on postgres. 
This PR quotes everything to make work.


tractorcow:
Oh good find.

We probably should merge this into older branches too to bugfix.

robbieaverill:
@tractorcow yep, I'll cherry pick it backwards during the CWP 2.1 release

PR #422, opened 2018-06-06
BUG Fix invalid locale being set for domain mode

### LINKED ISSUE #419: Domain setup triggers invalid locale error
Potentially stumbled across an issue during some experimenting that was also reported by another user in the #translations channel in SilverStripe Users slack earlier in the month.

After attaching a unique domain to each of my locales, ensuring there is default locale for each domain and lastly setting `SS_FLUENT_FORCE_DOMAIN=true`,  I am greeted by the following error on the frontend on all but home page: `[Emergency] Uncaught InvalidArgumentException: Invalid "TractorCow\Fluent\Model\Locale" locale.`.

While the home page loads ok, it only uses the global default locale regardless of the domain.

Configuration is all default but I did try setting `disable_default_prefix` to `true` to see if that helped at all.

Everything works as expected when no domains are in play.

```
silverstripe/cms 4.1.0
silverstripe/framework 4.1.0
tractorcow/silverstripe-fluent 4.1.0 
```

tractorcow:
Fixes #419 

Merge up to 4.1 and so on after merge.

### DIFF
--- a/src/State/FluentState.php
+++ b/src/State/FluentState.php
@@ -2,6 +2,7 @@
 
 namespace TractorCow\Fluent\State;
 
+use InvalidArgumentException;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Injector\Injector;
 
@@ -20,9 +21,9 @@ class FluentState
     protected $locale;
 
     /**
-     * Current domain
+     * Current domain, if set
      *
-     * @var string
+     * @var string|null
      */
     protected $domain;
 
@@ -58,6 +59,9 @@ class FluentState
      */
     public function setLocale($locale)
     {
+        if (empty($locale) || !is_string($locale)) {
+            throw new InvalidArgumentException("Invalid locale");
+        }
         $this->locale = $locale;
         return $this;
     }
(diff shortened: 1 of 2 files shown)


robbieaverill:
Need to target 4.0 branch in order to merge up to 4.1?

src/State/FluentState.php
```diff
@@ -75,11 +79,14 @@ public function getDomain()
     /**
      * Set the current domain code
      *
-     * @param string $domain
+     * @param string|null $domain
      * @return $this
```
robbieaverill: Can we add some PHPDocs for the exceptions this throws?

tractorcow: We don't really do that in core; It could be a standard, but I'd probably wait until we did it consistently before adding it to other modules.

And then, I'd do it everywhere of course. :)

robbieaverill: Oh? Ok 😆

robbieaverill:
Re-target at 4.0 branch if needed before merge

tractorcow:
oh oops.

tractorcow:
Fixed.

robbieaverill:
@tractorcow I'll merge up now

PR #429, opened 2018-06-18
FIX Only force change for written records when not in current locale, or not versioned

robbieaverill:
The existing logic would always change a record when it is saved. The updated logic checks for a valid current locale and that the record doesn't already exist in the current locale before forcing change on it.

Also if the record isn't versioned then we always force a change.

Doing some testing with this change now but it appears to fix the issue.

Fixes https://github.com/silverstripe/cwp/issues/102


tractorcow:
A much more precise fix than I had imagined. Good job.

PR #442, opened 2018-07-19
Opt-out of cookie and session locale persistence (fixes #440)

### LINKED ISSUE #440: Fluent sends cookies by default, preventing HTTP caching in most layers
`Set-Cookie: FluentLocale=deleted` gets added to a default response (e.g. in the CWP 2.x recipe). That's killing caching, since most CDNs won't cache if cookie values are present, regardless what `Cache-Control` says. This might sound like a small thing, but cachability is a major focus of the upcoming CWP 2.1 release, so I consider this a high impact bug.

chillu:
It kills any HTTP caching otherwise. The distinction between `is_null` and "empty string" isn't ideal of course.

See https://github.com/tractorcow/silverstripe-fluent/issues/440

### DIFF
--- a/docs/en/configuration.md
+++ b/docs/en/configuration.md
@@ -240,19 +240,17 @@ Will cascade through different checks in order:
 1. Routing path (e.g. `/de/ueber-uns`)
 2. Request variable (e.g. `ueber-uns?FluentLocale=de`)
 3. Domain (e.g. `http://example.de/ueber-uns`)
-4. Session (if `DetectLocaleMiddleware.persist_session` is configured)
+4. Session (if a session is already started)
 5. Cookie (if `DetectLocaleMiddleware.persist_cookie` is configured)
 6. Request headers (if `FluentDirectorExtension.detect_locale` is configured)
 
-Additionally, detected locales will be set in session and cookies.
-This behaviour can be configured through `DetectLocaleMiddleware.persist_session`
-and `DetectLocaleMiddleware.persist_cookie`. To solely rely on stateless
-request data (routing path, request variable or domain),
-configure as follows:
+Additionally, detected locales will be set in cookies.
+This behaviour can be configured through `DetectLocaleMiddleware.persist_cookie`.
+To solely rely on sessions and stateless request data (routing path, request
+variable or domain), configure as follows:
 
 ```yaml
 TractorCow\Fluent\Middleware\DetectLocaleMiddleware:
-  persist_session: false
   persist_cookie: false
 ```
 
--- a/docs/en/configuration.md
+++ b/docs/en/configuration.md
@@ -231,9 +231,33 @@ visible locales for this object.
 Note: Although these objects will be filtered in the front end, this filter is disabled
 in the CMS in order to allow access by site administrators in all locales.
 
-## Locale detection
+## Routing and Locale Detection
 
-When a visitor lands on the home page for the first time, Fluent can attempt to detect that user's locale based
+The `DetectLocaleMiddleware` will detect if a locale has been requested (or is default) and is not the current
+locale, and will redirect the user to that locale if needed.
+
+Will cascade through different checks in order:
+1. Routing path (e.g. `/de/ueber-uns`)
+2. Request variable (e.g. `ueber-uns?FluentLocale=de`)
+3. Domain (e.g. `http://example.de/ueber-uns`)
+4. Session (if `DetectLocaleMiddleware.persist_session` is configured)
+5. Cookie (if `DetectLocaleMiddleware.persist_cookie` is configured)
+6. Request headers (if `FluentDirectorExtension.detect_locale` is configured)
+
+Additionally, detected locales will be set in session and cookies.
+This behaviour can be configured through `DetectLocaleMiddleware.persist_session`
+and `DetectLocaleMiddleware.persist_cookie`. To solely rely on stateless
+request data (routing path, request variable or domain),
+configure as follows:
+
+```yaml
+TractorCow\Fluent\Middleware\DetectLocaleMiddleware:
+  persist_session: false
+  persist_cookie: false
+```
+
+When a visitor lands on the home page for the first time,
+Fluent can also attempt to detect that user's locale based
 on the `Accept-Language` http headers sent.
 
 This functionality can interfere with certain applications, such as Facebook Open Graph tools, so it
--- a/docs/en/configuration.md
+++ b/docs/en/configuration.md
@@ -244,15 +244,20 @@ Will cascade through different checks in order:
 5. Cookie (if `DetectLocaleMiddleware.persist_cookie` is configured)
 6. Request headers (if `FluentDirectorExtension.detect_locale` is configured)
 
-Additionally, detected locales will be set in cookies.
-This behaviour can be configured through `DetectLocaleMiddleware.persist_cookie`.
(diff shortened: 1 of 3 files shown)


chillu:
OK, pivoted a bit on this - fluent detects locales from all kinds of markers, all rolled into one monolithic middleware. Since most browsers send some kind of marker by default (`Accept-Language`), it'll *always* set a cookie and session value, so the "skip cookie if null locale" case never applies.

I think assuming *stateful* rather than *stateless* requests (with cookies and sessions) is a bad default, but we can't change that in this release line due to semver. So instead, I'm introducing the ability to turn off this behaviour. Which then allows devs to make their sites cacheable, e.g. in the default CWP recipe. I'll also add a note to the CWP performance guide about this.

chillu:
@robbieaverill @tractorcow I can't reproduce the test failures in https://travis-ci.org/tractorcow/silverstripe-fluent/jobs/406075939 locally, when reproducing the Travis build steps. The only difference I can see is that I'm running PHP 7.2 rather than 7.0

robbieaverill:
Those test failures are unrelated to this PR, they were part of the in progress work for allowing localised sorting with #415

chillu:
OK can someone please PR and merge then? Without this patch, you can't get caching to work when you have the CWP default recipe installed (which includes fluent). Even when you don't use fluent. I don't think this needs to make it into the CWP 2.1.0 recipe, but it would be good to have a stable release out that CWP devs could upgrade to (and we can reference in the performance guide)

tractorcow:
Just to throw a spanner into the works, persist_session = false will break SOME things in the CMS. Do you want to make this conditional on frontend/backend? Otherwise you could make it conditional on whether a session is started or not. I.e. use a session if started, but avoid starting one. That would get you a lot better performance than a fixed on/off.

tractorcow:
As an aside, "don't start a session" should be the default anyway, meaning it's a zero-config fix. :)

src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -140,11 +164,11 @@ protected function getPersistLocale(HTTPRequest $request)
         }
 
         // check session then cookies
-        if ($locale = $request->getSession()->get($key)) {
+        if (static::config()->get('persist_session') && $locale = $request->getSession()->get($key)) {
```
robbieaverill: I think because we're assigning `$locale` a value here that we need to wrap it in its own set of brackets to ensure the logic is executed correctly

tests/php/Middleware/DetectLocaleMiddlewareTest.php
```diff
@@ -109,4 +119,122 @@ public function testLocaleIsAlwaysPersistedEvenIfNotSetByTheMiddleware()
...
+        $middleware->process($request, function () {
+            // no-op
+        });
+    }
+
+    public function testLocaleIsPersistedFromCookie()
+    {
+        $newLocale = 'fr_FR';
+        $middleware = $this->middleware;
+        $key = $this->middleware->getPersistKey();
```
robbieaverill: Might as well use `$middleware` here

tests/php/Middleware/DetectLocaleMiddlewareTest.php
```diff
@@ -109,4 +119,122 @@ public function testLocaleIsAlwaysPersistedEvenIfNotSetByTheMiddleware()
...
+        // TODO PHPUnit's headers_sent() always returns true, so we can't check for cookie values.
+        // PHPUnit has process isolation for this purpose, but we can't use it because autoloading breaks.
+        // $this->assertEquals($newLocale, Cookie::get($key));
+    }
+
+    public function testLocaleIsPersistedFromSession()
+    {
+        $newLocale = 'fr_FR';
+        $middleware = $this->middleware;
+        $key = $this->middleware->getPersistKey();
```
robbieaverill: Minor: use `$middleware`

robbieaverill:
These changes look fine to me. I'll divert to @tractorcow and @chillu to resolve the comments above about starting sessions

frankmullenger:
@chillu @tractorcow ability to prevent sessions auto starting on the frontend is important for our current project, any possibility we could start the conversation up again? :-)

tractorcow:
The request was to remove the need for added configs, and we only set the session variable and use the logic I requested above.

> I.e. use a session if started, but avoid starting one. That would get you a lot better performance than a fixed on/off.

`if (static::config()->get('persist_session'))` becomes `if ($session->isStarted() && $session->getAll())`.

Can we do that instead please?

tractorcow:
I guess I should step in and fix this huh. Sorry been busy.

robbieaverill:
I've pushed a commit which removes the `persist_session` config prop and checks whether the session is started before persisted reads or writes happen. I hope this is what you wanted @tractorcow !

If not, please feel free to rebase my commit out again =D

ScopeyNZ:
Can we really remove a config setting like this in a patch?

docs/en/configuration.md
```diff
@@ -231,9 +231,31 @@ visible locales for this object.
...
+2. Request variable (e.g. `ueber-uns?FluentLocale=de`)
+3. Domain (e.g. `http://example.de/ueber-uns`)
+4. Session (if a session is already started)
+5. Cookie (if `DetectLocaleMiddleware.persist_cookie` is configured)
+6. Request headers (if `FluentDirectorExtension.detect_locale` is configured)
+
+Additionally, detected locales will be set in cookies.
+This behaviour can be configured through `DetectLocaleMiddleware.persist_cookie`.
+To solely rely on sessions and stateless request data (routing path, request
+variable or domain), configure as follows:
```
chillu: @robbieaverill Technically, this is no longer accurate advice. To *store* in sessions, you have to force a session to be started *before* this logic kicks in. Either in the same request (through higher priority middleware), or more commonly through an explicit `Session::start()` in your own logic. It's a bit weird to rely on something outside of this module logic flow to trigger this behaviour, and it can lead to side effects. If you naively rely on sessions to store language, you might end up diagnosing issues where sometimes sessions aren't started, and language selections "don't stick". Can you expand on that in the docs?

robbieaverill: Cool, I've pushed an updated commit - I hope it represents what you mean 😃

ScopeyNZ:
I think this looks good to merge. It will not start a session now which seems to address @tractorcow's feedback. We can make incremental PRs if needed 🙂

src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -171,14 +188,17 @@ protected function setPersistLocale(HTTPRequest $request, $locale)
         }
 
         // Save locale
-        if ($locale) {
-            $request->getSession()->set($key, $locale);
-        } else {
-            $request->getSession()->clear($key);
+        $session = $request->getSession();
+        if ($session->isStarted()) {
```
tractorcow: Missing ->getAll() check; A started session isn't valid if it's empty. I can add this in. It was in my pasted code.

tractorcow: All sessions are started on every request, by the way.

tractorcow: PR at https://github.com/tractorcow-farm/silverstripe-fluent/pull/476

PR #443, opened 2018-07-19
v4: ENHANCEMENT Update CMS localisation status messages.

chrispenny:
Added back a message that was accidentally removed in #361.

When `frontend_publish_required` is set to false, there is the possibility that content in being inherited from another locale.

We could add further checks for content inheritance to make these message more specific, but I thought that would be overkill. A simple "may/may-not" message felt sufficient.

Not published, not drafted:
screen shot 2018-07-20 at 8 42 32 am

Not published, drafted:
screen shot 2018-07-20 at 8 40 20 am

Added a docblock to `FallbackLocale`.

### DIFF
--- a/src/Extension/FluentSiteTreeExtension.php
+++ b/src/Extension/FluentSiteTreeExtension.php
@@ -183,22 +183,51 @@ class FluentSiteTreeExtension extends FluentVersionedExtension
             return;
         }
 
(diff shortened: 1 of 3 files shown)


chrispenny:
Added unit tests to cover the different messages.

PR #445, opened 2018-07-19
v4: Proposal to add config setting for CMS publish required.

chrispenny:
We currently have this logic sitting in project code through an extension, and I'm more than happy for it to stay there, but...

## The situation
Current project relies heavily on `DataObject`s that are managed through a 3rd party system - there is no ability to modify these `DataObject`s in SilverStripe, but they can be viewed and they can be added as relationships on other `DataObject`s.

As such, we have a requirement to hide un-localised `DataObject`s not only on the frontend, but in the CMS as well (so that authors don't add unavailable `DataObject`s as relationships).

## Proposal
Add a config value for `cms_publish_required` so that devs can separately control frontend and CMS visibility options.

### DIFF
--- a/tests/php/Extension/FluentSiteTreeExtensionTest.php
+++ b/tests/php/Extension/FluentSiteTreeExtensionTest.php
@@ -10,6 +10,7 @@ use SilverStripe\Core\Config\Config;
 use SilverStripe\Dev\SapphireTest;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\DataObject;
 use SilverStripe\View\ArrayData;
 use TractorCow\Fluent\Extension\FluentDirectorExtension;
 use TractorCow\Fluent\Extension\FluentSiteTreeExtension;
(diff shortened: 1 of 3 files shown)


robbieaverill:
I guess this could make Fluent more like Translatable (entirely separate site trees) if devs wanted it, right?

chrispenny:
@robbieaverill that certainly could be another use for this.

I'm still trying to finalise my UX proposal (didn't get any time last hackday), but I think this could also work well with the "transfer to locale" dropdown:
https://github.com/tractorcow/silverstripe-fluent/issues/444

tests/php/Extension/FluentSiteTreeExtensionTest.php
```diff
@@ -255,6 +255,24 @@ public function testAddLocalePrefixToUrlSegment($localeCode, $fixture, $expected
...
+        FluentState::singleton()->setIsFrontend(true);
+
+        $page = Page::get()->filter('URLSegment', 'home')->first();
+
+        $this->assertNull($page);
+    }
+
+    public function testHomeVisibleInCMS()
+    {
+        FluentState::singleton()->setIsFrontend(false);
```
robbieaverill: I think these tests would be more verbose if we were explicitly setting the config props you're adding that affect these results. Would you mind adding them?

chrispenny: Thanks, @robbieaverill . Just to check that I'm on the right track here, you mean doing something like this?

```
    public function testHomeVisibleInCMS()
    {
        Config::nest();

        Config::modify()->set(DataObject::class, 'cms_publish_required', false)

        ... perform tests

        Config::unnest();
    }
```

robbieaverill: Yep exactly only you don't need the nesting

chrispenny:
Sorry this took me so long, @robbieaverill. I've also added some more coverage to the unit tests just to make me feel a bit more secure about these two configs not mixing purposes/results.

PR #446, opened 2018-07-19
v4: Remove Fluent status flag text.

chrispenny:
We still want the status flag, but as discussed in the UX Issues/PRs, core does not support multiple status flag texts being present.

I think this one was just missed when we removed the others.


ScopeyNZ:
Hey Chris. Happy to get this merged but as it's a bugfix could you switch the branch to 4.1 and rebase.

If you could also edit that commit message to have a prefix of "FIX" that would be peachy 🙂

chrispenny:
Hi @ScopeyNZ,

This should be sorted now. The failing test is not from my changes.

robbieaverill:
We have some intermittent test failures now from localised sorting changes =( I've restarted the builds

PR #450, opened 2018-08-01
Fix deprecated filter method in SearchVariant

sanderha:


ScopeyNZ:
This is a patch fix so can you rebase and target 4.1?

sanderha:
Not sure how to do that, not too experienced with doing pull requests ;)
I've tried to target the 4.1 branch now, is that okay or no?

ScopeyNZ:
I fixed it for you 🙂

robbieaverill:
Since we don't mandate fulltextsearch as a requirement we can't enforce that developers use version 3.3 or newer which is where this change was made. This would be a semver violation without backwards compatibility, so I've made a PR to support older versions too in https://github.com/tractorcow/silverstripe-fluent/pull/451

PR #452, opened 2018-08-14
[FIX] Ensure the default locale is set correctly for domain based configuration

scott1702:
Fixes https://github.com/tractorcow/silverstripe-fluent/issues/439

### DIFF
--- a/src/Extension/FluentDirectorExtension.php
+++ b/src/Extension/FluentDirectorExtension.php
@@ -2,8 +2,12 @@
 
 namespace TractorCow\Fluent\Extension;
 
+use Exception;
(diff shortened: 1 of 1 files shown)


robbieaverill:
You OK with this @tractorcow?

robbieaverill:
Context: it sucks that the director doesn't apply middleware until after it's done all the routing, but it is what it is. This is why we need to use the director extension instead of making that a middleware as well. @scott1702 found that the InitStateMiddleware needs to be run before the fluent director extension makes routing decisions based on the state, so this is the workaround.

robbieaverill:
@tractorcow we can follow up on this if you're not happy with it

PR #463, opened 2018-09-06
v4: Added config setting to bypass filtered locale logic in stage. Fixes #461

### LINKED ISSUE #461: v4: Fluent Filtered applied even in stage=Stage.
We have some time soon to implement a solution, but I'd like to know what the preference is.

## The problem

You have a project where `frontend_publish_required` is set to `false`, as you want to be able to display Pages on the frontend using inherited content.

`FluentFilteredExtension` is then used to control when you do or don't want to see a particular Page in a particular Locale (sometimes you **don't** want to display inherited content on the frontend for a particular Locale).

The issue is when you have purposefully not added a Filtered Locale, but you want to start drafting the Page in that Locale. Because `FluentFilteredExtension::augmentSQL()` is applied anytime you are viewing the Frontend (regardless or reading mode), you cannot use the "preview" panel, or view it on the frontend with `stage=Stage`. You are forced to add a Filtered Locale which would expose the "inheriting content" version of the Page to the frontend.

## Solution

I believe that `FluentFilteredExtension::augmentSQL()` should **not** be applied when reading mode is `DRAFT`.

### Force this outcome

I suspect that this might actually be the desired outcome for all projects - if so, we'll just add the reading mode check to `augmentSQL` all the time.

### Config setting

Let everyone decide whether it's required or not, and we'll add a new config value for `apply_filtered_locales_to_stage` (or something like that).

chrispenny:
I wasn't too sure if we wanted the default set to `true` or `false`. Having it as `true` by default leaves functionality the same as it is atm. Maybe next major release we could change that.

### DIFF
--- a/src/Extension/FluentFilteredExtension.php
+++ b/src/Extension/FluentFilteredExtension.php
@@ -12,6 +12,7 @@ use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataQuery;
 use SilverStripe\ORM\Queries\SQLSelect;
+use SilverStripe\Versioned\Versioned;
 use TractorCow\Fluent\Model\Locale;
 use TractorCow\Fluent\State\FluentState;
 
@@ -115,6 +116,11 @@ class FluentFilteredExtension extends DataExtension
             return;
         }
 
+        // Dev has requested that we do not apply Filtered Locales when reading mode is set to DRAFT.
+        if (!$this->owner->config()->get('apply_filtered_locales_to_stage') && $this->getModeIsStage()) {
+            return;
+        }
+
         $table = $this->owner->baseTable();
         $filteredLocalesTable = $table . '_' . self::SUFFIX;
 
@@ -147,5 +153,23 @@ class FluentFilteredExtension extends DataExtension
         }
 
         return null;
+    }
+
+    /**
+     * There are two different DRAFT modes. One when browsing stage, and one when browsing archive. Both modes have
+     * "Stage" at the very end of their reading_mode name.
+     *
+     * @return bool
+     */
+    private function getModeIsStage()
+    {
+        $str = Versioned::get_reading_mode();
+        $test = Versioned::DRAFT;
+
+        if (strlen($str) === 0) {
+            $str = Versioned::DEFAULT_MODE;
+        }
+
+        return substr_compare($str, $test, strlen($str) - strlen($test), strlen($test)) === 0;
     }
 }
--- a/src/Extension/FluentFilteredExtension.php
+++ b/src/Extension/FluentFilteredExtension.php
@@ -161,15 +161,15 @@ class FluentFilteredExtension extends DataExtension
      *
      * @return bool
      */
-    private function getModeIsStage()
+    protected function getModeIsStage()
     {
-        $str = Versioned::get_reading_mode();
-        $test = Versioned::DRAFT;
+        $readingMode = Versioned::get_reading_mode();
+        $draft = Versioned::DRAFT;
 
-        if (strlen($str) === 0) {
-            $str = Versioned::DEFAULT_MODE;
+        if (strlen($readingMode) === 0) {
+            $readingMode = Versioned::DEFAULT_MODE;
         }
 
-        return substr_compare($str, $test, strlen($str) - strlen($test), strlen($test)) === 0;
+        return substr_compare($readingMode, $draft, strlen($readingMode) - strlen($draft), strlen($draft)) === 0;
     }
 }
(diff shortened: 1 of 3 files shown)


chrispenny:
Ah, `withVersionedMode()` is not available in lower versions. I will update those tests.

chrispenny:
I'm not sure the failing `php 7` test is related to my changes?
```
1) TractorCow\Fluent\Tests\Extension\FluentExtensionTest::testLocalisedFieldsCanBeSorted with data set "german ascending on unlocalised field" ('de_DE', array('"Description"'), array('Lesen Sie mehr', 'Rennen', 'Eine Akte'))
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'Lesen Sie mehr'
-    1 => 'Rennen'
+    0 => 'Rennen'
+    1 => 'Lesen Sie mehr'
     2 => 'Eine Akte'
 )
```

robbieaverill:
Yeah the test failures are intermittent at the moment

src/Extension/FluentFilteredExtension.php
```diff
@@ -148,4 +154,22 @@ protected function getDataQueryLocale(DataQuery $dataQuery = null)
...
         return null;
     }
+
+    /**
+     * There are two different DRAFT modes. One when browsing stage, and one when browsing archive. Both modes have
+     * "Stage" at the very end of their reading_mode name.
+     *
+     * @return bool
+     */
+    private function getModeIsStage()
```
robbieaverill: We always prefer protected over private in module code

src/Extension/FluentFilteredExtension.php
```diff
@@ -148,4 +154,22 @@ protected function getDataQueryLocale(DataQuery $dataQuery = null)
...
+    {
+        $str = Versioned::get_reading_mode();
+        $test = Versioned::DRAFT;
```
robbieaverill: I'm not a big fan of this variable naming, if I'm going to be nit-picky

src/Extension/FluentFilteredExtension.php
```diff
@@ -148,4 +154,22 @@ protected function getDataQueryLocale(DataQuery $dataQuery = null)
...
+
+        if (strlen($str) === 0) {
+            $str = Versioned::DEFAULT_MODE;
+        }
+
+        return substr_compare($str, $test, strlen($str) - strlen($test), strlen($test)) === 0;
```
robbieaverill: Is there a reason we can't use `Versioned::get_stage()` or a combination of that and `Versioned::current_archived_stage()`? I'd really like to try and use Versioned API methods here if possible

chrispenny: I did this because `current_archived_stage` returns a value of `Stage` if you are browsing the `Live` site.

screen shot 2018-09-07 at 6 57 34 am

`get_reading_mode` is consistent though - you get `stage.Stage` or `Archive.[date].Stage`. This seemed like the most succinct/consistent way to check if I'm browsing in `Stage` in or outside of Archive.

tests/php/Extension/FluentFilteredExtensionTest.php
```diff
@@ -29,27 +32,99 @@ protected function setUp()
...
+        $currentStage = Versioned::get_stage();
+
+        Versioned::set_stage(Versioned::DRAFT);
+
+        FluentState::singleton()->withState(function (FluentState $newState) {
+            $newState
+                ->setLocale('en_NZ')
+                ->setIsFrontend(true);
+
+            $this->assertEquals(1, SiteTree::get()->count());
```
robbieaverill: We can use `$this->assertCount(1, SiteTree::get())`

tests/php/Extension/FluentFilteredExtensionTest.php
```diff
@@ -29,27 +32,99 @@ protected function setUp()
...
+        $currentStage = Versioned::get_stage();
+
+        Versioned::set_stage(Versioned::LIVE);
+
+        FluentState::singleton()->withState(function (FluentState $newState) {
+            $newState
+                ->setLocale('en_NZ')
+                ->setIsFrontend(true);
+
+            $this->assertEquals(0, SiteTree::get()->count());
```
robbieaverill: `$this->assertCount(0, SiteTree::get())`

tests/php/Extension/FluentFilteredExtensionTest.php
```diff
@@ -29,27 +32,99 @@ protected function setUp()
...
+
+        Versioned::set_stage(Versioned::DRAFT);
+
+        // Run test
+        FluentState::singleton()->withState(function (FluentState $newState) {
+            $newState
+                ->setLocale('en_NZ')
+                ->setIsFrontend(true);
+
+            $this->assertEquals(2, SiteTree::get()->count());
```
robbieaverill: `$this->assertCount(2, SiteTree::get())`

chrispenny:
Thanks, @robbieaverill . I've made those changes to the tests.

PR #464, opened 2018-09-13
ENHANCEMENT: Update unit tests to use withState where appropriate. Resolves #462

### LINKED ISSUE #462: v4: Unit tests should (probably) use withState() closure
I can pick this up next hackday (next Friday) - feel free to put my face on it.

We should (probably) update unit tests to use `withState()`, EG:
```
FluentState::singleton()->withState(function (FluentState $newState) {
    $newState
        ->setLocale('en_NZ')
        ->setIsFrontend(false);

    ... test
});
```

It keeps everything predictable.

chrispenny:
Failing tests are the sorting ones we're already aware of.

Desired outcome is that nothing has changed. Using `withState` just keeps everything nicely in a bubble.

Resolves #462

### DIFF
--- a/tests/php/Extension/FluentBadgeExtensionTest.php
+++ b/tests/php/Extension/FluentBadgeExtensionTest.php
@@ -44,12 +44,14 @@ class FluentBadgeExtensionTest extends SapphireTest
         // Clear cache
         Locale::clearCached();
 
-        FluentState::singleton()->setLocale('en_NZ');
+        FluentState::singleton()->withState(function (FluentState $newState) {
+            $newState->setLocale('en_NZ');
 
-        $this->mockPage = $this->objFromFixture(SiteTree::class, 'test_page');
-        $this->mockController = new FluentStubController($this->mockPage->ID);
-        $this->extension = new FluentBadgeExtension();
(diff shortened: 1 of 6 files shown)


tests/php/Extension/FluentBadgeExtensionTest.php
```diff
@@ -44,12 +44,14 @@ protected function setUp()
...
 
-        $this->mockPage = $this->objFromFixture(SiteTree::class, 'test_page');
-        $this->mockController = new FluentStubController($this->mockPage->ID);
-        $this->extension = new FluentBadgeExtension();
-        $this->extension->setOwner($this->mockController);
+            $this->mockPage = $this->objFromFixture(SiteTree::class, 'test_page');
+            $this->mockController = new FluentStubController($this->mockPage->ID);
+            $this->extension = new FluentBadgeExtension();
+            $this->extension->setOwner($this->mockController);
+        });
```
robbieaverill: My only thought here is that this would mean that the state would be reset back to what it was after this line (before any test actually runs). Perhaps the setUp method should be left the way it was? The tests are green though so it'd only be a semantic/code comprehension issue

Issue #454, opened 2018-08-16
Localised sorting tests are failing intermittently


robbieaverill:
Some of the unit tests for sorting across locales are failing. These were introduced in #415.

We need to investigate and fix so they're consistent.

```
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.
.............FFF................................................. 65 / 90 ( 72%)
.........................                                         90 / 90 (100%)
Time: 43.43 seconds, Memory: 360.00MB
There were 3 failures:
1) TractorCow\Fluent\Tests\Extension\FluentExtensionTest::testLocalisedFieldsCanBeSorted with data set "english descending on unlocalised field" ('en_US', array('"Description" DESC'), array('A record', 'Read about things', 'Go for a run'))
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => 'A record'
-    1 => 'Read about things'
-    2 => 'Go for a run'
+    1 => 'Go for a run'
+    2 => 'Read about things'
 )
/home/travis/build/tractorcow/silverstripe-fluent/tests/php/Extension/FluentExtensionTest.php:179
/home/travis/build/tractorcow/silverstripe-fluent/vendor/phpunit/phpunit/phpunit:52
2) TractorCow\Fluent\Tests\Extension\FluentExtensionTest::testLocalisedFieldsCanBeSorted with data set "german ascending on unlocalised field" ('de_DE', array('"Description"'), array('Lesen Sie mehr', 'Rennen', 'Eine Akte'))
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'Lesen Sie mehr'
-    1 => 'Rennen'
+    0 => 'Rennen'
+    1 => 'Lesen Sie mehr'
     2 => 'Eine Akte'
 )
/home/travis/build/tractorcow/silverstripe-fluent/tests/php/Extension/FluentExtensionTest.php:179
/home/travis/build/tractorcow/silverstripe-fluent/vendor/phpunit/phpunit/phpunit:52
3) TractorCow\Fluent\Tests\Extension\FluentExtensionTest::testLocalisedFieldsCanBeSorted with data set "german descending on unlocalised field" ('de_DE', array('"Description" DESC'), array('Eine Akte', 'Lesen Sie mehr', 'Rennen'))
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
[...]

PR #466, opened 2018-09-17
Mismatching _FilteredLocales table when using FluentFilteredExtension

jakxnz:
https://github.com/tractorcow/silverstripe-fluent/issues/465

Adjusted the `_FilteredLocales` table reference to check all tables in the object's ancestry.

I notice that `dev/master` has a different commit history to `4.1`. Our current project uses `4.1` so I've raised it against that branch.


jakxnz:
Many many through sort failures... mumble, mumble...

robbieaverill:
Unrelated, will re-run them (#454)

PR #468, opened 2018-09-21
FIX Opt-in performance fix for many consecutive lookups using isLocalisedInStage

### LINKED ISSUE #467: Allow caching and prepopulation for isLocalisedInStage
`FluentVersionedExtension::isLocalisedInStage()` is heavily used in the CMS, e.g. when rendering tree items.

In order to cut down on the query count on large sites, this value should be able to be cached and the cache prepopulated, as we do for page versions, and as I have recommended for Hierarchy::numChildren() in https://github.com/silverstripe/silverstripe-framework/issues/8379.

ScopeyNZ:
This PR adds a static method to pre-populate a cache of object IDs that exist in the given locale. This can be opted into by calling this method from appropriate user-code (or user-code extension) or alternatively by providing a list of classes that should be pre-populated (done when the first call of `isLocalisedInStage` is performed for that class).

Ideally I wouldn't have this config option but it is currently quite difficult to find a location to prepopulate this cache when running the SiteTree, although I'm open to suggestions here and more than happy to remove the configuration part of this PR.

Fixes #467

### DIFF
--- a/src/Extension/FluentVersionedExtension.php
+++ b/src/Extension/FluentVersionedExtension.php
@@ -6,6 +6,7 @@ use InvalidArgumentException;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataQuery;
 use SilverStripe\ORM\DB;
@@ -369,7 +370,7 @@ class FluentVersionedExtension extends FluentExtension
      * @param array $options A map of hints about what should be cached. "numChildrenMethod" and
      *                       "childrenMethod" are allowed keys.
      */
-    public function onPrepopulateTreeDataCache(DataList $recordList = null, array $options = [])
+    public function onPrepopulateTreeDataCache($recordList = null, array $options = [])
     {
         if (!Config::inst()->get(self::class, 'prepopulate_localecontent_cache')) {
             return;
@@ -388,7 +389,9 @@ class FluentVersionedExtension extends FluentExtension
      * Prepopulate the cache of IDs in a locale, to optimise batch calls to isLocalisedInStage.
      *
      * @param string $locale
-     * @param string $dataObject
+     * @param $dataObjectClass
+     * @param bool $populateLive
+     * @param bool $populateDraft
      */
     public static function prepoulateIdsInLocale($locale, $dataObjectClass, $populateLive = true, $populateDraft = true)
     {
--- a/src/Extension/FluentVersionedExtension.php
+++ b/src/Extension/FluentVersionedExtension.php
@@ -3,8 +3,13 @@
 namespace TractorCow\Fluent\Extension;
 
 use InvalidArgumentException;
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataQuery;
 use SilverStripe\ORM\DB;
+use SilverStripe\ORM\Queries\SQLExpression;
 use SilverStripe\ORM\Queries\SQLSelect;
 use SilverStripe\Versioned\Versioned;
 use TractorCow\Fluent\Model\Locale;
@@ -59,6 +64,21 @@ class FluentVersionedExtension extends FluentExtension
      * @var array
      */
     protected $localisedStageCache = [];
+
+    /**
+     * Array of objectIds keyed by table (ie. stage) and locale. Indicates which object IDs exist in which locales.
+     *
+     * static::$idsInLocaleCache[ $locale ][ $table(.self::SUFFIX_LIVE) ][ $objectId ] === true
+     *
+     * @var string[][][]
+     */
+    protected static $idsInLocaleCache = [];
+
+    /**
+     * @config
+     * @var array
+     */
+    private static $prepopulate_locale_object_types = [];
 
     protected function augmentDatabaseDontRequire($localisedTable)
     {
@@ -297,11 +317,21 @@ class FluentVersionedExtension extends FluentExtension
             }
         }
 
+        $ownerClass = get_class($this->owner);
+        $prepopulateClasses = Config::inst()->get(self::class, 'prepopulate_locale_object_types');
+        if (is_array($prepopulateClasses) && in_array($ownerClass, $prepopulateClasses)) {
+            self::prepoulateIdsInLocale($locale, $ownerClass);
+        }
+
         // Get table
         $baseTable = $this->owner->baseTable();
         $table = $this->getLocalisedTable($baseTable);
         if ($stage === Versioned::LIVE) {
             $table .= self::SUFFIX_LIVE;
+        }
+
+        if (isset(self::$idsInLocaleCache[$locale][$table])) {
+            return in_array($this->owner->ID, self::$idsInLocaleCache[$locale][$table]);
         }
 
         // Check cache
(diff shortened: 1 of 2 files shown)


sminnee:
Let's implement https://github.com/silverstripe/silverstripe-framework/issues/8391 and then refactor this PR to provide a `onPrepopulateTreeDataCache()` implementation on the FluentSiteTreeExtension.

sminnee:
This should probably be committed to master / 4.2.x-dev

sminnee:
A patch for this is here https://gist.github.com/bec52b2d0f50a42f44a0aea2a05bc2af

If you can give me commit rights to the creative-commoners repo I'll push it to this branch

src/Extension/FluentVersionedExtension.php
```diff
@@ -329,4 +357,69 @@ public function flushCache()
...
+     * @param array $options A map of hints about what should be cached. "numChildrenMethod" and
+     *                       "childrenMethod" are allowed keys.
+     */
+    public function onPrepopulateTreeDataCache(DataList $recordList = null, array $options = [])
+    {
+        if (!Config::inst()->get(static::class, 'prepopulate_localecontent_cache')) {
+            return;
+        }
+
+        // This functionality hasn't been implemented; it's not actually used right now, so assumeYAGNI
```
robbieaverill: Can we clarify this comment a bit? The end doesn't make sense to me

ScopeyNZ:
I have updated the comment. I just translated the acronym 🙂

robbieaverill:
Ok so is this PR dependant on https://github.com/silverstripe/silverstripe-framework/pull/8380? If so, we'll need to prepare a new minor release which has a 4.3 minimum in core, since that PR was merged into 4.3.x-dev

ScopeyNZ:
Right. I can rebase this onto `4` in the morning...

sminnee:
@robbieaverill @ScopeyNZ this doesn't _require_ 4.3+, but it's a no-op until 4.3 is installed. I don't think the constraint change is required then?

robbieaverill:
OK sweet

robbieaverill:
In that case there's no rebase required. It'd be cool to get Travis going before merge (I've pinged Damian), but if they pass locally then go for it

sminnee:
I wonder if @tractorcow wants to invite some commoners to help till the farm?

src/Extension/FluentVersionedExtension.php
```diff
@@ -304,21 +327,26 @@ protected function isLocalisedInStage($stage, $locale = null)
             $table .= self::SUFFIX_LIVE;
         }
 
+        if (isset(self::$idsInLocaleCache[$locale][$table])) {
+            return in_array($this->owner->ID, self::$idsInLocaleCache[$locale][$table]);
```
maxime-rainville: Checking if an array key exists is a lot faster than checking if a value is present in array. Might be worthwhile restructuring `$idsInLocaleCache` to do that instead.

src/Extension/FluentVersionedExtension.php
```diff
@@ -329,4 +357,70 @@ public function flushCache()
...
+     * Prepopulate the cache of IDs in a locale, to optimise batch calls to isLocalisedInStage.
+     *
+     * @param string $locale
+     * @param string $dataObject
+     */
+    public static function prepoulateIdsInLocale($locale, $dataObjectClass, $populateLive = true, $populateDraft = true)
+    {
+        // Get the table for the given DataObject class
+        $self = new static();
+        $table = $self->getLocalisedTable(Injector::inst()->get($dataObjectClass)->baseTable());
```
maxime-rainville: Wouldn't you be better off getting a singleton of `$dataObjectClass` and calling `getLocalisedTable` on that.

```php
$singleton = DataObject::singleton($dataObjectClass);
$table = $singleton->getLocalisedTable();
```

maxime-rainville:
Probably need to write some unit test for this as well.

src/Extension/FluentVersionedExtension.php
```diff
@@ -304,21 +327,26 @@ protected function isLocalisedInStage($stage, $locale = null)
...
             '"RecordID"' => $this->owner->ID,
             '"Locale"' => $locale,
         ]);
-        $query->firstRow();
-        $result = $query->execute()->value() !== null;
+
+        $result = $query->firstRow()->execute()->value() !== null;
 
         // Set cache
         $this->localisedStageCache[$key] = $result;
```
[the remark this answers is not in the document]
ScopeyNZ: The only reason I didn't do this is the static cache only knows about records that are localised - but it knows about all of them, so if it's missing then we can reliably say it is _not_ localised in the given stage. With the fallback, it caches either `true` or `false`. I can make the other one static too which should provide a little boost.

src/Extension/FluentVersionedExtension.php
ScopeyNZ: Good call! I'll change this

src/Extension/FluentVersionedExtension.php
ScopeyNZ: I can use the `singleton` method :+1: - I'll also grab a singleton of the extension instead... I don't think the `DataObjectSchema` method will give the same result as there's some additional logic in `FluentExtension::getLocalisedTable`?

src/Extension/FluentVersionedExtension.php
```diff
@@ -329,4 +357,70 @@ public function flushCache()
...
     }
+
+    /**
+     * Hook into {@link Hierarchy::prepopulateTreeDataCache}.
+     *
+     * @param DataList $recordList The list of records to prepopulate caches for. Null for all records.
+     * @param array $options A map of hints about what should be cached. "numChildrenMethod" and
+     *                       "childrenMethod" are allowed keys.
+     */
+    public function onPrepopulateTreeDataCache(DataList $recordList = null, array $options = [])
```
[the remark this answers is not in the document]
robbieaverill: If it accepts an array, won't the DataList type hint throw an error when you give it an array?

src/Extension/FluentVersionedExtension.php
```diff
@@ -304,29 +331,97 @@ protected function isLocalisedInStage($stage, $locale = null)
...
+
+        self::prepoulateIdsInLocale(FluentState::singleton()->getLocale(), $this->owner->baseClass());
+    }
+
+    /**
+     * Prepopulate the cache of IDs in a locale, to optimise batch calls to isLocalisedInStage.
+     *
+     * @param string $locale
+     * @param string $dataObject
+     */
```
robbieaverill: There's a couple of new arguments to add here now

src/Extension/FluentVersionedExtension.php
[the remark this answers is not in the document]
ScopeyNZ: Oh, apparently I didn't change this bit 🤦‍♂️

src/Extension/FluentVersionedExtension.php
```diff
@@ -304,29 +332,99 @@ protected function isLocalisedInStage($stage, $locale = null)
             $table .= self::SUFFIX_LIVE;
         }
 
+        if (isset(static::$idsInLocaleCache[$locale][$table])) {
```
robbieaverill: I feel like this isset check should include the owner's ID like the line below

robbieaverill:
I've done a little bit of a refactor in order to unit test this easily, and have pushed some new tests. LGTM. I think @tractorcow's feedback has been addressed

robbieaverill:
We don't have Travis set up on this org yet (cc @tractorcow)

@tractorcow if you don't think we've correctly addressed your comments feel free to ping us and we'll look more into it =)

src/Extension/FluentVersionedExtension.php
ScopeyNZ: The reason I didn't is because we know all the IDs that exist in the locale for that table - the absence of an ID implies it's not there. So items that _aren't_ in that stage/locale will skip this cache if you isset check the ID too.

PR #475, opened 2018-09-28
FIX: Add _complete flag to ensure that missing record requests are ca…

sminnee:
…ched

FIX: Remove unnecessary 2nd cache implementation.

This code improves performance when there are a significant number of
draft pages. Related, there was a duplicate implementation of caching
that wasn’t really necessary.

### DIFF
--- a/src/Extension/FluentVersionedExtension.php
+++ b/src/Extension/FluentVersionedExtension.php
@@ -57,13 +57,6 @@ class FluentVersionedExtension extends FluentExtension
     ];
 
     /**
-     * Cache of published status of this record
-     *
-     * @var array
-     */
-    protected static $localisedStageCache = [];
-
-    /**
      * Array of objectIds keyed by table (ie. stage) and locale. This knows ALL object IDs that exist in the given table
      * and locale.
      *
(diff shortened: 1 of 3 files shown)


sminnee:
FYI @ScopeyNZ found this wee issue missed.

ScopeyNZ:
Haha we have options now...

https://github.com/tractorcow-farm/silverstripe-fluent/pull/474

ScopeyNZ:
Tests are passing 😉

src/Extension/FluentVersionedExtension.php
```diff
@@ -331,17 +324,15 @@ protected function isLocalisedInStage($stage, $locale = null)
...
         if (isset(static::$idsInLocaleCache[$locale][$table][$this->owner->ID])) {
-            return isset(static::$idsInLocaleCache[$locale][$table][$this->owner->ID]);
-        }
+            return true;
 
-        // Check cache from local instance calls
-        $key = $table . '/' . $locale . '/' . $this->owner->ID;
-        if (isset(static::$localisedStageCache[$key])) {
-            return static::$localisedStageCache[$key];
+        } elseif (!empty(static::$idsInLocaleCache[$locale][$table]['_complete'])) {
```
robbieaverill: `elseif` adds unnecessary complexity over just `if` since you're returning before this

tests/php/Extension/FluentVersionedExtensionTest.php
```diff
@@ -80,15 +80,15 @@ public function testLocalisedStageCacheIsUsedForIsLocalisedInLocale()
...
 
         // Initial request
         $result = $extension->isPublishedInLocale('en_NZ');
-        $this->assertSame('foo', $result, 'Original method result is returned');
+        $this->assertSame(true, $result, 'Original method result is returned');
 
         // Checking the cache
         $result2 = $extension->isPublishedInLocale('en_NZ');
-        $this->assertSame('foo', $result2, 'Cached result is returned');
+        $this->assertSame(true, $result2, 'Cached result is returned');
```
robbieaverill: I realise the original wasn't representative of the API, but I'd used it so there was no chance of confusion about whether the mock was being used. The update is fine but just thought I'd note why I wrote it that way =)

PR #476, opened 2018-10-01
BUG Prevent force-creating empty sessions

tractorcow:
Reasoning is that all sessions are always started (isStarted() is true after SessionMiddleware is run), but not all sessions eventuate into a session in finalise. This check catches that.

### DIFF
--- a/_config/tests.yml
+++ b/_config/tests.yml
@@ -1,0 +1,8 @@
+---
+Name: fluentsapphiretest
+---
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\Dev\State\SapphireTestState:
+    properties:
+      States:
+        fluent: '%$TractorCow\Fluent\Dev\FluentTestState'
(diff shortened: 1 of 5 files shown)


tractorcow:
Just re-added travis to the new organisation.

chillu:
Good catch Damian, I was wondering about that in peer review but then ignored it - I shouldn't have.

There's a test failure:

> 2) TractorCow\Fluent\Tests\Middleware\DetectLocaleMiddlewareTest::testLocaleIsPersistedFromSession
SilverStripe\Control\Session::isStarted() was not expected to be called more than 2 times.

ScopeyNZ:
> Good catch Damian, I was wondering about that in peer review but then ignored it - I shouldn't have.
> 
> There's a test failure:
> 
> > 1. TractorCow\Fluent\Tests\Middleware\DetectLocaleMiddlewareTest::testLocaleIsPersistedFromSession
> >    SilverStripe\Control\Session::isStarted() was not expected to be called more than 2 times.

This might be my fault - I forgot to check the PR before I merged it earlier today and it may have already been failing 😧

PR #477, opened 2018-10-02
Adding TestState to clear static locale caching between tests

ScopeyNZ:
This resolves an issue where tests (in a "key project") were failing due to some locale caching with various locale configuration. This test state busts the locale caching between every test (as it probably should do)


robbieaverill:
I've cherry picked this into #476 to get the tests green, cheers =)

PR #478, opened 2018-10-03
existsInLocale() returns true when hitting the cache

wernerkrauss:
even when a page does not exist in that locale


robbieaverill:
It was already green 😆

Issue #486, opened 2018-10-29
Generate default base url with default locale's url segment appended


jakxnz:
# Requirements
```
silverstripe/cms                                       4.3.x-dev d44e761
tractorcow/silverstripe-fluent                         4.1.x-dev d5a278e
```

# Details

I have a case where it is undesirable to default to `/` when querying the base url of the default locale.

# Steps to recreate

1. Use `Locale::getDefault()->getBaseURL()`

**Current result**

Returns a base url that excludes the default locale's url segment

**Desired result**

The ability to return a base url that includes the default locale's url segment

jakxnz:
https://github.com/tractorcow-farm/silverstripe-fluent/pull/487

PR #487, opened 2018-10-29
Display the current domain's default locale url segment in base urls

jakxnz:
* Added `use_full_default_base_url` config to Locale
* Added tests to cover new default behaviour

https://github.com/tractorcow-farm/silverstripe-fluent/issues/486

### DIFF
--- a/docs/en/configuration.md
+++ b/docs/en/configuration.md
@@ -270,3 +270,12 @@ is turned off by default. To turn it on set the below setting:
 TractorCow\Fluent\Extension\FluentDirectorExtension:
   detect_locale: true
 ```
+
+## Use full default base URL for all locales
+
+By default, fluent will return `/` as the default locale's default base url. To enforce the use of full default base URLs for all locales (e.g always return `/en/`), set the below setting:
+
+```yaml
+TractorCow\Fluent\Model\Locale:
+  use_full_default_base_url: true
+```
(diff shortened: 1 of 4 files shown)


tractorcow:
We already have `disable_default_prefix`. Shouldn't you just use this?

tractorcow:
Full phpdoc.

```php
    /**
     * Allow the prefix for the default {@link Locale} (IsDefault = 1) to be disabled
     *
     * @config
     * @var bool
     */
    private static $disable_default_prefix = false;
```

It looks like it should have the prefix by default. Locale::getBaseURL() just ignores that for some reason.

ScopeyNZ:
> It looks like it should have the prefix by default. Locale::getBaseURL() just ignores that for some reason.

🤔 So it's intended to be there but hasn't been there for as long as `getBaseURL` has existed? So it's a "bugfix" technically - we make it honour the existing config setting. Can we really justify it in a patch change?

ScopeyNZ:
We could make the new config setting enable `getBaseURL` to consider the original `disable_default_prefix` config but it's feeling a bit clunky at this point 😖

jakxnz:
@ScopeyNZ you can take a look at the change set here https://github.com/tractorcow-farm/silverstripe-fluent/pull/488, and see if it feels ok

PR #488, opened 2018-10-30
BUGFIX: Display the current domain's default locale url segment in base urls

jakxnz:
* Added check to locale base url generation
* Added tests to cover new default behaviour
* Docs altered

## Warning
This may be a breaking change

https://github.com/tractorcow-farm/silverstripe-fluent/issues/486


tractorcow:
To me this feels like a bugfix. Just want another person to validate the thinking so we can avoid the risk of a semver breakage.

PR #503, opened 2019-02-05
FIX: Fluent rules block other controllers from defining rules on the locale

UndefinedOffset:
Fluent's rules are appended to the beginning of Director's rules list this means that the rule that handles routing to `ModelAsController` trumps all other rules in the site.

For example if we're using the slug "us" for the "en_US" locale and we want to have a controller with a rule "en/search//$Action/$ID/$OtherID" that directs to our own controller. Rather than our own controller as expected handling the request `ModelAsController` handles the request instead. This is due to the order of the rules for Fluent when they are added to the rules array in `FluentDirectorExtension`.

This pull request changes the way the rules are merged so the `RootURLController` rules and the `ModelAsController` rules appear just before the `$URLSegment//$Action/$ID/$OtherID` rule. Having the root url's at the top seems to break the model as controller rules, it could be something to do with the rule itself but I'm not sure.

### DIFF
--- a/src/Extension/FluentDirectorExtension.php
+++ b/src/Extension/FluentDirectorExtension.php
@@ -161,13 +161,16 @@ class FluentDirectorExtension extends Extension
      * @param array $rules Array of rules to insert before
      * @param string $key Rule to insert the new rules before
      * @param array $rule New Rules to insert
+     * @param boolean $prependIfMissing Prepend the new rules if the insert before rule cannot be found
      * @return array Resulting array of rules
      */
-    protected function insertRuleBefore(array $rules, $key, array $rule)
+    protected function insertRuleBefore(array $rules, $key, array $rule, $prependIfMissing = true)
     {
         $i = array_search($key, array_keys($rules));
         if ($i !== false) {
             return array_slice($rules, 0, $i, true) + $rule + array_slice($rules, $i, null, true);
+        } elseif ($prependIfMissing) {
+            $rules = $rule + $rules;
         }
 
         return $rules;
--- a/src/Extension/FluentDirectorExtension.php
+++ b/src/Extension/FluentDirectorExtension.php
@@ -68,14 +68,10 @@ class FluentDirectorExtension extends Extension
     public function updateRules(&$rules)
     {
         $originalRules = $rules;
-        $rules = $this->getExplicitRoutes($originalRules);
+        $fluentRules = $this->getExplicitRoutes($rules);
 
-        // Merge all other routes (maintain priority)
-        foreach ($originalRules as $key => $route) {
-            if (!isset($rules[$key])) {
-                $rules[$key] = $route;
-            }
-        }
+        // Insert Fluent Rules before the default '$URLSegment//$Action/$ID/$OtherID'
+        $rules = $this->insertRuleBefore($rules, '$URLSegment//$Action/$ID/$OtherID', $fluentRules);
 
         $request = Injector::inst()->get(HTTPRequest::class);
         if (!$request) {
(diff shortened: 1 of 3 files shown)


UndefinedOffset:
Sure can do, I'll see if I can tackle that in the am.

One thing I didn't like about the change is that it puts the root url rules for the locale's slug (i.e /us/ and /ca/ for example) down near the bottom of the list too unlike the default root url rule which is at the very top. But if they're up there too every page renders as the home page which is less than ideal ;). I couldn't figure out why the rule would cause that, so I left them at the bottom which seems to work fine anyways.

UndefinedOffset:
@robbieaverill do you know of a way to set a config after another in a unit test? If I use the `extra_controllers` config option in a test it places the rule at the top. Which causes the test to pass regardless of whether this pull request is included or not.

UndefinedOffset:
@robbieaverill Found a work around actually, unit test added and it seems to be passing. I tried it with and without the changes in this pull request and it fails without the changes.

src/Extension/FluentDirectorExtension.php
```diff
@@ -159,4 +155,21 @@ protected function getRuleController($existingRule, $localeObj)
...
+     * @param string $key Rule to insert the new rules before
+     * @param array $rule New Rules to insert
+     * @return array Resulting array of rules
+     */
+    protected function insertRuleBefore(array $rules, $key, array $rule)
+    {
+        $i = array_search($key, array_keys($rules));
+        if ($i !== false) {
+            return array_slice($rules, 0, $i, true) + $rule + array_slice($rules, $i, null, true);
+        }
```
ScopeyNZ: Should we expect this function to always add the rule? The `FieldList` has `appendIfMissing` as an argument which defaults to `true`.

UndefinedOffset: Since it's looking for the rule defined by the [cms module](https://github.com/silverstripe/silverstripe-cms/blob/4.3/_config/routes.yml#L22) it should based on the current module requirements. That said I could add the appendIfMissing idea, I do like that as a fall back for sure though I might do a prepend so it works essentially like Fluent currently does.

UndefinedOffset:
> 
> 
> Looks pretty good. I'd like to set-up your original issue in an environment to understand the issue better but I don't have the time right now for that 😅

Ya and unfortunately due to confidentiality with the client I can't share it unfortunately, it's also in a pretty broken state big big site upgrade to SilverStripe 4 and migration from translatable so it's a bit unhappy to begin with (even with a pretty much empty database). At least this issue is a routing problem that was fairly easily solved, I'll take the fact that it shows 0 search results on my search controller (where I was having this issue) as win now if it had returned a result well... baby steps lol.

ScopeyNZ:
I'll see if I can get a colleague who's familiar with Fluent to have a look over it

UndefinedOffset:
This might help from our routes config we had this:

```yml
---
Name: mysite_routes
After:
    - "framework/*"
    - "#fluentrouting"
---
SilverStripe\Control\Director:
    rules:
        'us/search//$Action/$ID/$OtherID': 
            Controller: 'Project\Control\SearchController'
            l: "en_US"
        'ca/search//$Action/$ID/$OtherID': 
            Controller: 'Project\Control\SearchController'
            l: "en_CA"
````

With Fluent's rules for `us/$URLSegment//$Action/$ID/$OtherID` and `ca/$URLSegment//$Action/$ID/$OtherID` being added as the first entries in director's rules this nullifies the rules for our search controller since the two Fluent rules were matching first. Where those rules are added via a director extension there's no way to prioritize via the config layer at least above Fluent's rules.

chrispenny:
@ScopeyNZ @robbieaverill Green light from the Terraformers.

I've given this a run through on a project of ours that uses modules such as RestfulServer, GoogleSiteMap (all with Fluent then attached), and which includes a tonne of custom routing: Everything is still looking golden.

Issue #507, opened 2019-03-14
Fluent localized fields cause the sort columns order to be changed


UndefinedOffset:
We have a object that is sorting first by a "Date" field, then the "Created", then the "Title" of the object. Fluent is added to this object and has the "Date" field included so it gets localized. When the query runs without Fluent the query has the correct order of `"Resource"."Date" DESC, "Resource"."Created" DESC, "Resource"."Title"`, but with Fluent enabled on the object the order becomes `""Resource"."Created" DESC, Resource"."Date" DESC, "Resource"."Title"`.

This seems to be due to the fact that Fluent is modifying the original query to include the `COALESCE` for the localized fields.

Seen on SilverStripe 4.3.2 with Fluent 4.2.0.

robbieaverill:
Fixed in #508 - thanks @UndefinedOffset!

PR #508, opened 2019-03-14
BUGFIX: Fixed issue where the sort order of queries would not be maintained

UndefinedOffset:
This pull request fixes the issue mentioned in #507 in short if the sort columns of a query are a mix of localized and non-localized fields the order will become non-localized fields first then the localized fields. This is caused by how the fields are removed and re-added to the associative array not factoring in the non-localized fields keeping them in order at the start of the query.

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -394,7 +394,13 @@ class FluentExtension extends DataExtension
                     // Wrap sort in group to prevent dataquery messing it up
                     unset($order[$column]);
                     $order["({$localisedColumn})"] = $direction;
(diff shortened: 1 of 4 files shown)


robbieaverill:
The change looks good to me, and the fact that the tests pass (of which there are several related to ORM sorting) is good too. Is there any chance you could write a new test for these changes too?

UndefinedOffset:
Sure will do that shortly :)

UndefinedOffset:
There tests added :) and I checked without my fix and the test fails as expected

robbieaverill:
Thanks!

PR #515, opened 2019-04-11
FIX: Fixed issue where order by would not be re-written properly if the sort column is not quoted

UndefinedOffset:
This pull request addresses an issue where the sort column would not be properly re-written when the sort column is not double quoted and is a localized column. On MySQL 5.7.5+ this causes the query to crash due to `ONLY_FULL_GROUP_BY` being enabled in `ANSI` mode and by default.

For example given the below object, attempting to do a `MyObject::get()` would result in the query's select being rewritten accounting for the localized `Title` column. However the order by is not re-written properly to account for it.

```php
class MyObject extends DataObject
{
    private static $db = [
        'Title' => 'Varchar(50)',
    ];
    
    private static $field_include = [
        'Title',
    ];
    
    private static $extensions = [
        FluentExtension::class,
    ];
    
    private static $default_sort = 'Title';
}
```

I'm not 100% confident in this fixing it fully as my site's tests erroring in one specific case on MySQL 5.7.25 due to only full group by but the rest of the cases that were erroring out do work now.

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -382,6 +382,10 @@ class FluentExtension extends DataExtension
                 $localisedColumn = $column;
                 // Fix non-fully-qualified name
                 if (!$fqn) {
+                    if (strpos($localisedColumn, '"') === false) {
+                        $localisedColumn = '"' . $localisedColumn . '"';
+                    }
+                    
                     $localisedColumn = str_replace(
                         "\"{$field}\"",
                         "\"{$table}\".\"{$field}\"",
@@ -985,6 +989,18 @@ class FluentExtension extends DataExtension
 
         // Check sole "field" without table specifier ("name" without leading or trailing '.')
         if (preg_match('/(?<![.])"(?<field>\w+)"(?![.])/i', $sql, $matches)) {
+            $field = $matches['field'];
+
+            // Check if this field is in any of the tables, and just pick any that match
+            foreach ($tables as $table => $fields) {
+                if (in_array($field, $fields)) {
+                    return [$table, $field, false];
+                }
+            }
+        }
+
+        // Check sole "field" without table specifier ("name" without leading or trailing '.' and without quotes)
+        if (preg_match('/(?<![.])(?<field>\w+)(?![.])/i', $sql, $matches)) {
             $field = $matches['field'];
 
             // Check if this field is in any of the tables, and just pick any that match


src/Extension/FluentExtension.php
```diff
@@ -995,6 +999,18 @@ protected function detectLocalisedTableField($tables, $sql)
             }
         }
 
+        // Check sole "field" without table specifier ("name" without leading or trailing '.' and without quotes)
+        if (preg_match('/(?<![.])(?<field>\w+)(?![.])/i', $sql, $matches)) {
```
ScopeyNZ: Isn't this negative lookahead/behind pointless given the field should be quoted (ie. the field should always have a `"` delimiting it anyway)? 

Wouldn't this make more sense:

```
if (preg_match('/"?(?<field>\w+)"?(?!\.)/i', $sql, $matches)) {
```

Match a `"` maybe, then your field, then another `"` maybe, then not followed by a `.`.

ScopeyNZ: Actually I thought about this a little more. I'm pretty sure the negative lookahead you've got will only apply to the last character. So you'll have `Foo.Bar` matching and `$matches['field']` will be `Ba`.

I think you want a regex more like this:

```
/(?!\w+"?\.)(\w+)/i
```

This is a negative match ahead on anything followed by a dot, and then a match on whatever is left. All these regexes only work if the `$sql` being matched is only the field btw. If there's `SELECT` or `WHERE` conditions then it's going to be weird.

UndefinedOffset: Ya that's what I'm worried about actually... this method is used for more than just sort columns, I wonder if perhaps it would be better to check for quotes and add them if they're not present prior to this method being called when updating the sort.

UndefinedOffset: I wonder if something like this makes more sense:

```diff
@@ -376,6 +376,11 @@ class FluentExtension extends DataExtension
         // Iterate through each order clause, replacing each with the translated version
         $order = $query->getOrderBy();
         foreach ($order as $column => $direction) {
+            // Fix conditions where the column is a non-fully-qualified name and is not quoted
+            if (strpos($column, '.') === false && strpos($column, '"') === false) {
+                $column = '"' . $column . '"';
+            }
+            
             // Parse fragment for localised field and table
             list ($table, $field, $fqn) = $this->detectLocalisedTableField($tables, $column);
             if ($table && $field) {

```

robbieaverill:
On a side note, would you mind adding a test for this? The ORM modifications that fluent performs are pretty well covered by automated tests already, and we'd like to keep that up - especially since it's quite a complex set of logic

tractorcow:
Just revisiting some old PRs. I think this PR probably makes sense at first glance, but it does need some tests to provide reassurance. :)

Issue #389, opened 2018-04-04
Show full locale in the 'badges'


JorisDebonnet:
Commit e75cbc1d has introduced a badge showing part of the locale next to the Page Title. That's nice to have, but it only shows the suffix, e.g. NZ or US. That is useful when you have localized pages such as en_NZ and en_US, but is rather useless when using _translated_ pages that may have the same country component, such as en_CA and fr_CA, or nl_BE and fr_BE. 

Possible solutions: 
1. Show the whole locale instead of just the country part
2. Show the locale URLSegment instead
3. Make it a config (locale / language / country / urlsegment / ?)

I can make the change if we can agree on option 1 or 2. 
I personally suggest 1, as I don't think it will hurt anyone, and it is the most informative. But option 2 could be a good compromise because it likely focuses on whatever the developer is using Fluent for.

robbieaverill:
We suspected this might be a problem. Currently the templates on the frontend are doing a similar thing too.

I'd be in favour of option 3, or provide an extension point to update the formatting for what gets used in the badge.

We want to keep the badge small if possible, and recognise that some users will have different requirements.

How would that sit with you?

JorisDebonnet:
I would prefer a config to an extension point then (or both; perhaps people want to put little flags), but in any case those options seem like a little too much work for myself to work on :)

clarkepaul:
Question, what does the URLSegment output generally? 
Option 1 seems like a good incremental improvement, I think there would enough space for en_CA (even though its not as visually pleasing its more informative), but I it sound like if we would do it properly then it should be dev defined with opt. 3 as it provides the most flexibility.

JorisDebonnet:
The reason I added option 2 with the URLSegment is because this is in itself already a user-defined abbreviation for the locale. In that sense, it could serve as the badge as well. By default it is the locale itself (e.g. `/en_US/`) but if a website focusses on localization within one language, the admin will likely use something like `/us/` and `/nz/`, while websites focussing on languages will likely have something like `/en/` and `/nl/`.

I agree 3 has the most flexibility, but top flexibility is not "necessarily" the best option because it adds complexity. If we believe option 2 could serve all cases, I believe it should be considered. Can someone argue in which situation option 2 would not suffice? (perhaps in conjunction with an extension point, as @robbieaverill suggested)

ninty9notout:
[...]

PR #517, opened 2019-04-17
Use the locale's URLSegment as default badge label

JorisDebonnet:
See the comments since April 2019 on #389. Improves #510.

### DIFF
--- a/src/Model/Locale.php
+++ b/src/Model/Locale.php
@@ -147,7 +147,7 @@ class Locale extends DataObject
     }
 
     /**
-     * Get a short segment of the locale code for display in things like badges
+     * Get the locale's country part
      *
      * @return string e.g. "NZ" for "en_NZ"
      */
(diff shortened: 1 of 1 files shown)


robbieaverill:
@clarkepaul are you happy with this becoming the new default? It'd be `en_NZ` rather than `NZ`, for example

robbieaverill:
Actually I think this is an enhancement, so we should target the master branch instead. Would you mind changing the target branch and rebasing the PR?

JorisDebonnet:
> 
> 
> @clarkepaul are you happy with this becoming the new default? It'd be `en_NZ` rather than `NZ`, for example

That's assuming you leave the default URLSegment, of course. Personally, I like using /en/ etc in the URLs, and then the labels also just say EN.

robbieaverill:
Sorry yeah I was thinking of the full locale. I think that's OK, because it's probably been customised so it looks better than the default locale

JorisDebonnet:
Space removed and base changed. I guess this means it will only go live in Fluent 4.3... aww.

robbieaverill:
Thanks, yeah - but you can use `4.3.x-dev` in your projects until it's ready

JorisDebonnet:
Is anything else needed to make this fit for merging?

robbieaverill:
The bump was enough, thank you

PR #523, opened 2019-05-30
FIX Fluent now respects existing base URL prefixes in URL segment fields when no fluent domain exists

robbieaverill:
Fluent sets the URLSegment field's URLPrefix to be the base domain plus the parent page path including the current locale. When no fluent domains are configured, this defaulted to `Director::absoluteBaseURL()`. This was a bit heavy handed, since it doesn't consider any previous modifications to `SiteTreeURLSegmentField::setURLPrefix()` (e.g. in subsites, where it sets the subsite domain).

I've changed this to use the existing URL prefix instead.

There will still be conflicts if you use subsite domains and fluent domains at the same time, since they represent the same functionality, but that can be documented.

Fixes https://github.com/silverstripe/silverstripe-subsites/issues/416

### DIFF
--- a/src/Extension/FluentSiteTreeExtension.php
+++ b/src/Extension/FluentSiteTreeExtension.php
@@ -244,9 +244,9 @@ class FluentSiteTreeExtension extends FluentVersionedExtension
             return $this;
         }
 
-        // Mock frontend and get link to parent object / page
-        $baseURL = FluentState::singleton()
-            ->withState(function (FluentState $tempState) {
+        // Mock frontend and set link to parent object / page
+        FluentState::singleton()
+            ->withState(function (FluentState $tempState) use ($segmentField) {
                 $tempState->setIsDomainMode(true);
                 $tempState->setIsFrontend(true);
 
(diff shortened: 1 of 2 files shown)


tractorcow:
Looks good to me.

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -264,15 +264,14 @@ protected function addLocalePrefixToUrlSegment(FieldList $fields)
                 if ($domain) {
                     $parentBase = Controller::join_links($domain->Link(), Director::baseURL());
                 } else {
-                    $parentBase = Director::absoluteBaseURL();
+                    // Use any existing pre-configured base URL prefix for the field
+                    $parentBase = $segmentField->getURLPrefix();
```
tractorcow: Ah, the bug is that this link includes the parent url, whereas so does $parentRelative. It's adding parent urlsegment twice.

tractorcow: I have to revert this at https://github.com/tractorcow-farm/silverstripe-fluent/pull/539

PR #530, opened 2019-07-02
FIX Support archive query selecting with conditional subselects

ScopeyNZ:
See (and wait for merge on) silverstripe/silverstripe-versioned#225. This PR ensures that max versions returned by the subquery are limited to the current locale.


ScopeyNZ:
There's a lot of push-back on the related PR - this might be something that will forever live in client code but putting this filtering into conditional subselects is making such a huge improvement in performance that it's worth it despite the slight increase in complexity.

PR #531, opened 2019-07-15
WIP: fluent migration task

wernerkrauss:
This task is still work in progress but pretty finished. I mainly submit this PR for feedback and automatic testing via travis

What it does:
* create SQL statements for all fluent enabled dataobjects and pages
* update the DB

Things to do:
* handle partially translated records
* test if versions have been translated properly (??)

Thanks to @robingram for [providing the actual task](https://gist.github.com/robingram/e27844d19c4853949c79c640d1419799).


robingram:
@tractorcow Am I right in thinking that there is no per-field fallback in the SS4 version of Fluent and that creating a draft in a new locale is a one-off copy of the fields? I.e. updating a field in the default or fallback locale will not be refelected in any other existing locales even if the field is empty there?

If that's correct then this script will need to be updated to copy the field value from the default if the translated version is empty as @wernerkrauss indicates. Also probably skip the copy if all translated fields are empty to enable fallback, which is more complicated.

robingram:
@tractorcow, @robbieaverill would one of you be able to confirm my beliefs as posted above? I'm working on this now and want to update @wernerkrauss' version to handle missing/partial translations:

* There is no per field fallback - if a row exists in the translation table then its values will be used even if they're blank and the default isn't
* If there is no translation for a locale then there shouldn't be any records in the translation tables for that locale

Cheers.

tractorcow:
SS3: Locale fallback was based on the field being null for the shared row.
SS4: Locale fallback was based on the ID of the `_Localised` row for that locale being null (I.e. join failed). This means that you could override a field value with a null, whereas in ss3 you couldn't. :)

In migration I would follow this logic:
 - At least one field is non-null for a locale: Create a `_Localised` table row for that locale.
 - All fields are null for a locale; No `_Localised` row for that locale.
 - Always create a `_Localised` row for the default locale; Copy from base fields (e.g. copy `Title` to a `SiteTree_Localised` row for Locale = default_locale).

robingram:
Thanks @tractorcow, I'd implemented the first two recommendations but I hadn't considered the third. I'll take a look at it.

At the moment I'm copying the default value for blank fields where there are only some translated fields. This is on a 'no surprises' basis since it will mimic the SS3 behaviour but then allow the value to be removed if it really should be null.

wernerkrauss:
IMHO we're ready to merge. the test if the versions were translated properly is still missing. Is this a show stopper?

src/Task/FluentMigrationTask.php
```diff
@@ -205,18 +206,38 @@ function ($className) {
...
 
+    /**
+     * Return the fields that actually exist on a table
+     *
+     * @param string $table
+     * @return array
+     */
+    protected function getExistingFields($table)
+    {
+        $query = sprintf("SHOW COLUMNS from %s", $table);
```
tractorcow: Check `DB::field_list`

src/Task/FluentMigrationTask.php
```diff
@@ -205,18 +206,38 @@ function ($className) {
...
      * @return string
      */
-    protected function buildSelectFieldList($sourceTable, $fields, $locale)
+    protected function buildSelectFieldList($sourceTable, $fields, $locale, $existingFields)
     {
         return implode(
             ', ',
             array_map(
-                function ($field) use ($locale, $sourceTable) {
-                    return sprintf('COALESCE(`%s_%s`, `%s`.`%s`) AS `%s`', $field, $locale, $sourceTable, $field, $field);
```
tractorcow: Pre-existing issue, but I highly prefer ansi-quoted table identifiers (`"`) not backticks when forming SQL queries. One is standard, other is mysql specific.

tractorcow:
Small review :D

tractorcow:
Heads up I may need to come back to this later when I have more time.

bummzack:
Having this directly available would be awesome… saved me a lot of time right now 👍

PR #532, opened 2019-07-16
Add `$SelectedLanguage` template variable to fluent

bergice:
Already exists in `cwp/cwp` but it should be available in `FluentExtension`

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -997,4 +997,14 @@ class FluentExtension extends DataExtension
 
         return [null, null, false];
     }
+
+    /**
+     * Returns the selected language
+     *
+     * @return ArrayData
+     */
+    public function getSelectedLanguage()
+    {
+        return $this->LocaleInformation(FluentState::singleton()->getLocale());
+    }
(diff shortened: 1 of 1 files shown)


robbieaverill:
For what it's worth, the use of this in the CWP themes has been removed with https://github.com/silverstripe/cwp-starter-theme/pull/153

src/Extension/FluentExtension.php
```diff
@@ -997,4 +997,17 @@ protected function detectLocalisedTableField($tables, $sql)
...
+     * Returns the native language name for the selected locale/language
+     *
+     * @return string
+     */
+    public function getSelectedLanguage()
+    {
+        /** @var ArrayData $information */
+        $information = $this->LocaleInformation(FluentState::singleton()->getLocale());
+
+        return $information->LanguageNative;
```
tractorcow: Instead of adding a method that only accesses this single field, why don't we return `$information` and let you pick what field you use in the template?

Suggest to name this method `SelectedLocaleInformation` (same as `LocaleInformation` but in selected locale).

PR #535, opened 2019-07-19
Replace filter grid with checkbox set

ian-patel:


tractorcow:
This looks a lot nicer and easier for users. :)

PR #536, opened 2019-07-25
ENHANCEMENT Configurable policies for handling the deletion of objects

tractorcow:
Hello I am back from the dead. :)

This feature attempts to solve some of the inconsistencies involved with deleting objects; If you have an object that's localised, deleting the record should remove that localisation, but not all others.

Acceptance Criteria:
 - When  deleting an object in one locale, it should not delete it from any other locale (unless it's already deleted there).
- After an object has been deleted in all locales the "delete" button deletes the base record.
- Deleting filtered objects should disable this item in the current locale
- Deleting localised objects should delete the localisation for this item in the current locale
- Deleting a localised + filtered object should delete both the localisation, and disable this item in the current locale

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -16,9 +16,9 @@ use SilverStripe\ORM\DataQuery;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\Queries\SQLConditionGroup;
-use SilverStripe\ORM\Queries\SQLDelete;
 use SilverStripe\ORM\Queries\SQLSelect;
 use SilverStripe\View\ArrayData;
+use TractorCow\Fluent\Model\Delete\UsesDeletePolicy;
 use TractorCow\Fluent\Model\Locale;
 use TractorCow\Fluent\State\FluentState;
 
@@ -37,6 +37,11 @@ use TractorCow\Fluent\State\FluentState;
 class FluentExtension extends DataExtension
 {
     /**
+     * Deletions are managed via DeletePolicy
+     */
+    use UsesDeletePolicy;
+
+    /**
      * The table suffix that will be applied to create localisation tables
      */
     const SUFFIX = 'Localised';
@@ -53,9 +58,9 @@ class FluentExtension extends DataExtension
      * @var array
      */
     private static $db_for_localised_table = [
-        'ID' => 'PrimaryKey',
+        'ID'       => 'PrimaryKey',
         'RecordID' => 'Int',
-        'Locale' => 'Varchar(10)',
+        'Locale'   => 'Varchar(10)',
     ];
 
     /**
@@ -66,7 +71,7 @@ class FluentExtension extends DataExtension
      */
     private static $indexes_for_localised_table = [
         'Fluent_Record' => [
-            'type' => 'unique',
+            'type'    => 'unique',
             'columns' => [
                 'RecordID',
                 'Locale',
@@ -173,7 +178,7 @@ class FluentExtension extends DataExtension
      * Check if a field is marked for localisation
      *
      * @param string $field Field name
-     * @param string $type Field type
+     * @param string $type  Field type
      * @param string $class Class this field is defined in
      * @return bool
      */
@@ -244,8 +249,8 @@ class FluentExtension extends DataExtension
      * Helper function to check if the value given is present in any of the patterns.
      * This function is case sensitive by default.
      *
-     * @param string $value A string value to check against, potentially with parameters (E.g. 'Varchar(1023)')
-     * @param array $patterns A list of strings, some of which may be regular expressions
+     * @param string $value    A string value to check against, potentially with parameters (E.g. 'Varchar(1023)')
+     * @param array  $patterns A list of strings, some of which may be regular expressions
      * @return bool True if this $value is present in any of the $patterns
      */
     protected function anyMatch($value, $patterns)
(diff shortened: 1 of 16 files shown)


src/Extension/FluentExtension.php
```diff
@@ -642,7 +573,7 @@ public function getLocalisedTable($tableName, $locale = '')
      * @param string $locale If passed, this is the locale we wish to delete in. If empty this is the root table
      * @return string
      */
-    protected function getDeleteTableTarget($tableName, $locale = '')
+    public function getDeleteTableTarget($tableName, $locale = '')
```
robbieaverill: This is a breaking change - any way we can avoid that?

src/Model/Delete/DeleteFilterPolicy.php
```diff
@@ -0,0 +1,39 @@
...
+ * A policy that deletes the filtered rows.
+ *
+ * Requires {@see FluentFilteredExtension} on the object
+ */
+class DeleteFilterPolicy implements DeletePolicy
+{
+
+    /**
+     * @param DataObject|FluentFilteredExtension $record
+     * @return bool
```
robbieaverill: Since this operation doesn't return a boolean result of whether it was successful (which is what I might assume it does), it may be worth adding a note about what the boolean return is for

```suggestion
     * @return bool Whether this record is visible in any other locale after the delete operation
```

src/Model/Delete/DeleteLocalisationPolicy.php
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * A policy that deletes all localisations for a record
+ *
+ * Requires that the object has {@see FluentExtension}
+ */
+class DeleteLocalisationPolicy implements DeletePolicy
+{
+    /**
+     * @param DataObject|FluentExtension $record
+     * @return bool
```
robbieaverill: Same suggestion here re: description

src/Model/Delete/DeletePolicy.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace TractorCow\Fluent\Model\Delete;
+
+use SilverStripe\ORM\DataObject;
+
+interface DeletePolicy
+{
+    /**
+     * @param DataObject $record
+     * @return bool Determines if any dependent objects block upstream deletion (e.g. db / model constraints)
```
robbieaverill: Is this correct?

src/Model/Delete/DeletePolicyFactory.php
```diff
@@ -0,0 +1,47 @@
...
+        $dependantPolicies = [];
+
+        // Fluent extension requires rows in current locale to be removed
+        if ($object->hasExtension(FluentExtension::class)) {
+            $dependantPolicies[] = new DeleteLocalisationPolicy();
+        }
+
+        // Filtered extension requires mapping table to locale to be removed
+        if ($object->hasExtension(FluentFilteredExtension::class)) {
+            $dependantPolicies[] = new DeleteFilterPolicy();
```
robbieaverill: Do you want to use Injector to instantiate these? Or do we treat this factory as the API, meaning devs would replace the factory in order to customise any of the policies under it?

tests/php/Model/Delete/DeleteRecordPolicyTest.php
```diff
@@ -0,0 +1,176 @@
...
+        Domain::clearCached();
+
+        // Set default
+        FluentState::singleton()
+            ->setLocale('es_US')
+            ->setIsDomainMode(false);
+    }
+
+    /**
+     * @throws ValidationException
```
robbieaverill: What's this for? Is this a PHPStorm addition, or an expected exception?

tests/php/Model/Delete/Fixtures/FilteredRecord.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace TractorCow\Fluent\Tests\Model\Delete\Fixtures;
+
+use SilverStripe\ORM\DataObject;
+use TractorCow\Fluent\Extension\FluentFilteredExtension;
+
+/**
+ * @mixin FluentFilteredExtension
+ */
+class FilteredRecord extends DataObject
```
robbieaverill:
```suggestion
class FilteredRecord extends DataObject implements TestOnly
```

BYO namespace - same for the other test stubs

ian-patel:
@tractorcow Good job!

tests/php/Model/Delete/DeleteRecordPolicyTest.php
tractorcow: First lol.

tests/php/Model/Delete/Fixtures/FilteredRecord.php
tractorcow: Oh wow I forgot about that. Thanks for the reminder.

src/Model/Delete/DeletePolicyFactory.php
tractorcow: Well, the factory is used by injector, so I would probably consider this already "injected". Otherwise we run the risk of loops. :)

src/Model/Delete/DeletePolicy.php
tractorcow: Yes and it's annoying, but it's the main reason I had to create a nested policy; The use case where an object had both extensions was a pain to solve.

src/Extension/FluentExtension.php
tractorcow: Hm yeah I can put it into a public accessor and call that, instead of changing the visibility. :)

tractorcow:
It's nice to see you again @robbieaverill :)

tractorcow:
@robbieaverill done :)

tractorcow:
Nice, this PR bumped the coverage over the golden 70% mark. :)

stevie-mayhew:
Great work @tractorcow

PR #545, opened 2019-09-03
FIX Support archive query selecting with conditional subselects

adrhumphreys:
Dupe of https://github.com/tractorcow-farm/silverstripe-fluent/pull/530 but with a bugfix for fallback locales

### DIFF
--- a/src/Extension/FluentVersionedExtension.php
+++ b/src/Extension/FluentVersionedExtension.php
@@ -146,6 +146,49 @@ class FluentVersionedExtension extends FluentExtension
     }
 
     /**
+     * Hook into Versioned::prepareMaxVersionSubSelect - ensure that max version returned by the subquery are limited to
+     * the current locale
+     *
+     * @param SQLSelect $subSelect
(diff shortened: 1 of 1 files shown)


src/Extension/FluentVersionedExtension.php
```diff
@@ -145,6 +145,49 @@ public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null)
...
+            $joinCondition
+        );
+
+        $locales = "'{$locale->getLocale()}'";
+
+        foreach ($locale->FallbackLocales() as $fallbackLocale) {
+            $locales .= ", '{$fallbackLocale->Locale()->getLocale()}'";
+        }
+
+        $subSelect->addWhere("\"$localisedTable\".\"Locale\" in ({$locales})");
```
tractorcow: I hope you don't mind but can we please parameterise this? You can use `DB::placeholders()` to generate the `IN ($placeholders)` component, and then the arguments are the list of locales in fallbacks. That would greatly reduce the risk of SQL injection :)

PR #547, opened 2019-09-13
Add SS Core Fluent search extension

taoceanz:
Add search extension to enable using SS Core search functionality with Fluent.

### DIFF
--- a/docs/en/configuration.md
+++ b/docs/en/configuration.md
@@ -194,6 +194,23 @@ MySite\Model\MyModel:
 The result is that a DataObject that has *not* been Localised, will display on the frontend with content populated by
 it's Fallbacks (the same beheviour as what you see when viewing DataObjects from within the CMS).
 
+## SilverStripe Fluent and search
+
+### SilverStripe Core search
+
+To use SilverStripe Core search with Fluent, add the `FluentSearchForm` to your configuration.
+
+```yaml
+---
+Name: myextensions
+---
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\CMS\Search\SearchForm:
+    class: TractorCow\Fluent\Search\FluentSearchForm
+``` 
+
+**Note:** If you're using the SilverStripe CMS `$SearchForm` to display the default SilverStripe template on the front-end, this will update form and input id and class selectors from 'SearchForm' to 'FluentSearchForm'. E.g. form element `id='SearchForm_SearchForm'` will become `id='FluentSearchForm_SearchForm'`.
+
 ## Locale based filter configuration
 
 In addition to localising fields within a DataObject, a filter can also be applied
--- a/src/Search/FluentSearchForm.php
+++ b/src/Search/FluentSearchForm.php
@@ -1,0 +1,103 @@
+<?php
+
+/**
+ * SilverStripe core search work around when using Fluent
+ * @link https://gist.github.com/baukezwaan/266c469758319daea5460f2a5647c525
+ */
+
+namespace TractorCow\Fluent\Search;
+
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\CMS\Search\SearchForm;
+use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\PaginatedList;
+use TractorCow\Fluent\State\FluentState;
+
+/**
+ * Simple extension to enable SS Core search to work with SS Fluent
+ */
+class FluentSearchForm extends SearchForm
+{
+    /**
+     * Return dataObjectSet of the results using current request to get info from form.
+     * Wraps around {@link searchEngine()}.
+     *
+     * @return PaginatedList
+     */
+    public function getResults()
+    {
+        $keywords = $this->getSearchQuery() ?: '';
+        $request = $this->getRequestHandler()->getRequest();
+        // Define and generate keyword search patterns
+        $keyword_patterns = [
+            [
+                'patterns' => [
+                    '/()("[^()"]+")( and )("[^"()]+")()/i',
+                    '/(^| )([^() ]+)( and )([^ ()]+)( |$)/i',
+                ],
+                'callback' => function ($matches) {
+                    return ' +' . $matches[2] . ' +' . $matches[4] . ' ';
+                },
+            ],
+            [
+                'patterns' => [
+                    '/(^| )(not )("[^"()]+")/i',
+                    '/(^| )(not )([^() ]+)( |$)/i',
+                ],
+                'callback' => function ($matches) {
+                    return ' -' . $matches[3];
+                },
+            ],
+        ];
+        foreach ($keyword_patterns as $k_p) {
+            $keywords = preg_replace_callback(
+                $k_p['patterns'],
+                $k_p['callback'],
+                $keywords
+            );
+        }
+
+        // Generate and query for locale based search results
+        $current_locale = FluentState::singleton()->getLocale();
+        $sitetree_table_localised = "SiteTree_Localised_{$current_locale}";
+        $sitetreetables = '"SiteTree_Live"."Title", "SiteTree_Live"."MenuTitle", "SiteTree_Live"."Content", "SiteTree_Live"."MetaDescription"';
+        $sql = <<<SQL
+            SELECT DISTINCT "SiteTree_Live"."ID", "SiteTree_Live"."Title", MATCH ({$sitetreetables}) AGAINST (?) AS Relevance
+            FROM "SiteTree_Live"
+            LEFT JOIN "SiteTree_Localised" AS "{$sitetree_table_localised}"
+            ON "SiteTree_Live"."ID" = "{$sitetree_table_localised}"."RecordID" AND "{$sitetree_table_localised}"."Locale" = ? AND (("{$sitetree_table_localised}"."ID" IS NOT NULL))
+            WHERE (MATCH ({$sitetreetables}) AGAINST (? IN BOOLEAN MODE) + MATCH ({$sitetreetables}) AGAINST (? IN BOOLEAN MODE) AND "SiteTree_Live"."ShowInSearch" <> 0)
+            ORDER BY Relevance DESC
+SQL;
+        $params = [
+            str_replace(['*', '+', '-'], '', $keywords),
+            $current_locale,
+            $keywords,
+            htmlentities($keywords, ENT_NOQUOTES, 'UTF-8')
+        ];
+        // Generate results list
+        $sitetree_objects = SiteTree::get()
+            ->innerJoin(
+                '(' . $sql . ')',
+                '"Fulltext"."ID" = "SiteTree"."ID"',
+                'Fulltext',
+                20,
+                $params
+            )
+            ->sort('"Fulltext"."Relevance" DESC');
+        // Filter out non-viewable
+        $results = new ArrayList();
+        foreach ($sitetree_objects as $sto) {
+            if (empty($sto) || !$sto->canView()) {
+                continue;
+            }
+            $results->push($sto);
+        }
+
+        // Paginate results
+        return PaginatedList::create($results)
+            ->setPageStart($request->requestVar('start') ?: 0)
+            ->setPageLength($this->getPageLength())
+            ->setTotalItems($results->count());
+    }
+}


src/Extension/FluentSearchFormExtension.php
```diff
@@ -0,0 +1,93 @@
...
+            'WHERE (',
+            "MATCH ({$sitetreetables}) AGAINST ('{$keywords}' IN BOOLEAN MODE)",
+            "+ MATCH ({$sitetreetables}) AGAINST ('{$htmlEntityKeywords}' IN BOOLEAN MODE)",
+            'AND SiteTree_Live.ShowInSearch <> 0',
+            ')',
+            'ORDER BY Relevance DESC',
+            'LIMIT 20'
+        ]);
+
+        $customResults = DB::query($query);
```
tractorcow: I highly recommend to use parameterised queries here; There's a high risk of SQL injection with variable substitution in SQL code.

taoceanz: Thanks for your suggestion, Damian. Are you meaning all variables in the query need to be parameterised? I'm not an expert with parameterised queries so could be missing basic understand here, but attempting to pass `$sitetreetables` or `$sitetree_table_localised` i.e. table names, through as params, the query fails.

What are your thoughts on changing it to this?
```
$sql = implode(
    ' ',
    [
        'SELECT DISTINCT SiteTree_Live.ID, SiteTree_Live.Title,',
        "MATCH ({$sitetreetables}) AGAINST (?) AS Relevance",
        'FROM SiteTree_Live',
        "LEFT JOIN SiteTree_Localised AS {$sitetree_table_localised}",
        "ON SiteTree_Live.ID = {$sitetree_table_localised}.RecordID AND {$sitetree_table_localised}.Locale = ? AND (({$sitetree_table_localised}.ID IS NOT NULL))",
        "WHERE (MATCH ({$sitetreetables}) AGAINST (? IN BOOLEAN MODE) + MATCH ({$sitetreetables}) AGAINST (? IN BOOLEAN MODE) AND SiteTree_Live.ShowInSearch <> 0)",
        'ORDER BY Relevance DESC',
        'LIMIT 20'
    ]
);
$params = [
    $keywords_strrpl,
    $current_locale,
    $keywords,
    $htmlEntityKeywords
];
$custom_results = DB::prepared_query($sql, $params);
```
@tractorcow

tractorcow: Easy way to remember;
 - If it's a table or column name, don't parameterise it, but do double quote it
- If it's a value, do parameterise it, use a `?` placeholder.

E.g. 
```php
DB::prepared_query("SELECT * FROM \"{$table\"} WHERE "\"{$table}\".\"{$column}\" = ?", [$value]);
```

tractorcow: > $sql = implode(
    ' ',
    [
        'SELECT DISTINCT SiteTree_Live.ID, SiteTree_Live.Title,',
        "MATCH ({$sitetreetables}) AGAINST (?) AS Relevance",
        'FROM SiteTree_Live',
        "LEFT JOIN SiteTree_Localised AS {$sitetree_table_localised}",
        "ON SiteTree_Live.ID = {$sitetree_table_localised}.RecordID AND {$sitetree_table_localised}.Locale = ? AND (({$sitetree_table_localised}.ID IS NOT NULL))",
        "WHERE (MATCH ({$sitetreetables}) AGAINST (? IN BOOLEAN MODE) + MATCH ({$sitetreetables}) AGAINST (? IN BOOLEAN MODE) AND SiteTree_Live.ShowInSearch <> 0)",
        'ORDER BY Relevance DESC',
        'LIMIT 20'
    ]
);

Suggest to use HEREDOC for formatting these long-line strings.

```php
$sql = <<<SQL
SELECT DISTINCT "SiteTree_Live"."ID", "SiteTree_Live"."Title"
MATCH ("{$sitetreetables}") AGAINST (?) AS "Relevance"
FROM "SiteTree_Live"
LEFT JOIN "SiteTree_Localised" AS "{$sitetree_table_localised}"
ON
    "SiteTree_Live"."ID" = "{$sitetree_table_localised}"."RecordID"
    AND "{$sitetree_table_localised}"."Locale"" = ?
    AND "{$sitetree_table_localised}"."ID" IS NOT NULL
    WHERE (
        MATCH ("{$sitetreetables}") AGAINST (? IN BOOLEAN MODE) > 0
        AND "SiteTree_Live"."ShowInSearch" > 0
    )
SQL
    ;

Comment (tractorcow) [reply src/Extension/FluentSearchFormExtension.php]: With heredoc `$` var substitution works, but you don't need to escape double quotes. :)

Comment (taoceanz) [reply src/Extension/FluentSearchFormExtension.php]: Hey Damian, please have a look now. I've updated the query to be parameterised and have defined the query via HEREDOC.
@tractorcow

File: src/Extension/FluentSearchFormExtension.php [hunk 2]
```diff
@@ -0,0 +1,94 @@
...
+        };
+
+        $notProcessor = function ($matches) {
+            return ' -' . $matches[3];
+        };
+
+        $keyword_patterns = [
+            'andProcessor' => '/()("[^()"]+")( and )("[^"()]+")()/i',
+            'andProcessor' => '/(^| )([^() ]+)( and )([^ ()]+)( |$)/i',
+            'notProcessor' => '/(^| )(not )("[^"()]+")/i',
```
tractorcow: You have duplicate array keys here; You'll end up with a 2-length array not a 4-length one.

Instead, try this

```php
$patterns = [
    [
        'pattern' => '/()("[^()"]+")( and )("[^"()]+")()/i',
        'callback' => $andProcessor,
    ],
    // etc...
];
```

src/Extension/FluentSearchFormExtension.php
```diff
@@ -0,0 +1,94 @@
...
+            'notProcessor' => '/(^| )(not )([^() ]+)( |$)/i',
+        ];
+
+        foreach ($keyword_patterns as $callback => $pattern) {
+            $keywords = preg_replace_callback($pattern, $$callback, $keywords);
```
tractorcow: `$$` is unsafe code execution. If you follow my other example, you can use `$item['callback']` instead.

src/Extension/FluentSearchFormExtension.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        $keywords = DB::get_conn()
+            ->escapeString(
```
tractorcow: With parameterised queries don't escapeString(). You must pass it in without escape.

src/Extension/FluentSearchFormExtension.php
```diff
@@ -0,0 +1,94 @@
...
+        $sitetree_table_localised = "SiteTree_Localised_{$current_locale}";
+
+        $sql = <<<SQL
+            SELECT DISTINCT "SiteTree_Live"."ID", "SiteTree_Live"."Title", MATCH ({$sitetreetables}) AGAINST (?) AS Relevance
+            FROM "SiteTree_Live"
+            LEFT JOIN "SiteTree_Localised" AS "{$sitetree_table_localised}"
+            ON "SiteTree_Live"."ID" = "{$sitetree_table_localised}"."RecordID" AND "{$sitetree_table_localised}"."Locale" = ? AND (("{$sitetree_table_localised}"."ID" IS NOT NULL))
+            WHERE (MATCH ({$sitetreetables}) AGAINST (? IN BOOLEAN MODE) + MATCH ({$sitetreetables}) AGAINST (? IN BOOLEAN MODE) AND "SiteTree_Live"."ShowInSearch" <> 0)
+            ORDER BY Relevance DESC
+            LIMIT 20
```
tractorcow: I wonder if we could soft-code this limit? Also when handling pagination, you will need a skip argument to show the following pages of results.

Check MySQLDatabase::searchEngine() for how it does fulltext search and pagination for inspiration.

src/Extension/FluentSearchFormExtension.php
taoceanz: What would you use in place of `escapeString()`? Would `quoteString()` suffice?

src/Extension/FluentSearchFormExtension.php
taoceanz: Good suggestion. I've updated to remove limit and applied pagination. Page length is defined the same way SS core page limit is set.

Thanks for addressing this, it was something I'd forgotten to address.

src/Extension/FluentSearchFormExtension.php
taoceanz: Sure thing, I've fixed this up.

src/Extension/FluentSearchFormExtension.php
tractorcow: I wouldn't use either. By using prepared_query(), you are delegating the escaping of this string to the internal SQL engine. :)

```
$keywords = $request->requestVar('Search');
```

What you've put is exactly what you want.

src/Extension/FluentSearchFormExtension.php
```diff
@@ -0,0 +1,94 @@
...
+            $current_locale,
+            $keywords,
+            $htmlEntityKeywords
+        ];
+        $custom_results = DB::prepared_query($sql, $params);
+
+        $results = new ArrayList();
+
+        foreach ($custom_results as $record) {
+            $st = SiteTree::get_by_id($record['ID']);
```
tractorcow: You could have a quicker result by putting it into one query.

```php
$ids = $custom_results->column('ID');
if ($ids) {
    $pages = SiteTree::get()->filter('ID', $ids);
} else {
    return null;
}
```

tractorcow:
You're doing a great job @taoceanz , I hope you don't feel discouraged by the amount of feedback I've given. :)

taoceanz:
Thanks @tractorcow I really appreciate your feedback. I know you're a wizard so I'm grateful you're taking the time to critique everything in the PR to ensure it's up to standard.

I've pushed the updated code now which contains changes for the comments your made [yesterday](https://github.com/tractorcow-farm/silverstripe-fluent/pull/547#pullrequestreview-291559149).

src/Extension/FluentSearchFormExtension.php
taoceanz: Just saw this comment. Thanks for the suggestion, I'll make the update.

taoceanz: Change made and pushed.

src/Extension/FluentSearchFormExtension.php
```diff
@@ -0,0 +1,103 @@
...
+            $keywords,
+            htmlentities($keywords, ENT_NOQUOTES, 'UTF-8')
+        ];
+        $custom_results = DB::prepared_query($sql, $params);
+
+        // Generate results list
+        $results = new ArrayList();
+        $ids = $custom_results->column('ID');
+        if ($ids) {
+            $pages = SiteTree::get()->filter('ID', $ids);
```
tractorcow: I just realised that my feedback before ruined your `ORDER BY RELEVANCE` you had working nicely in the earlier query. You could possibly implement it via a join instead, and you could retain the sort column.

tractorcow: If your results are large, you may want to move your pagination limit up into this query. See how `setLimitItems` works in MySQLDatabase; By doing this, even if your result sets is large (10,000 items), you only ever load the items for the current page into the php memory.

`->limit($limit, $offset)` to limit the query, and `$list->setLimitItems(false);` when creating paginated list.

See https://github.com/silverstripe/silverstripe-framework/blob/4/src/ORM/Connect/MySQLDatabase.php#L292-L298

tractorcow:
@taoceanz if you get it close, I might fork your work to get it over the line before I merge it. :)

src/Extension/FluentSearchFormExtension.php
taoceanz: There's trouble in getting the total valid query amount to `->setTotalItems()`. I've thought to use `count($ids)`, but that results in a higher count than the total `$results` count as `$ids` contains records we don't have permission to view.

I thought to slice certain indexes from `$ids` in the if statement to check whether we have permission to view, though that doesn't work when adding a limit to our `SiteTree::get()` which will return 10 or 20 records per load, for example. 

I'm unsure how to get only a page lengths worth of results while also attaining the full number of viewable results without doing a query for all SiteTree recods filtered by id.

Does that make sense?

tractorcow: Ok well when you do `$custom_results->column()` that invokes the SQL query and gets the list of ids. Instead of building a query, you could put that into an inner join on the SiteTree::get() query.

E.g.


```php

$results = SiteTree::get()
->addInnerJoin(
    '(' . $sql . ')',
    '"Fulltext"."ID" = "SiteTree"."ID"',
    'Fulltext',
    20,
    $params
)
->sort('"Fulltext"."Relevance" DESC');
```

You'll end up with a $results that matches the full results, sorted by the correct relevance.

Then you can do:

```php
$totalCount = $results->count();
$pagedResults = $results->limit($pageSize, $offset);
```

taoceanz: Thanks for the suggestion on `innerJoin()`. I've not used that before when returning DOs via the ORM. This seems to be more concise and easier to read IMO.

Now I've got the following. I tried limiting the results before passing them into the `PaginatedList` object and to not do `->setPageLength` or `setPageStart`, though that messed with the pagination left page numbers displaying incorrectly e.g. get param of start=30 exists which should lead to is displaying as page 4, though it displays as page 1.

How does this look to you?

```
// Generate results list
$pages = SiteTree::get()
    ->innerJoin(
        '(' . $sql . ')',
        '"Fulltext"."ID" = "SiteTree"."ID"',
        'Fulltext',
        20,
        $params
    )
    ->sort('"Fulltext"."Relevance" DESC');
// Filter out non-viewable
foreach ($pages as $page) {
    if (empty($page) || !$page->canView()) {
        continue;
    }
    $results->push($page);
}

$results_total_items = $results->count();

// Paginate results
$results_paged = new PaginatedList($results);
$results_paged->setPageStart($page_start);
$results_paged->setPageLength($page_length);
$results_paged->setTotalItems($results_total_items);

return $results_paged;
```

taoceanz:
Thanks for the suggestion @xini I've moved the file to Search and renamed to `TractorCow\Fluent\Search\FluentSearchForm` + reflected these changes in the docs. I've also rebased all commits since the original PR and pushed those changes. @tractorcow are there any other changes you'd suggest on this?

taoceanz:
@tractorcow are there any other change requests before we can merge this?

src/Search/FluentSearchFormExtension.php
```diff
@@ -0,0 +1,110 @@
...
+                    '/(^| )( not )("[^"()]+")/i',
+                    '/(^| )( not )([^() ]+)( |$)/i',
+                ],
+                'callback' => function ($matches) {
+                    return ' -' . $matches[3];
+                },
+            ],
+        ];
+        foreach ($keyword_patterns as $k_p) {
+            $keywords = preg_replace_callback(
```
tractorcow: you assign $keywords on each loop. This means that only the last loop will be executed (it will overwrite each prior loop).

Assign $keywords = $this->getSearchQuery(); before this loop to fix, and pass in $keywords to the preg_replace_callback()

tractorcow:
Just one bug left in the code. ;d

xini:
Can you please rename the file for TractorCow\Fluent\Search\FluentSearchForm as well? Cheers

src/Search/FluentSearchFormExtension.php
taoceanz: Good spotting, thank you. I've fixed that up to assign the search term before the loop to make sure it replaces find for both 'and' and 'not'.

taoceanz:
Damian, keywords issue I think is fixed and replaces correctly (as far as I've tested) for both sets of patterns. Additionally I've removed some of the variables that were being used only once e.g. when generating the PaginatedList I've set the page length with `$this->getPageLength()` rather than setting that to a variable then passing that variable into the set the page length.
Florian, the file name has been updated too. Thanks for pointing out it the file name hadn't been updated.
@tractorcow @xini

tractorcow:
Good job @taoceanz, I appreciate your work on this issue. :)

taoceanz:
My pleasure, @tractorcow I appreciate all the suggestions and feedback you've given to get this merged. Could you please create a new tag so these changes can be pulled via Composer?

PR #552, opened 2019-09-19
BUG Don't use unsecure cookies if session is secure

### LINKED ISSUE #534: Cookie does not respect cookie_secure and is not HttpOnly
TractorCow\Fluent\Middleware\DetectLocaleMiddleware.php:201

The Cookie::set() call sets

a) secure to false
- seems like this should respect SilverStripe\Control\Session.$cookie_secure
- $cookie_secure is set to true by default on CWP for non-dev environments

b) httpOnly is false
- this is different from the default value of true in SilverStripe\Control\Cookie::set()
- seems as though this should be true, or we should be able to configure this in yml? (or maybe there's a reason why it's not httpOnly?)

tractorcow:
Let http-only cookies be configurable, but default to secure
Fixes #534

### DIFF
--- a/src/Middleware/DetectLocaleMiddleware.php
+++ b/src/Middleware/DetectLocaleMiddleware.php
@@ -3,8 +3,10 @@
 namespace TractorCow\Fluent\Middleware;
 
 use SilverStripe\Control\Cookie;
+use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\Middleware\HTTPMiddleware;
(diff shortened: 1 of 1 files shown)


robbieaverill:
Are you able to get the builds green?

tractorcow:
The build is green; only the /push failed because I accidentally pushed to upstream, and not my fork.

tractorcow:
Ah I can't delete the /push build. just ignore it sorry.

robbieaverill:
Oh that’s what it is! I was looking on my phone - didn’t you even check the changes :-D

PR #556, opened 2019-10-15
BUG Support localised types specified as explicit FQN class names

tractorcow:


tractorcow:
Thanks @robbieaverill !

PR #557, opened 2019-10-16
Add docs for adding fluent to an existing project

wernerkrauss:


tractorcow:
I'm going to look at developing some more tools for managing localisations (e.g. copying, clearing, and so on). Maybe some tools in-cms would be helpful. :) that publishall tip is great thanks @wernerkrauss

PR #558, opened 2019-10-21
addFluentBadge should return existing badge

wernerkrauss:
Fixes a bug where an existing badge (like "modified") was destroyed in unversioned, translated data objects

### DIFF
--- a/src/Extension/FluentBadgeExtension.php
+++ b/src/Extension/FluentBadgeExtension.php
@@ -58,7 +58,7 @@ class FluentBadgeExtension extends Extension
         if (!FluentState::singleton()->getLocale()
             || !$record->has_extension(FluentVersionedExtension::class)
         ) {
-            return null;
+            return $badgeField;
         }
 
         $fluentBadge = $this->getBadge($record);


src/Extension/FluentBadgeExtension.php
```diff
@@ -58,7 +58,7 @@ protected function addFluentBadge($badgeField, DataObject $record)
         if (!FluentState::singleton()->getLocale()
             || !$record->has_extension(FluentVersionedExtension::class)
         ) {
-            return null;
+            return $badgeField;
```
robbieaverill: Just wondering, do we need to check whether the record has FluentExtension instead?

wernerkrauss: yes, we need to check for fluent VERSIONED extension, cause the code checks for `$record->existsInLocale($currentLocale->Locale)`, which is unfortunately only available for versioned records - not for unversioned records

tractorcow: We probably should implement this method there too.

tractorcow: I don't understand exactly what's going on here. If this condition fails through, we are appending both $fluentBadge and $badgeField together? What is this code trying to do there?

tractorcow: I feel like we could also fix this issue by updateBreadcrumbs wrapping the setField() in an `if ($newBadge)` as well.

tractorcow: Does anything even call `updateBadge`? @robbieaverill do you have any context? I think you wrote this code. :)

tractorcow:
I had concerns, but this PR fixes the issue, so my concerns can be added to another PR if I feel like making it later. :)

PR #562, opened 2019-10-29
WIP Initial commit of new localisation menu

tractorcow:
Just getting the look and feel right

image

Not sure how to fit descriptions in. Maybe tooltips that show on hover?

If I get https://github.com/silverstripe/silverstripe-framework/pull/9305 merged I might work on adding JS tooltips to each option.


tractorcow:
I have a feeling I'm going to need to set a custom template for this menu.

clarkepaul:
Looks cool @tractorcow, any chance you could use the ReactStrap [dropdown menu](https://silverstripe.github.io/silverstripe-admin/?knob-Search=true&selectedKind=Admin%2FActionMenu&selectedStory=Single&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) ? I realise this might not function in this area but would align more to the way things are progressing.

tractorcow:
I'd like to but it might be a bit of work to migrate to a react app.

tractorcow:
I really wanted a user friendly menu like the below

image

clarkepaul:
Bootstrap HTML/CSS might be the next best thing if react is not do-able? 
You can add text descriptions below the links by adding something like this... <p class="pl-4 pr-4 text-muted">
image.

It's just I'm trying to phase out all older UI components if possible.

tractorcow:
I think the descriptions are a bit ugly, so instead I'll scale it back, and get it working well. Then I'll look at building it back up as you have suggested. :)

At the moment new menu has been reworded a bit. I've tried to make it match the existing menu options, so it's more clear that this is the global-equivalent of the same actions.

image

tractorcow:
I'm also going to build up a proper gridfield for each record, so that you can view the draft / live status of the record in each locale. The gridfield will also have individual actions (E.g. "publish this locale", "unpublish this locale") without having to navigate through the left hand dropdown.

Screenshots pending implementation. :)

tractorcow:
Screenshot of WIP localisation information grid

image

Eventually I'll have custom actions for each locale (publish this locale, unpublish, etc)... and add some nice colours / icons for each status.

@clarkepaul what are your thoughts?

tractorcow:
Note that this grid doesn't supersede the menu I was working on above; The idea is that these features complement one another.

cc @robbieaverill , @prij who I also want feedback on :)

clarkepaul:
Couldn't you just have the last column header as State, and then use Draft, Modified, Published/Live?
Perhaps you are trying to avoid using Modified because for the US locale it can't be draft and live or maybe I am understanding it wrong.

tractorcow:
I am thinking about putting the version. E.g. if you have the same version on two locales, it can show up in that cell. I'm not 100% sure about how to make it work though.

tractorcow:
You are right though, I could use a single state. I wonder if that might add another layer of complexity, since people then need to understand what each term means. I'm using the space for extra columns because it's pretty much free space otherwise so why not fill it out. :)

tractorcow:
Sample options available on gridfield:

image


image

tractorcow:
Better screenshot in-situ

image

tractorcow:
Some more UX @clarkepaul for you to critique. :)

image

tractorcow:
I'm going to shift this PR from my fork to a develop branch; That way other people can more easily modify / contribute to it.

tractorcow:
Please go to https://github.com/tractorcow-farm/silverstripe-fluent/pull/577 if you are following this PR.

PR #573, opened 2019-11-07
BUG Fix issue with localisations being randomly deleted

### LINKED ISSUE #570: Wrong record is being deleted from _Localised table
We are experiencing an issue with the handling of deletes in the latest version(s) of Fluent. When the latest record of for instance a Page is deleted, the deteleBaseRecord is being called. 
This function deletes the record from several tables, and does this based on ID. This goes wrong for the _Localised table. There the selection should be on RecordID instead.

How to reproduce:
- fresh install of SS4 + Fluent ^4 (tested on 4.4.1 and 4.4.3)
- create some Locales
- add some pages, and save them in multiple languages (to make the ID higher than the RecordID in the version table) 
- Now; add a new page via the CMS
- Remember the ID (for instance id=16)
- Check which record is in SiteTree_Localised on ID=16
- Delete the page via the CMS
- See record ID=16 is gone in SiteTree_Localised (instead of RecordID=16)

The snippet below fixes the issue, but probably not in the most elegant way... 🤕 

```
    protected function deleteBaseRecord(DataObject $record)
    {

        //... (collapsed)

        foreach ($queriedTables as $table) {
            
            if (strpos(strtolower($table), '_localised') > 0 ) {
                $delete = SQLDelete::create("\"{$table}\"", array('"RecordID"' => $record->ID));
            } else {
                $delete = SQLDelete::create("\"{$table}\"", array('"ID"' => $record->ID));
            }
```

tractorcow:
Fixes #570

note: This is a critical fix, and once this is tagged (4.4.4) everyone on an older 4.4 version must upgrade immediately.

### DIFF
--- a/src/Model/Delete/DeleteRecordPolicy.php
+++ b/src/Model/Delete/DeleteRecordPolicy.php
@@ -5,6 +5,7 @@ namespace TractorCow\Fluent\Model\Delete;
 use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\Queries\SQLDelete;
+use TractorCow\Fluent\State\FluentState;
 
 /**
  * A policy that deletes the root record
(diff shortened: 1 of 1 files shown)


tractorcow:
thanks @baukezwaan for identifying and reporting this issue.

tractorcow:
4.4.4 is tagged, thanks @robbieaverill :D

PR #584, opened 2020-01-08
#582 Alternate href added to $MetaTags even if no localised version i…

### LINKED ISSUE #582: Alternate href added to $MetaTags even if no localised version is published
Fluent always adds alternate hrefs to the $MetaTags, even though there is no localised version in another language published.

How to reproduce :
1> Set frontend_publish_required to true in config
2> Add 2 locales
3> Add and publish page in the base locale, but do not publish it in the 2nd locale
4> Look at output of $MetaTags of base language page, it will add an alternate link containing just the base domain + locale, without a page segment. E.g. :  www.domain.com/en

Cas-terRiet-Directlease:
Fix #582 Alternate href added to $MetaTags even if no localised version is published bug

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -853,7 +853,7 @@ class FluentExtension extends DataExtension
      */
     public function getLinkingMode($locale)
     {
-        if ($this->owner->hasMethod('canViewInLocale') && !$this->owner->canViewInLocale($locale)) {
+        if ($this->owner->hasMethod('isLocalisedInCurrentStage') && !$this->owner->isLocalisedInCurrentStage($locale)) {
             return 'invalid';
         }
 
--- a/src/Extension/FluentVersionedExtension.php
+++ b/src/Extension/FluentVersionedExtension.php
@@ -298,6 +298,17 @@ class FluentVersionedExtension extends FluentExtension
     }
 
     /**
+     * Check if this record exists (in current stage) in this locale
+     *
+     * @param string $locale
+     * @return bool
+     */
+    public function isLocalisedInCurrentStage($locale = null)
+    {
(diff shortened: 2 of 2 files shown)


Cas-terRiet-Directlease:
I think this is a fix for this problem, I only can't figure out what the problem is within the PHP test. can you guys maybe look at the code to see what the problem is?

tractorcow:
In `develop` branch there is a new helper you can use here. `RecordLocale`, and the best method here is the `IsPublished` method, which will check if a record is both published, and visible in that locale. It includes all the business logic you need, including handling both localisation, filtering, and configuration based rules.

tractorcow:
https://github.com/tractorcow-farm/silverstripe-fluent/blob/develop/src/Model/RecordLocale.php#L252

tractorcow:
Possibly we could do this

```diff
-<% if $LinkingMode != 'invalid' %>
+<% if $IsPubilshed %>
```

In FluentSiteTree_MetaTags.ss

tractorcow:
@Cas-terRiet-Directlease test `develop` branch; Let me know if you are happy with this fix.

https://github.com/tractorcow-farm/silverstripe-fluent/commit/ea8619a5d24280b14fbf47716bd6a34a0860500f

Note: Develop isn't stable yet, so be careful about updating your site to this version. :)

We can additionally post another fix to the prior version, but I'll try to keep the fixes as close to one another as possible.

Cas-terRiet-Directlease:
@tractorcow this fix also works for us, thank you :D

tractorcow:
Thanks @Cas-terRiet-Directlease ; I'm going to close this for now, because I'd like to prioritise upgrading over changing current behaviour. I will see if I can back-port the fix without possibly breaking existing logic too much. :)

tractorcow:
note: Anyone can get the new version with `dev-develop` as their composer dependency.

Issue #587, opened 2020-01-30
TODO Write up contributor access guidelines


tractorcow:
cc @chillu I promised to write this up to help guide future contributors, and current core team.

PR #592, opened 2020-02-04
Update contributing guidelines

tractorcow:
Closes https://github.com/tractorcow-farm/silverstripe-fluent/issues/587

cc @chillu

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,28 @@
 # Contributing to Fluent
 
-Before submitting bugs, please make sure to run the dev/tasks/FluentValidateTask task to debug any issues locally.
+As an open source project, fluent is open for anyone to use, fork, modify, or contribute back to.
+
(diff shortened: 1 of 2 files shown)


chillu:
Yeah that all sounds reasonable, thanks for taking the time to write it up! Do you see a difference between "maintainers" and the "contributing team" (as you phrased it)? Technically, everyone participating in this module is a contributor, I think calling it a "maintainer" is a bit less ambiguous? In core, we've carved out that role of a "contributing committer", but that's also a bit strange naming.

tractorcow:
Na it's just a wording. :) I'll consolidate to use the popular nomenclature. :)

chillu:
OK I've merged, as a way of +1'ing. I'll also circulate it with the product devs here at the mothership.

tractorcow:
Thanks :D

brynwhyman:
FWIW, I think this a nice set of guidelines! Totally reasonable.

PR #597, opened 2020-03-04
Latest version issue fix

manja:
Scenario: You have a data object that's both versioned and localised. Let's say that locales available are en_US and nl_NL. You first save english, and after some time, dutch version. Later, somewhere it the code you want to get latest version of that data object.

Expected behaviour: Depending of the locale set in FluentState, you will get latest version written in that locale.
Actual behaviour: if latest version is written in same locale which is set, you will get it. Otherwise returns false.

Or, following example above, if locale is set to nl_NL, you'll get the latest version. If it's en_US - no luck - since there is a version after it, but in different locale.

This PR addresses that issue and makes Versioned return latest version in requested locale, no matter if there are any newer versions in other locales.


tractorcow:
Nice fix, smart and well thought out. Thanks @manja for this.

PR #599, opened 2020-03-19
Additional 5.0 alpha features

tractorcow:
- Add FluentIsolatedExtension (objects that belong in one locale only)
- Add support for `x-default` hreflang
- Bugfixes for badges and saving localised dataobjects

### DIFF
--- a/src/Control/LocaleAdmin.php
+++ b/src/Control/LocaleAdmin.php
@@ -3,9 +3,11 @@
 namespace TractorCow\Fluent\Control;
 
 use SilverStripe\Admin\ModelAdmin;
+use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Security;
 use SilverStripe\View\Requirements;
+use Symbiote\GridFieldExtensions\GridFieldOrderableRows;
 use TractorCow\Fluent\Extension\FluentDirectorExtension;
 use TractorCow\Fluent\Extension\FluentMemberExtension;
 use TractorCow\Fluent\Model\Domain;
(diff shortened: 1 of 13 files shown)


src/Control/LocaleAdmin.php
```diff
@@ -36,6 +38,21 @@ protected function init()
...
+    public function getEditForm($id = null, $fields = null)
+    {
+        $form = parent::getEditForm($id, $fields);
+
+        // Add sortable field to locales
+        if ($this->modelClass === Locale::class) {
+            /** @var GridField $listField */
+            $listField = $form->Fields()->dataFieldByName($this->sanitiseClassName($this->modelClass));
+            $config = $listField->getConfig();
+            $config->addComponent(new GridFieldOrderableRows('Sort'));
```
robbieaverill: Is this already a dependency of fluent?

src/Model/Locale.php
```diff
@@ -250,6 +273,14 @@ public function getCMSFields()
                         'Note: Per-domain specific locale can be assigned on the Locales tab'
                         . ' and will override this value for specific domains.'
                     )),
+                CheckboxField::create(
+                    'UseDefaultCode',
+                    _t(__CLASS__ . '.USE_X_DEFAULT', 'Use {code} as hreflang', ['code' => self::X_DEFAULT])
```
robbieaverill: Is there maybe a way we can avoid using `hreflang` here? It might be confusing for users

src/Model/Locale.php
```diff
@@ -250,6 +273,14 @@ public function getCMSFields()
...
+                )
+                    ->setDescription(_t(
+                        __CLASS__ . '.USE_X_DEFAULT_DESCRIPTION',
+                        'Use of this code indicates to search engine that this is a non-localised global landing page'
```
robbieaverill:
```suggestion
                        'Use of this code indicates to search engines that this is a non-localised global landing page'
```

src/Control/LocaleAdmin.php
tractorcow: Yep `"symbiote/silverstripe-gridfieldextensions": "^3.1",`

src/Model/Locale.php
tractorcow: Ah, possibly, except that's literally all it's used for. :P I can look at other description maybe. Like `SEO Language code`?

robbieaverill: Maybe both e.g. `hreflang (SEO language code)` ?

tractorcow:
So nice to see you again @robbieaverill :)

src/Model/Locale.php
tractorcow: Changed.

src/Model/Locale.php
tractorcow: Accepted.

PR #615, opened 2020-05-20
BUG Properly handle x-default for home page tags

tractorcow:


tractorcow:
This changes some behaviour. Now the canonical url for home pages is `/locale/`, even if home page.

The only exceptions are now if the domain has only one single locale, or if the entire locale has its prefix disabled (and is default).

PR #617, opened 2020-05-27
BUG Fix compatibility issues with mysql 5.7 and above

### LINKED ISSUE #257: Fluent extension causes MySQL Errors with version >= 5.7.5
The problem arises when the [`FluentExtension` uses `COALESCE`](https://github.com/tractorcow/silverstripe-fluent/blob/master/code/extensions/FluentExtension.php#L660) on colums that should be sorted.

Say you have a translatable DataObject `Article`, with a `Title` field and you want to sort by `Title`. It works in MySQL 5.6, but generates the following error on MySQL 5.7.13:

    Article::get()->sort('Title'); // using the localized column name doesn't help either

> Expression #1 of ORDER BY clause is not in SELECT list, references column 'Article.Title_en_US' which is not in SELECT list; this is incompatible with DISTINCT

I've looked around for the same issue and it seems to be caused by the following default setting: https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by

Multiple sources state, that it can be fixed by setting `sql-mode` to a custom value that doesn't include `ONLY_FULL_GROUP_BY`, but I was unable to make it work locally. I'm assuming that a lot of people don't have full control over their mysql-server settings (depending on hosting-provider), so I guess this would be better fixed in code?

The only workaround so far seems to be to disable the coalescing of fields, by setting the sortable field to be nullable. Eg.

```yml
Article:
  extensions:
    - FluentExtension
  translate:
    - Title
    - Content
nullable_fields:
  - Title # adding this makes the error go away!
```

tractorcow:
Fixes #257

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -1,5 +1,9 @@
 <?php
 
 use SilverStripe\Dev\Deprecation;
+use SilverStripe\ORM\DB;
 
 Deprecation::notification_version('4.0.0', 'tractorcow/silverstripe-fluent');
+
+// Credit to https://github.com/sunnysideup thank you for the mysql fix
+DB::query("SET SESSION sql_mode='REAL_AS_FLOAT,PIPES_AS_CONCAT,ANSI_QUOTES,IGNORE_SPACE';");


sminnee:
FYI we're looking at introducing a sql_mode yml config option here and so it might be worth using that instead:
https://github.com/silverstripe/silverstripe-framework/pull/9721

It would mean that you need Silverstripe 4.7 to support MySQL 5.7, though.

sminnee:
Alternatively, fluent could be patched to not use invalid queries. Notably, that would mean that fluent works with PostgreSQL (I imagine that it's broken at the moment)

sminnee:
Something that is also a little confusing to me is how the tests are currently passing, when they already run on MySQL 5.7. Is that because the test suite doesn't cover the broken functionality? In that case, it might be valuable to cover it with a test.

tractorcow:
It's not so much that fluent is generating incorrect queries (at least not anymore), it's more that there's always been a SQL surface area that could trigger this error.

The specific issue due to sort https://github.com/tractorcow-farm/silverstripe-fluent/issues/257 was fixed since then, but I'm aware there are still situations where manual queries can still get trapped.

Fluent does rely on users being super careful when writing queries (i.e. you pretty much MUST use `"Table"."Column"`), but in practice we have to accept that hand crafted SQL may not always be that precise. This patch is more to soften the blow of these cases, rather than guarantee that no invalid queries ever reach the DB.

That's the trouble with using raw SQL strings as a query representation. :P

sminnee:
Ah, right. Maybe it does work on PGSQL then... (I just commented suggesting it wouldn't on the other PR)

tractorcow:
I actually built fluent originally on postgres, then worked backwards to mysql. ;P

PR #618, opened 2020-05-27
BUG Fix compatibility issues with mysql 5.7 and above (fluent 4)

### LINKED ISSUE #257: Fluent extension causes MySQL Errors with version >= 5.7.5

tractorcow:
Fixes #257

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -1,5 +1,9 @@
 <?php
 
 use SilverStripe\Dev\Deprecation;
+use SilverStripe\ORM\DB;
 
 Deprecation::notification_version('4.0.0', 'tractorcow/silverstripe-fluent');
+
+// Credit to https://github.com/sunnysideup thank you for the mysql fix
+DB::query("SET SESSION sql_mode='REAL_AS_FLOAT,PIPES_AS_CONCAT,ANSI_QUOTES,IGNORE_SPACE';");


_config.php
```diff
@@ -1,5 +1,9 @@
 <?php
 
 use SilverStripe\Dev\Deprecation;
+use SilverStripe\ORM\DB;
 
 Deprecation::notification_version('4.0.0', 'tractorcow/silverstripe-fluent');
+
+// Credit to https://github.com/sunnysideup thank you for the mysql fix
+DB::query("SET SESSION sql_mode='REAL_AS_FLOAT,PIPES_AS_CONCAT,ANSI_QUOTES,IGNORE_SPACE';");
```
tractorcow, line 9: Same problem as https://github.com/tractorcow-farm/silverstripe-fluent/pull/617

Need to conditional and apply only to mysql, not postgres / sqlite etc...

sminnee:
FYI we're looking at introducing a sql_mode yml config option here and so it might be worth using that instead:
https://github.com/silverstripe/silverstripe-framework/pull/9721

It would mean that you need Silverstripe 4.7 to support MySQL 5.7, though.

tractorcow:
I much prefer the config option to this; I don't want to have to conditionally run this query based on DB backend.

sminnee:
Note that I assume Fluent doesn't support PGSQL at all right now as there's no way of turning off strict group by on PGSQL.

tractorcow:
It doesn't support postgresql any less than framework does. Until a while back the tests were running postgres (they got turned off at some point I can't recall).

tractorcow:
It looks like the original issue was due to COALESCE, which fluent probably should not be using anymore for conditions, since locale row should be deterministic (the localisation row exists, or it doesn't).

Here's the comment on FluentExtension::localiseCondition().

```
* Note that unlike localiseSelect, this uses a simple COLASECLE() for performance and to reduce
* overall query size.
```

We probably should just remove COALESCE anyway, but better to turn on postgres tests before we go making any assumptions. :)

PR #622, opened 2020-06-18
Various fixes for Alpha 5.

mfendeksilverstripe:
* Various fixes for Alpha 5
* this is a batch branch where all fixes are dumped
* individual fixes should be split into separate PRs after reaching project bespoke project goal

## This is still WIP, DO NOT MERGE!

### DIFF
--- a/src/Forms/CopyLocaleAction.php
+++ b/src/Forms/CopyLocaleAction.php
@@ -75,7 +75,7 @@ class CopyLocaleAction extends BaseAction
      */
     public function handleAction(GridField $gridField, $actionName, $arguments, $data)
     {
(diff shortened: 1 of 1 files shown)


tractorcow:
@mfendeksilverstripe I've merged your first PR. 

I'd also accept changes to the labels / terminology if you felt like taking charge of it. I understand we wanted to adjust some of the verbage.

Do you want to do a skype call some time so we can catch up with how your team is using fluent? I'd like to involve you in the beta release (which should not be too far away).

mfendeksilverstripe:
Thanks for the feedback @tractorcow . We are setting up a fairly new project with Fluent so I'll probably have more feedback going your way. The project size is small so we should be able to cover it in like two weeks or so. Happy to catch up with you after that ;-)

PR #623, opened 2020-06-18
Redundant copy locale actions fix (alpha 5).

mfendeksilverstripe:
# Redundant copy locale actions fix (alpha 5)

* `CopyLocaleAction` component is created as many times as there are locales available, see code snippet below (from `FluentExtension`):

```
foreach (Locale::getCached() as $locale) {
    $config->addComponents([
        new CopyLocaleAction($locale->Locale, true),
        new CopyLocaleAction($locale->Locale, false),
    ]);
}
```

* these components have slightly different UI but they all subscribe to the same action
* whenever either `copy to` or` copy from` action is executed, all of the components will execute the action
* this change-set fixes this by adding additional condition to make the components aware of the locale they are responsible for

## Related issues

https://github.com/tractorcow-farm/silverstripe-fluent/issues/624

### DIFF
--- a/src/Forms/CopyLocaleAction.php
+++ b/src/Forms/CopyLocaleAction.php
@@ -75,7 +75,7 @@ class CopyLocaleAction extends BaseAction
      */
     public function handleAction(GridField $gridField, $actionName, $arguments, $data)
     {
-        if (!in_array($actionName, ['fluentcopyto', 'fluentcopyfrom'])) {
+        if (!$this->validateAction($actionName, $arguments['FromLocale'], $arguments['ToLocale'])) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


tractorcow:
Thanks, I checked this and found that I had some incorrect assumptions when I wrote this code. I wrote up my findings at https://github.com/tractorcow-farm/silverstripe-fluent/issues/624#issuecomment-646320236.

src/Forms/CopyLocaleAction.php
```diff
@@ -186,4 +186,25 @@ public function getGroup($gridField, $record, $columnName)
...
+     * Ensures that action is executed only once and not once per per locale
+     *
+     * @param string $actionName
+     * @param string $fromLocale
+     * @param string $toLocale
+     * @return bool
+     */
+    private function validateAction($actionName, $fromLocale, $toLocale)
+    {
+        if ($actionName === 'fluentcopyto' && $this->otherLocale === $fromLocale) {
```
tractorcow, line 200: Hm I have a feeling this may be backwards.

```php
    /**
     * If true, this is "copy to $otherlocale". If false "copy from $otherLocale"
     *
     * @var bool
     */
    protected $isTo;
``` 

Let me test :)

tractorcow: It would have worked, but you would end up with the `nz to us` action handling the `us to nz` event (still only one action though). I pushed up a fix to flip the from / to locales, and tested that it worked. :) I can't imagine there would be many cases it would effectively caused a bug, but at least the code is consistent.

tractorcow: See https://github.com/tractorcow-farm/silverstripe-fluent/commit/81eff09f554551d8a9999c6c64f0286d18c5dbcb

tractorcow:
Really good work @mfendeksilverstripe I'm impressed at how well you quickly became an expert at this convoluted code! (fluent 5 is pretty complex huh... sorry).

tractorcow:
https://github.com/tractorcow-farm/silverstripe-fluent/releases/tag/5.0.0-alpha8 tagged

tractorcow:
@mfendeksilverstripe  I sent you an invitation to give you write access to the repo.

Issue #621, opened 2020-06-05
Latest 5 alpha testing feedback


mfendeksilverstripe:
# Feedback for 5.0.0-alpha7

I'm providing some feedback on the latest alpha version.

### Locales

* NZ global default
* US - fallback to NZ
* AU - fallback to NZ
* JP - no fallback

### Pages

Test page localised in NZ and AU.

### Terminology

**Unlocalise** - should be used when we archive a locale and there is at least one other locale left
**Archive** - should be used to indicate that the last locale is being archived

It's really important to make clear to content author the distinction between these two actions as **Archive** is potentially much more destructive than the **Unlocalise**.

#### Localise widget UI

![Screen Shot 2020-06-05 at 11 54 14 AM](https://user-images.githubusercontent.com/26395487/83821818-811bd400-a723-11ea-9dd1-fb7ba7e844c3.png)

`Clear from all except current locale` should be changed to `Unlocalise page in all other locales`.
`Copy current locale to other locales` should be changed to `Localise page in all other locales using the content of this page`.

### More actions widget

This widget doesn't work well with Fluent as it's reading data from the base record instead of the localised record.

![Screen Shot 2020-06-05 at 12 31 03 PM](https://user-images.githubusercontent.com/26395487/83823613-80d20780-a728-11ea-8886-ec48ac24c4bd.png)

Base record always contains data from the most recently updated locale. I understand the technical reason for it but from the point of content author this seems really random and misleading.

For example in my test setup I have a page which is unpublished in NZ and later published in AU. If you go to the JP which does't have localised content nor it inherits content it will still show Last edited and publish dates from the AU locale as the that data is on the base record. You get the same result on the US locale.

Additionally, some of the actions should not be even displayed for this page such as `Cancel draft changes` and `Unpublish`

There are several option on how to tackle this. For example:

* apply extension after the vanilla fields are created and update them as needed
* override `isPublished` and other methods (this would require some refactoring of the `Versioned` module)

### How to localise a page

Currently, there are 3 ways how to create / publish a page localisation. This is a bit confusing as these options overlap. A simpler more consistent UI is preferred. First, let's have a look at current state.

#### Option 1 - Localise by action buttons

![Screen Shot 2020-06-05 at 12 09 04 PM](https://user-images.githubusercontent.com/26395487/83822517-5df22400-a725-11ea-9887-1d5c261864be.png)

#### Option 2 - Batch actions

[...]

PR #627, opened 2020-06-22
NEW: Localisation manager UI improvements.

mfendeksilverstripe:
# Localisation manager UI improvements / configuration improvements

These changed can be split into separate pull requests and processed individually.

## Configuration changes

New configuration options were added.

### Batch actions configuration

* Fluent menu batch actions are potentially quite destructive and may not be needed in some projects
* added an option to disable this menu (menu is enabled by default)
* following actions are not shown if the menu is disabled batch unlocalise, batch localise, batch unpublish, batch publish, batch archive

### Copy to / from locale configuration

* copy content from one locale to another can be achieved by either action and this may be confusing for the content authors (simplicity over flexibility)
* this new configuration allows the developer to choose which actions are desired
* both actions are enabled by default (previous behaviour)

## Terminology

These change are related to how we name things.

* `isLive` - refers to an object being available on the frontend
* `IsPublished` refers to an object which has a a localised record in its live table

### Features

New features were added.

## Localisation state

* current UI is a little bit confusing as it's mixing multiple concepts together
* concern 1 - where data is available and in which state, this information is mostly useful for operating copy to / from actions
* concern 2 - what is displayed on the frontend
* new UI separates these concerns
* `Draft` column renamed to `Status`
* `Published` column renamed to `Live`
* `Source` column added to reflect where the content is coming from (which locale)

#### Non-versioned UI

* `Status` column shows either Localised or Not localised
* `Source` show the locale which is the source of the content

Screen Shot 2020-06-23 at 11 24 06 AM

#### Versioned UI

* `Status` column shows either Draft, Published, Modified or Not localised
* `Source` show the locale which is the source of the content
* `Live` column shows either Yes or No

Screen Shot 2020-06-23 at 11 22 45 AM

### Detect modified state

* added `stagesDifferInLocale` which is a Fluent friendly version of `stagesDiffer`

### Badges UI

* badges now only reflect localisation state which makes the UI less confusing (separation of concerns - don't mix versioned state and localisation state)
* any terminology which is related to Versioned state has been removed as the information is represented elsewhere
* new badge states are: localised (has localised record), inherited (inherits from another locale) and unknown (no localised record and no inheritance)
* messages updated accordingly

### copy to draft, copy & publish action configuration

* added an option to enable / disable localise actions (copy to draft, copy & publish), enabled by default
* these actions may not be needed as the same result can be achieved by using localisation manager

### Global template helper for current locale

* current locale object helper method added
* this is useful for situation where locale object is needed in the template for cases like outputting current locale Title or other data

### Publish state fix for multiple actions / areas

* Following actions now reflect the published state based on localised record instead of base record (publish action, information panel, unpublish action, archive action)

### Non-versioned UI improvements

Pushed up Non-versioned UI improvements:

* save action is changed to Localise in case the record is not localised in current locale (edits only)

Screen Shot 2020-06-29 at 1 54 49 PM

* delete action is changed to Unlocalise in case he record is localised and i's not the last locale

Screen Shot 2020-06-29 at 1 55 01 PM

## Related issues

https://github.com/tractorcow-farm/silverstripe-fluent/issues/621

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -870,6 +870,30 @@ class FluentExtension extends DataExtension
     }
 
     /**
+     * Get list of locales where record is localised in
+     *
+     * @return array
+     */
+    public function LocaleInstances()
+    {
+        $locales = [];
+
+        foreach (Locale::getCached() as $locale) {
+            /** @var Locale $locale */
+            $info = $this->owner->LocaleInformation($locale->getLocale());
+            $source = $info->SourceLocale();
+
+            if ($source->Locale !== $locale->Locale) {
+                continue;
+            }
+
+            $locales[] = $locale;
+        }
+
+        return $locales;
+    }
+
+    /**
      * Return the linking mode for the current locale and object
      *
      * @param string $locale
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -874,14 +874,14 @@ class FluentExtension extends DataExtension
      *
      * @return array
      */
-    public function LocaleInstances()
+    public function getLocaleInstances(): array
     {
         $locales = [];
 
         foreach (Locale::getCached() as $locale) {
             /** @var Locale $locale */
             $info = $this->owner->LocaleInformation($locale->getLocale());
-            $source = $info->SourceLocale();
+            $source = $info->getSourceLocale();
 
             if ($source->Locale !== $locale->Locale) {
                 continue;
@@ -1102,32 +1102,32 @@ class FluentExtension extends DataExtension
         $summaryColumns['Status'] = [
             'title' => 'Status',
             'callback' => function (Locale $object) {
-                if ($object->RecordLocale()) {
-                    if ($object->RecordLocale()->IsDraft()) {
-                        return _t(self::class . '.LOCALISED', 'Localised');
-                    }
-
-                    return _t(self::class . '.NOTLOCALISED', 'Not localised');
+                if (!$object->RecordLocale()) {
+                    return '';
                 }
 
-                return '';
+                if ($object->RecordLocale()->IsDraft()) {
+                    return _t(self::class . '.LOCALISED', 'Localised');
+                }
+
+                return _t(self::class . '.NOTLOCALISED', 'Not localised');
             }
         ];
 
         $summaryColumns['Source'] = [
             'title' => 'Source',
             'callback' => function (Locale $object) {
-                if ($object->RecordLocale()) {
-                    $sourceLocale = $object->RecordLocale()->SourceLocale();
-
-                    if ($sourceLocale) {
-                        return $sourceLocale->getLongTitle();
-                    }
-
-                    return _t(self::class . '.NOSOURCE', 'No source');
+                if (!$object->RecordLocale()) {
+                    return '';
                 }
 
-                return '';
+                $sourceLocale = $object->RecordLocale()->getSourceLocale();
+
+                if ($sourceLocale) {
+                    return $sourceLocale->getLongTitle();
+                }
+
+                return _t(self::class . '.NOSOURCE', 'No source');
             }
         ];
     }
@@ -1165,13 +1165,13 @@ class FluentExtension extends DataExtension
         foreach (Locale::getCached() as $locale) {
             if ($copyToLocaleEnabled) {
                 $config->addComponents([
-                    new CopyLocaleAction($locale->Locale, true),
+                    CopyLocaleAction::create($locale->Locale, true),
                 ]);
             }
 
             if ($copyFromLocaleEnabled) {
                 $config->addComponents([
-                    new CopyLocaleAction($locale->Locale, false),
+                    CopyLocaleAction::create($locale->Locale, false),
                 ]);
             }
         }
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -155,6 +155,30 @@ class FluentExtension extends DataExtension
     private static $data_exclude = [];
 
     /**
+     * Enable copy to locale action in the localisation manager
+     *
+     * @config
+     * @var bool
+     */
+    private static $copy_to_locale_enabled = true;
+
+    /**
+     * Enable copy from locale action in the localisation manager
+     *
+     * @config
+     * @var bool
+     */
+    private static $copy_from_locale_enabled = true;
+
+    /**
+     * Enable batch actions in the edit form
+     *
+     * @config
+     * @var bool
+     */
+    private static $batch_actions_enabled = true;
+
+    /**
      * Cache of localised fields for this model
      */
     protected $localisedFields = [];
@@ -1051,10 +1075,35 @@ class FluentExtension extends DataExtension
      */
     public function updateLocalisationTabColumns(&$summaryColumns)
     {
-        $summaryColumns['IsDraft'] = [
-            'title'    => 'Saved',
+        $summaryColumns['Status'] = [
+            'title' => 'Status',
             'callback' => function (Locale $object) {
-                return $object->RecordLocale()->IsDraft() ? 'Saved' : '';
+                if ($object->RecordLocale()) {
+                    if ($object->RecordLocale()->IsDraft()) {
+                        return 'Localised';
+                    }
+
+                    return 'Not localised';
+                }
+
+                return '';
+            }
+        ];
+
+        $summaryColumns['Source'] = [
+            'title' => 'Source',
+            'callback' => function (Locale $object) {
+                if ($object->RecordLocale()) {
+                    $sourceLocale = $object->RecordLocale()->SourceLocale();
+
+                    if ($sourceLocale) {
+                        return $sourceLocale->getLongTitle();
+                    }
+
+                    return 'No source';
+                }
+
+                return '';
             }
         ];
     }
@@ -1085,12 +1134,22 @@ class FluentExtension extends DataExtension
             new GroupActionMenu(GridField_ActionMenuItem::DEFAULT_GROUP)
         ]);
 
+        $copyToLocaleEnabled = $this->owner->config()->get('copy_to_locale_enabled');
+        $copyFromLocaleEnabled = $this->owner->config()->get('copy_from_locale_enabled');
+
         // Add each copy from / to
         foreach (Locale::getCached() as $locale) {
-            $config->addComponents([
-                new CopyLocaleAction($locale->Locale, true),
-                new CopyLocaleAction($locale->Locale, false),
-            ]);
+            if ($copyToLocaleEnabled) {
+                $config->addComponents([
+                    new CopyLocaleAction($locale->Locale, true),
+                ]);
+            }
+
+            if ($copyFromLocaleEnabled) {
+                $config->addComponents([
+                    new CopyLocaleAction($locale->Locale, false),
+                ]);
+            }
         }
     }
 }
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -1080,10 +1080,10 @@ class FluentExtension extends DataExtension
             'callback' => function (Locale $object) {
                 if ($object->RecordLocale()) {
                     if ($object->RecordLocale()->IsDraft()) {
-                        return 'Localised';
+                        return _t(self::class . '.LOCALISED', 'Localised');
                     }
 
-                    return 'Not localised';
+                    return _t(self::class . '.NOTLOCALISED', 'Not localised');
                 }
 
                 return '';
@@ -1100,7 +1100,7 @@ class FluentExtension extends DataExtension
                         return $sourceLocale->getLongTitle();
                     }
 
-                    return 'No source';
+                    return _t(self::class . '.NOSOURCE', 'No source');
                 }
 
                 return '';
--- a/src/Extension/FluentSiteTreeExtension.php
+++ b/src/Extension/FluentSiteTreeExtension.php
@@ -15,6 +15,7 @@ use SilverStripe\Forms\LiteralField;
 use SilverStripe\ORM\FieldType\DBHTMLText;
 use TractorCow\Fluent\Extension\Traits\FluentAdminTrait;
 use TractorCow\Fluent\Model\Locale;
+use TractorCow\Fluent\Model\RecordLocale;
 use TractorCow\Fluent\State\FluentState;
 
 // Soft dependency on CMS module
@@ -239,15 +240,25 @@ class FluentSiteTreeExtension extends FluentVersionedExtension
                 );
             }
         } else {
-            // If frontend publishing is *not* required, then we have two possibilities.
+            // If frontend publishing is *not* required, then we have multiple possibilities.
             if (!$this->isDraftedInLocale()) {
-                // Our content hasn't been drafted or published. If this Locale has a Fallback, then content might be
-                // getting inherited from that Fallback.
-                $message = _t(
-                    __CLASS__ . '.LOCALESTATUSFLUENTINHERITED',
-                    'Content for this page may be inherited from another locale. If you wish you make an ' .
-                    'independent copy of this page, please use one of the "Copy" actions provided.'
-                );
+                $info = RecordLocale::create($this->owner, Locale::getCurrentLocale());
+
+                // Our content hasn't been drafted or published.
+                if ($info->SourceLocale()) {
+                    // If this Locale has a Fallback, then content might be getting inherited from that Fallback.
+                    $message = _t(
+                        __CLASS__ . '.LOCALESTATUSFLUENTINHERITED',
+                        'Content for this page may be inherited from another locale. If you wish you make an ' .
+                        'independent copy of this page, please use one of the "Copy" actions provided.'
+                    );
+                } else {
+                    // This locale doesn't have any content source
+                    $message = _t(
+                        __CLASS__ . '.LOCALESTATUSFLUENTUNKNOWN',
+                        'No content is available for this page. Please localise this page or provide a locale fallback.'
+                    );
+                }
             } elseif (!$this->isPublishedInLocale()) {
                 // Our content has been saved to draft, but hasn't yet been published. That published content may be
                 // coming from a Fallback.
--- a/src/Extension/FluentSiteTreeExtension.php
+++ b/src/Extension/FluentSiteTreeExtension.php
@@ -2,6 +2,7 @@
 
 namespace TractorCow\Fluent\Extension;
 
+use SilverStripe\CMS\Controllers\CMSMain;
 use SilverStripe\CMS\Controllers\RootURLController;
 use SilverStripe\CMS\Forms\SiteTreeURLSegmentField;
 use SilverStripe\CMS\Model\SiteTree;
@@ -12,7 +13,10 @@ use SilverStripe\Forms\CompositeField;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\Form;
 use SilverStripe\Forms\LiteralField;
+use SilverStripe\Forms\Tab;
 use SilverStripe\ORM\FieldType\DBHTMLText;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\View\SSViewer;
 use TractorCow\Fluent\Extension\Traits\FluentAdminTrait;
 use TractorCow\Fluent\Model\Locale;
 use TractorCow\Fluent\Model\RecordLocale;
@@ -39,6 +43,15 @@ class FluentSiteTreeExtension extends FluentVersionedExtension
      * @var bool
      */
     private static $locale_published_status_message = true;
+
+    /**
+     * Enable localise actions (copy to draft and copy & publish actions)
+     * these actions can be used to localise page content directly via main page actions
+     *
+     * @config
+     * @var bool
+     */
+    private static $localise_actions_enabled = true;
 
     /**
      * Add alternate links to metatags
@@ -202,6 +215,18 @@ class FluentSiteTreeExtension extends FluentVersionedExtension
 
         // Update specific sitetree publish actions
         $this->updateSavePublishActions($actions);
+
+        // Update specific sitetree localise actions
+        $this->updateLocaliseActions($actions);
+
+        // Update information panel (shows published state)
+        $this->updateInformationPanel($actions);
+
+        // Update the state of publish action (if needed)
+        $this->updatePublishState($actions);
+
+        // Updat unpublish and archive actions
+        $this->updateMoreOptionsActions($actions);
 
         // Add extra fluent menu
         $this->updateFluentActions($actions, $this->owner);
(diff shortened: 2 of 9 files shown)


chrispenny:
@mfendeksilverstripe looking good!

If possible, I think my preference would be to have separate columns for the `Status` and a new column for `Content Source`. Otherwise you will need to specify that something is "Draft" + where it's inherited content comes from, which I think could get confusing.

`Status` is now only concerned with telling you what the current localisation status is:
- Not localised (or [blank])
- Draft
- Published
- Modified

`Live` unchanged, indicates whether or not a page is available on the frontend:
- Yes/No

`Content Source` is only concerned with telling you where the content viewed on the frontend is coming from:
- [Locale] (indicating the Locale where the content is coming from)
- [Blank] (indicating that the page is getting it's content from the base record, which is dangerous)
There might be a better option than [Blank], but it should be represented in some way.

mfendeksilverstripe:
@chrispenny Pushed new changes, updated screenshots :)

src/Extension/FluentVersionedExtension.php
```diff
@@ -325,6 +325,83 @@ public function existsInLocale($locale = null)
...
     }
 
+    /**
+     * Check if this record has modifications in this locale
+     * Fluent friendly version of @see Versioned::stagesDiffer()
+     *
+     * @param string|null $locale
+     * @return bool
+     */
+    public function stagesDifferInLocale($locale = null)
```
tractorcow, line 335: Good feature; Can I request that this be moved to RecordLocale object?

this makes this functionality available to all objects that have any fluent extension, not only fluentversionedextension. It also gives us the chance to consolidate some of the functionality that does record and locale inspection. :)

src/Extension/FluentVersionedExtension.php
```diff
@@ -325,6 +325,83 @@ public function existsInLocale($locale = null)
...
+            return false;
+        }
+
+        // Need to check that it is actually localised
+        if (!$record->hasExtension(FluentExtension::class)) {
+            return false;
+        }
+
+        // Need to check that the localised version is versioned
+        if (!$record->hasExtension(FluentVersionedExtension::class)) {
```
tractorcow, line 357: $record is the owner extended by FluentVersionedExtension; I think these are probably redundant.

note: If you move this to RecordLocale, then it's no longer redundant. :)

src/Extension/FluentVersionedExtension.php
tractorcow: It also means that you can check this field in template loops `<% loop $Locales %><% if StagesDifferInLocale %>...`

src/Model/RecordLocale.php
```diff
@@ -293,6 +293,35 @@ public function IsPublished()
         return true;
     }
 
+    /**
+     * Check for published and localised record
+     * useful when determining what data is available in which locale
+     *
+     * @return bool
+     */
+    public function IsPublished()
```
tractorcow, line 302: Is this method just the same but with frontend_publish_required ignored?

tractorcow: Suggestion: instead of a new method, simplify (and with clear documentation) add a `$inLocale = false` property.
```php
/**
 @param bool $inLocale Set to true if this record must be published in the specified locale. If frontend_publish_required is false (and $inLocale is false) then this page is considered live.
*/
```

src/Model/RecordLocale.php
```diff
@@ -315,4 +344,48 @@ public function IsDraft()
...
+    public function IsModified()
+    {
+        if (!$this->IsPublished()) {
+            return false;
+        }
+
+        /** @var DataObject|FluentVersionedExtension $record */
+        $record = $this->getOriginalRecord();
+
+        return $record->stagesDifferInLocale($this->getLocale());
```
tractorcow, line 363: If calling FluentVersionedExtension specific calls, please check for the extension on the record first.

src/Model/RecordLocale.php
```diff
@@ -315,4 +344,48 @@ public function IsDraft()
...
+    }
+
+    /**
+     * Get the locale which is the source of content for this record
+     *
+     * @return Locale|null
+     */
+    public function SourceLocale()
```
tractorcow, line 371: High 5 for this.

src/Model/RecordLocale.php
```diff
@@ -315,4 +344,48 @@ public function IsDraft()
 
         return true;
     }
+
+    /**
+     * Check if a record is modified
```
tractorcow, line 349: * "is published, but modified" 

Sorry I'm being pedantic. :)

src/Model/RecordLocale.php
tractorcow: I wouldn't be against having a new method that simply called the other with a different flag, but not one with 90% of the same code please.

src/Extension/FluentVersionedExtension.php
```diff
@@ -447,21 +524,53 @@ public static function prepoulateIdsInLocale($locale, $dataObjectClass, $populat
...
+
+                    if ($recordLocale->IsPublished()) {
+                        return 'Published';
+                    }
+
+                    if ($recordLocale->IsDraft()) {
+                        return 'Draft';
+                    }
+
+                    return 'Not localised';
```
tractorcow, line 545: I tried to find something to critique here but it's already pretty good. :) Good job.

We probably should _t() some of these labels.

src/Extension/FluentVersionedExtension.php
```diff
@@ -325,6 +325,83 @@ public function existsInLocale($locale = null)
...
+        }
+
+        $versionSuffix = FluentVersionedExtension::SUFFIX_VERSIONS;
+        $baseClass = DataObject::getSchema()->baseDataClass($class);
+        $stageTable = DataObject::getSchema()->tableName($baseClass);
+        $liveTable = $stageTable . $versionSuffix;
+        $stagedTable = $record->getLocalisedTable($stageTable) . $versionSuffix;
+
+        $query = <<<SQL
+SELECT SLV.Version
```
tractorcow, line 375: Can you please ansi sql quote (double quote) identifiers for postgres compat?

tractorcow:
General comment; Maybe we should add some colour to the output too? (nice to have)

E.g. source -> different colour for source == locale, vs source == inherited?
Use badge colours for status? (orange, red, green)

This is amazing work.

src/Extension/FluentVersionedExtension.php
mfendeksilverstripe: This functionality is only relevant for versioned objects. You can still call this method on `RecordLocale` via `IsModified` on any object though ;-) I'll tidy up the naming so it's more clear.

src/Model/RecordLocale.php
mfendeksilverstripe: Good catch :)

mfendeksilverstripe:
Pushed up some UI changes to the badges (PR description updated as well).

src/Extension/FluentVersionedExtension.php
tractorcow: Fair enough, I accept that.

I will request a smaller change then; Can we use getRecordLocale() as the default (if $locale is not provided).

```php
$locale = $locale ?: $this->getRecordLocale();
```

That way my template support still works for looping over locales.

```html
<% loop $Locales %>
   <% if $Record.stagesDifferInLocale %>
    Modified in $Locale
  <% end_if %>
<% end_loop %>
```

tractorcow:
Just one comment on getRecordLocale usage. I think we are almost ready to merge;

Is there anyone else you want to get a review / approval on this before I get this in?

tractorcow:
Review on the two new commits are fine by the way.

chrispenny:
Love these changes! Thanks, team 😄

mfendeksilverstripe:
Pushed up PR fixes. I still plan to add one more fix - address the page edit from actions. Should come through later today.

tractorcow:
I might still have some feedback on it, but we could worry about that in a subsequent alpha.

mfendeksilverstripe:
Pushed up last batch fo changes.

* added an option to enable / disable localise actions (copy to draft, copy & publish), enabled by default
* Following actions now reflect the published state based on localised record instead of base record (publish action, information panel, unpublish action, archive action)
* current locale object helper method added

mfendeksilverstripe:
Pushed up Non-versioned UI improvements:

* save action is changed to `Localise` in case the record is not localised in current locale (edits only)
* delete action is changed to `Unlocalise` in case he record is localised and i's not the last locale

src/Extension/Traits/FluentAdminTrait.php
```diff
@@ -566,4 +577,80 @@ protected function inEveryStage($doSomething)
...
+        if ($record->hasExtension(Versioned::class)) {
+            return;
+        }
+
+        if ($record->existsInLocale()) {
+            // keep the action unchanged as the record is localised
+            return;
+        }
+
+        $saveAction = $actions->fieldByName('MajorActions.action_doSave');
```
tractorcow, line 601: This won't work on gridfield save action though will it? It'll only for for sitetree / cmsmain actions. Is it possible to change this search to match doSave action at any level?

tractorcow:
I wonder if "Unlocalise" action could use a tooltop to explain what it does to users who haven't seen it before? It requires a bit of explanation right?

mfendeksilverstripe:
Added a tooltip for Unlocalise action.

src/Extension/Traits/FluentAdminTrait.php
```diff
@@ -651,6 +651,16 @@ protected function updateDeleteAction(FieldList $actions, DataObject $record)
...
+            ->setAttribute(
+                'title',
+                _t(
+                    FluentExtension::class . '.UnlocaliseTooltip',
+                    'Remove {name} from current locale',
+                    [
+                        'name' => $record->i18n_singular_name()
+                    ]
+                )
+            );
```
tractorcow, line 664: Looks good to me.

tractorcow:
@mfendeksilverstripe I suggest you get another reviewer to approve this and we can merge and get out a new alpha.

tractorcow:
@chrispenny do you want to be the second reviewer, since I guess @mfendeksilverstripe wrote most of this?

chrispenny:
@tractorcow Ah sweet, yes I can absolutely do that. I'll put some time aside for Monday.

src/Extension/FluentExtension.php
```diff
@@ -845,6 +869,30 @@ public function LocaleInformation($locale = null)
         return RecordLocale::create($this->owner, $localeObj);
     }
 
+    /**
+     * Get list of locales where record is localised in
+     *
+     * @return array
+     */
+    public function LocaleInstances()
```
robbieaverill, line 877: The master branch of fluent requires PHP 7.1 already, would you like to add some return types?

src/Extension/FluentExtension.php
```diff
@@ -1085,12 +1158,22 @@ public function updateLocalisationTabConfig(GridFieldConfig $config)
...
+
         // Add each copy from / to
         foreach (Locale::getCached() as $locale) {
-            $config->addComponents([
-                new CopyLocaleAction($locale->Locale, true),
-                new CopyLocaleAction($locale->Locale, false),
-            ]);
+            if ($copyToLocaleEnabled) {
+                $config->addComponents([
+                    new CopyLocaleAction($locale->Locale, true),
```
robbieaverill, line 1168: I'd recommend we use Injector to instantiate these action classes. I wouldn't block merge with this but it's generally a good idea for extensibility.

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -202,6 +216,18 @@ public function updateCMSActions(FieldList $actions)
...
+        // Update specific sitetree localise actions
+        $this->updateLocaliseActions($actions);
+
+        // Update information panel (shows published state)
+        $this->updateInformationPanel($actions);
+
+        // Update the state of publish action (if needed)
+        $this->updatePublishState($actions);
+
+        // Updat unpublish and archive actions
```
robbieaverill, line 228:
```suggestion
        // Update unpublish and archive actions
```

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -366,6 +402,158 @@ protected function updateSavePublishActions(FieldList $actions)
...
+        if (!$publishAction) {
+            return;
+        }
+
+        // make sure that changes only on the base record
+        // do not trigger "need to publish" button state
+        // this is needed because the default interface looks
+        // at the base record instead of the localised page
+        $publishAction
+            ->setTitle(_t(SiteTree::class . '.BUTTONPUBLISHED', 'Published'))
```
robbieaverill, line 442:
```suggestion
            ->setTitle(_t('SilverStripe\\CMS\\Model\\SiteTree.BUTTONPUBLISHED', 'Published'))
```

`SiteTree::class` won't work in the text collector unfortunately

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -366,6 +402,158 @@ protected function updateSavePublishActions(FieldList $actions)
...
+            ->addExtraClass('btn-outline-primary font-icon-tick');
+
+        if (!$owner->stagesDifferInLocale()) {
+            return;
+        }
+
+        // If staged and live is different we change the button to "Publish"
+        // as the page hasn't been published
+        $publishAction
+            ->setTitle(_t(SiteTree::class . '.BUTTONSAVEPUBLISH', 'Publish'))
```
robbieaverill, line 455:
```suggestion
            ->setTitle(_t('SilverStripe\\CMS\\Model\\SiteTree.BUTTONSAVEPUBLISH', 'Publish'))
```

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -366,6 +402,158 @@ protected function updateSavePublishActions(FieldList $actions)
...
+        $moreOptions->removeByName('action_unpublish');
+
+        // update the label on archive action as it could have "unpublish and archive" which is incorrect
+        $archiveAction = $moreOptions->fieldByName('action_archive');
+
+        if (!$archiveAction) {
+            return;
+        }
+
+        $archiveAction->setTitle(_t(CMSMain::class . '.ARCHIVE', 'Archive'));
```
robbieaverill, line 488: Use string literal for FQCN here instead of `CMSMain::class`

src/Extension/FluentVersionedExtension.php
```diff
@@ -325,6 +325,76 @@ public function existsInLocale($locale = null)
...
+FROM "$stagedTable" as "VL"
+INNER JOIN "$liveTable" as "V"
+    ON "VL"."RecordID" = "V"."RecordID"
+    AND "VL"."Version" = "V"."Version"
+WHERE "VL"."RecordID" = ?
+AND "VL"."Locale" = ?
+AND "V"."WasPublished" = ?
+ORDER BY "VL"."Version" DESC
+LIMIT 1
+SQL;
```
robbieaverill, line 378: Is raw SQL like this compatible with subsites, or other modules that have low-level ORM manipulations?

src/Extension/FluentVersionedExtension.php
```diff
@@ -447,21 +517,53 @@ public static function prepoulateIdsInLocale($locale, $dataObjectClass, $populat
...
+                return '';
+            }
+        ];
+
+        $summaryColumns['Live'] = [
+            'title' => 'Live',
             'callback' => function (Locale $object) {
                 if ($object && $object->RecordLocale()) {
-                    return $object->RecordLocale()->IsPublished() ? 'Live' : '';
+                    return $object->RecordLocale()->IsPublished() ? 'Yes' : 'No';
```
robbieaverill, line 566: Do you want to localise "Yes" and "No", as well as maybe the title "Live" ?

src/Extension/Traits/FluentAdminTrait.php
```diff
@@ -566,4 +577,90 @@ protected function inEveryStage($doSomething)
...
+
+        $saveAction = $actions->fieldByName('MajorActions.action_doSave');
+
+        if (!$saveAction) {
+            return;
+        }
+
+        // update save action label to localise in case record is not localised yet
+        $saveAction
+            ->setTitle(_t(FluentExtension::class . '.Localise', 'Localise'))
```
robbieaverill, line 609: Use a string literal here for the FQCN. There's a few other places in this PR but I won't mention them any more.

src/Model/RecordLocale.php
```diff
@@ -315,4 +318,53 @@ public function IsDraft()
...
+
+        return $record->stagesDifferInLocale($this->getLocale());
+    }
+
+    /**
+     * Get the locale which is the source of content for this record
+     *
+     * @return Locale|null
+     */
+    public function SourceLocale()
```
robbieaverill, line 350: Is there a way you can name this something without a capital letter at the start? It will be a blocker for PSR-12 compliance in future

src/Extension/FluentExtension.php
```diff
@@ -1051,10 +1099,35 @@ protected function findRecordInLocale($locale, $table, $id)
...
+            'callback' => function (Locale $object) {
+                if ($object->RecordLocale()) {
+                    if ($object->RecordLocale()->IsDraft()) {
+                        return _t(self::class . '.LOCALISED', 'Localised');
+                    }
+
+                    return _t(self::class . '.NOTLOCALISED', 'Not localised');
+                }
+
+                return '';
```
chrispenny, line 1113: [nitpick] Switch to return early.

src/Extension/FluentExtension.php
```diff
@@ -1051,10 +1099,35 @@ protected function findRecordInLocale($locale, $table, $id)
...
+                    $sourceLocale = $object->RecordLocale()->SourceLocale();
+
+                    if ($sourceLocale) {
+                        return $sourceLocale->getLongTitle();
+                    }
+
+                    return _t(self::class . '.NOSOURCE', 'No source');
+                }
+
+                return '';
```
chrispenny, line 1130: [nitpick] Switch to return early.

src/Extension/FluentVersionedExtension.php
```diff
@@ -325,6 +325,76 @@ public function existsInLocale($locale = null)
...
+    /**
+     * Check if this record has modifications in this locale
+     * Fluent friendly version of @see Versioned::stagesDiffer()
+     *
+     * @param string|null $locale
+     * @return bool
+     */
+    public function stagesDifferInLocale($locale = null)
+    {
+        $record = $this->owner;
```
chrispenny, line 337:
```suggestion
        /** @var DataObject|Versioned|FluentExtension $record */
        $record = $this->owner;
```

src/Extension/FluentVersionedExtension.php
```diff
@@ -325,6 +325,76 @@ public function existsInLocale($locale = null)
...
+        if (!$record->hasExtension(Versioned::class)) {
+            return false;
+        }
+
+        // Need to check that it has stages and is not new
+        if (!$id || !$record->hasStages()) {
+            return false;
+        }
+
+        $locale = $locale ?: $this->getRecordLocale()->Locale;
```
chrispenny, line 351: Please check that `getRecordLocale()` is not `null` before attempting to access a property on it.

src/Extension/FluentVersionedExtension.php
chrispenny: Might be a good idea to give folks an extension point (or something) so that they can  manipulate this query.

src/Extension/FluentVersionedExtension.php
```diff
@@ -447,21 +517,53 @@ public static function prepoulateIdsInLocale($locale, $dataObjectClass, $populat
...
+                    }
+
+                    if ($recordLocale->IsDraft()) {
+                        return _t(self::class . '.DRAFT', 'Draft');
+                    }
+
+                    return _t(self::class . '.NOTLOCALISED', 'Not localised');
                 }
 
                 return '';
```
chrispenny, line 541: [nitpick] which to return early.

src/Extension/FluentVersionedExtension.php
```diff
@@ -447,21 +517,53 @@ public static function prepoulateIdsInLocale($locale, $dataObjectClass, $populat
...
+                    $sourceLocale = $object->RecordLocale()->SourceLocale();
+
+                    if ($sourceLocale) {
+                        return $sourceLocale->getLongTitle();
+                    }
+
+                    return _t(self::class . '.NOSOURCE', 'No source');
+                }
+
+                return '';
```
chrispenny, line 558: [nitpick] which to return early.

src/Extension/Traits/FluentAdminTrait.php
```diff
@@ -566,4 +577,90 @@ protected function inEveryStage($doSomething)
...
     }
+
+    /**
+     * Update save action based on localisation state
+     * valid states:
+     * - localised (localise label)
+     * - not localised (save label)
+     *
+     * @param FieldList $actions
+     * @param DataObject $record
```
chrispenny, line 588:
```suggestion
     * @param DataObject|FluentExtension $record
```

src/Extension/Traits/FluentAdminTrait.php
tractorcow: Yeah it's annoying but I am not sure if the text collector supports class constants yet.

tractorcow:
I agree with all the feedback given above; thanks @chrispenny  and @robbieaverill

src/Extension/FluentVersionedExtension.php
mfendeksilverstripe: This is the same as `Versioned::stagesDiffer()`, raw query, no extension points. I think that extension point is beyond scope of this PR and can be added later as needed.

src/Extension/Traits/FluentAdminTrait.php
robbieaverill: It supports `self::class` and `__CLASS__` at the moment

mfendeksilverstripe:
@tractorcow @robbieaverill @chrispenny Pushed up changes based on review.

src/Model/RecordLocale.php
```diff
@@ -351,14 +351,20 @@ public function getSourceLocale(): ?Locale
     {
         /** @var DataObject|FluentExtension $record */
         $record = $this->getOriginalRecord();
+        $validationMethod = 'existsInLocale';
 
-        if ($record->existsInLocale($this->getLocale())) {
+        if ($record->hasExtension(FluentVersionedExtension::class)) {
```
tractorcow, line 356: Maybe `existsInLocale` should check only the CURRENT stage, instead of both stages?

FluentVersionedExtension has this code

```php
    /**
     * Check if this record exists (in either state) in this locale
     *
     * @param string $locale
     * @return bool
     */
    public function existsInLocale($locale = null)
    {
        return $this->isDraftedInLocale($locale) || $this->isPublishedInLocale($locale);
    }
```

maybe it should be this?

```php
    /**
     * Check if this record exists (in either state) in this locale
     *
     * @param string $locale
     * @return bool
     */
    public function existsInLocale($locale = null)
    {
        if (Versioned::get_stage() === Versioned::LIVE) {
            return $this->isPublishedInLocale($locale);
       }
        return $this->isDraftedInLocale($locale);
    }
```

And then methods which CALL existsInLocale don't need to care about the versioning type.

tractorcow: We already have `isDraftedInLocale` so I don't see the problem with existsInLocale being stage aware. :)

tractorcow:
@mfendeksilverstripe when do you want to go and get a coffee and catch up? :P my email is <myfirstname> <dot> <mylastname> @gmail.com

mfendeksilverstripe:
@tractorcow I've sent you an email with our catch up follow up ;-)

src/Model/RecordLocale.php
mfendeksilverstripe: I like this idea :)

tractorcow:
I think we are about there.

once the tests are green let's just merge and tag this, if that's ok?

tractorcow:
https://github.com/tractorcow-farm/silverstripe-fluent/releases/tag/5.0.0-alpha10

Feel free to re-open a new PR with any fixes / enhancements to this feature.

PR #632, opened 2020-07-07
NEW: Localised copy

mfendeksilverstripe:
# Localised copy

This is a new feature which provides a way to duplicate related objects during the localisation process. This feature is intended for Localise by relation approach (AKA indirect localisation).

## Example

Suppose you have the following setup:

```
Page A extends Page

has_one -> ElementalArea

private static $field_include = [
    'ElementalAreaID',
];
```

```
Page B extends PageA

has_one -> Banner

private static $field_include = [
    'BannerID',
];
```

We have Page B created in `en` locale. We're using either `copy to draft` button or `copy to locale` action (UI shouldn't matter) to localise Page B in `jp` locale. Expected result is that Page B in `jp` locale will become exact copy of the one from `en` locale when it comes to data. For plain data (text fields) it's quite simple and this already works fine. For the relations though, our expectation is different. A deep copy of the related object needs to be created (via `duplicate()`) and the newly created copy needs to be assigned to the Page B in `jp` locale.

This feature is intended to facilitate this need.

## How to use this feature (overview)

* include localised copy configuration in your class
* alternatively, use provided extension points to run your custom duplication code

Example configuration:

```
/**
 * @var array
 */
private static $localised_copy = [
    'ElementalArea',
];

/**
 * @var array
 */
private static $localised_copy = [
    'Banner',
];
```

### Why Extension and not a Trait?

It's much easier to apply extension to module classes as well as update their static configuration. Trait would require subclassing.

### Can I include some other code inside makeLocalisedCopy() not directly related to object duplication?

Yes. For example you can add a Versioned Snapshot event dispatch and make a record about what locale is your object getting localised into. `makeLocalisedCopy()` is executed in the scope of your object so `$this` is available. Note that this method is executed `onBeforeWrite()` so there may be some limitations.

### Does this feature do anything on its own?

No, you have to set proper configuration to your class in order for this feature to do anything.


tractorcow:
I love the attention given to this solution; I wonder if we could keep the spirit of features such as `cascade_duplicates` and `owns` and offer a way to provide a yml/config only option as well?

E.g.

```php
private static $localised_copy = [
    'ElementalArea', // copy area
];
```

Or if you want to do deep nesting, use dot syntax.

```php
private static $localised_copy = [
    'ElementalArea.Elements', // copy area plus each element in this area
];
```

src/Extension/Traits/LocalisedCopyTrait.php
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Extension point in @see CopyLocaleAction::handleAction()
+     *
+     * @param string $fromLocale
+     * @param string $toLocale
+     */
+    public function onBeforeCopyLocale(string $fromLocale, string $toLocale): void
+    {
+        $this->setLocalisedCopyActive(true);
```
tractorcow, line 70: Instead of a temporary variable, could we not just call executeLocalisedCopy() directly?

src/Extension/Traits/LocalisedCopyTrait.php
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Provides a safe way to temporarily alter the global flag
+     * useful for unit tests
+     *
+     * @param callable $callback
+     * @return mixed
+     */
+    public function withLocalisedCopyState(callable $callback)
```
tractorcow, line 51: When is this used outside of tests? I am worried about test-only code that could be testing a use case that doesn't reflect day to day usage.

mfendeksilverstripe: This is not test-only code as there are other use cases for it. For example data migration from third party (page content is recreated during update). I'll update the docblock.

src/Extension/Traits/LocalisedCopyTrait.php
mfendeksilverstripe: The call needs to happens onBeforeWrite() because there are other cases which `onBeforeCopyLocale` doesn't cover (note that this is Localisation manager action). Cases not covered here:

* copy-to-draft button
* manual `write()` call

mfendeksilverstripe:
I like the idea of `$localised_copy` static configuration. I don't think we need the nested syntax though as the `duplicate()` will create a nested copy if `cascade_duplicates` is properly configured. I think wee need the callback as well to cover other cases (like even dispatch). I'll try to make it work in a similar way to default values (can be static config only but also a method).

src/Extension/Traits/LocalisedCopyTrait.php
```diff
@@ -0,0 +1,132 @@
...
+            // object has a base record and doesn't have a localised record and we are localising it
+            return true;
+        }
+
+        if ($this->existsInLocale() && $this->getLocalisedCopyActive()) {
+            // object has a localised record and the content is being overridden
+            // from another locale (via copy to/from)
+            // note that we can't rely on isChanged() because writeToStage() calls forceChange()
+            // which would make this condition true every time
+            return true;
```
tractorcow, line 120: @mfendeksilverstripe thanks for the succinct answers to my questions. I think this is a great change, but I feel a bit unsure about the necessity of the temporary property for flagging this functionality on/off per calls.

I wonder if we could apply a convention-over-configuration approach that simplifies the logic.

I am making the assumption that a "copy to locale" is the act of loading a record in one locale, and copying that locale's values to another.

I feel that we could possibly detect this (rather than flag this) event by inspecting query parameters.

getRecordLocale() will determine the locale that the record originated from; 

We could then use the following instead:
```php
$sourceLocale = $this->getRecordLocale();
$currentLocale = FluentState::singleton()->getLocale();
if ($this->existsInLocale() && $sourceLocale && $sourceLocale->Locale === $currentLocale) {
   // ...
```

tractorcow: Note: I am assuming that even if we use a manual ->write() call, you would want the localised copy to still trigger. :)

tractorcow: I might also suggest keeping the `withLocalisedCopyState` pattern, but instead of enabling this flag, you could use a similar option to globally disable the default behaviour (allowing manual migration to be performed if necessary).

E.g..

```php
$obj->withLocalisedCopyDisabled($callback);
```

tractorcow: If you roll this into a yml config, you won't need a trait, so I would assume the method would be a static method on FluentExtension.

mfendeksilverstripe:
Pushed up new changes:

* Trait approach changed to extensions approach
* extension points added to provide hook for custom duplication code

src/Extension/Traits/LocalisedCopyTrait.php
mfendeksilverstripe: Thanks for suggestions. Honestly, I don't like the global flag approach either so if we can get something else to do the job I would be happy to replace it. I'll test out your idea :)

mfendeksilverstripe:
Pushed up new changes:

* source locale detection instead of global flag

tractorcow:
Yay it's getting closer. :)

mfendeksilverstripe:
Pushed up new changes:

* global state for localised copy feature

tractorcow:
Please rebase and we can focus on finishing this PR.

mfendeksilverstripe:
@tractorcow It's rebased now :)

mfendeksilverstripe:
@tractorcow @chrispenny I'm happy with the feature, please review it.

chrispenny:
Sweet. I've put this at the top of my list for hackday (tomorrow).

src/Extension/FluentExtension.php
```diff
@@ -1176,4 +1197,110 @@ public function updateLocalisationTabConfig(GridFieldConfig $config)
...
+            if (!$original instanceof DataObject) {
+                continue;
+            }
+
+            if (!$original->exists()) {
+                continue;
+            }
+
+            $duplicate = $original->duplicate();
+            $owner->{$field} = $duplicate->ID;
```
tractorcow, line 1258: Be careful, polymorphic relations (has_one -> dataobject) have both a RelationID and RelationClass field.

I may suggest that instead of localised_copy declaring both the relation and the field, that you make it a flat list of fields. Then instead of using `{$field}` you can just use `setComponent($relation, $duplicate)` directly. dataobject will handle all the relation polymorphism for you.

src/Extension/FluentExtension.php
```diff
@@ -1176,4 +1197,110 @@ public function updateLocalisationTabConfig(GridFieldConfig $config)
...
+        }
+
+        $owner->invokeWithExtensions('onAfterLocalisedCopy');
```
tractorcow, line 1261: In case anyone needs it, it wouldn't hurt to add both $original and $duplicate as arguments (in case code needs to adjust / copy fields).

src/Extension/FluentExtension.php
mfendeksilverstripe: Good catch, will try it out :)

mfendeksilverstripe:
Pushed new changes:

* localised copy config is now a flat list
* extension points improvements

src/Extension/FluentExtension.php
mfendeksilverstripe: Works like a charm :D

src/Extension/FluentExtension.php
tractorcow: Oh right, I forgot it's in a loop. You sorted out a solution anyway nice :)

tractorcow:
Hope it's not too quick.

https://github.com/tractorcow-farm/silverstripe-fluent/releases/tag/5.0.0-alpha11

I can do another alpha if you need any more alts made. :)

mfendeksilverstripe:
Thanks @tractorcow, will ask @chrispenny for his review as well and if there are some fixes needed I'll raise another PR. If he's happy with this we will be able to package this into a beta :)

PR #634, opened 2020-07-16
v5: Rename cms publish required config

chrispenny:
No functional change, it's just always bugged me that I didn't really call this config the correct name. Given that we're moving to a major version, now seems like the time to update the name.

I'm not entirely sold on `cms_localisation_required` either... `backend_localisation_required`, maybe?

This config never required "publishing" for the object to be displayed in the CMS, it required "localisation".

I think the feature is still useful in certain situations. EG: For one of our projects we have a Model that is 100% managed through a 3rd party system (we just consume the data). There is no data fallback allowed for these Models, they must be localised in a Locale to be available. This config means that we can easily hide these Models in all interfaces (Model Admin, Dropdown fields, etc) in Locales where those Models are not available.

### DIFF
--- a/_config/model.yml
+++ b/_config/model.yml
@@ -3,7 +3,7 @@ Name: fluentmodel
 ---
 SilverStripe\ORM\DataObject:
   # See FluentExtension
-  cms_publish_required: false
+  cms_localisation_required: false
   frontend_publish_required: true
   # See FluentFilteredExtension
   apply_filtered_locales_to_stage: true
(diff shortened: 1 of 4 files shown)


chrispenny:
@mfendeksilverstripe @tractorcow thoughts on this one before we go to `beta`? I don't want to change the behaviour of the config - just the name.

mfendeksilverstripe:
I think `localisation_required` is pretty descriptive and in-line with the recent terminology changes.

tractorcow:
I guess I don't mind changing the name. Just maybe add an upgrading notice or deprecation check on the old config.

chrispenny:
Wasn't sure if we should just add BC for now or not. I opted to just add it.

PR #639, opened 2020-07-24
Localised versioned history

mfendeksilverstripe:
# Localised versioned history

This is intended to go into `5.1` as these changes are high impact and possibly need more improvements to cover the intended scope.

## Details

* archive flag respects localised state - either it's archived in current locale or other locale
* restore action is only shown if there is archived content available in current locale
* archive view is only available for objects which have something in the localised archive
* Versioned history versions list localised - each locale shows only versions for the current locale
* Version history preview localised - viewing past versions in the history viewer shows correct data

### Archive behaviour

* archive view is available only for objects which have some archived content in the current locale
* user is expected to switch to the locale where the content was archived to do the restore action

### Merged

https://github.com/silverstripe/silverstripe-versioned/pull/287
https://github.com/silverstripe/silverstripe-versioned/pull/297

### Released

Waiting for the silverstripe/framework 4.7

### Locale

* NZ - global default
* US - fallbacks on NZ
* JP - no fallback

### Basic test setup

* create a new page in NZ
* copy to US locale
* copy to JP locale
* edit page in US and assign US Title
* edit page in JP and assign JP Title

### History view test

* Basic test setup steps
* view history in each locale and each version and check that the versions sets are distinct for each locale
* version preview works as intended (shows data only form current locale)

### Archive view test

* Basic test setup steps
* archive page in US
* archive page in JP
* switch back to US and `restore to draft`, you should get back the US version
* switch back to JP and `restore to draft`, you should get back the JP version

### Versioned methods

Versioned methods are now fully localised. For example `isPublished()` will look at localised record, not base record. This has a big impact on the CMS UI as lot of UI is built based on the versioned state.

### Archive view

CMS is able to show a page even though there is no content for the page - it will show the data from base record. If we localise version lookups this approach doesn't work for the archive view. If there wasn't any content for the page in specific locale the archive view will show page as not found. This might be consistent from technical point of view but it may be user unfriendly. It would be much better to show a read only page from other locale and show a label saying that this may has been archived in some other locale.

This example shows that consistency might need to be overridden with user friendly-ness ins some cases.

 ### Conclusion

Maybe we need to review this feature from the requirements point of view and check if we can achieve it technically.

The biggest decision is this one:

`isPublished()` localised lookup vs `isPublished()` base record & `isPublishedInLocale()` for localised record


mfendeksilverstripe:
@tractorcow @chrispenny This is now ready for review.

src/Extension/FluentVersionedExtension.php
```diff
@@ -794,36 +815,87 @@ public function updateGetVersionNumberByStage(&$version, $class, $stage, $id, $c
...
             $localisedTable,
-            Convert::raw2sql($locale, true),
             $versionedTable,
-            $liveSegment
+            $liveSegment,
+            $idSegment
         );
 
-        $localisedVersion = DB::prepared_query($query, [$id])->value();
+        $params = [$locale, $locale];
```
tractorcow, line 885: Ah thanks for doing that. :D

tractorcow:
Is there any chance of a test case to show the history is queried correctly?

mfendeksilverstripe:
@tractorcow Yes, I plan to add unit tests to cover the major cases :)

src/Extension/FluentVersionedExtension.php
```diff
@@ -595,4 +634,331 @@ public function updateLocalisationTabConfig(GridFieldConfig $config)
...
+     */
+    public function updateIsArchived(bool &$isArchived): void
+    {
+        $locale = FluentState::singleton()->getLocale();
+
+        if (!$locale) {
+            return;
+        }
+
+        $isArchived = $this->owner->isArchivedInLocale();
```
chrispenny, line 652: I think this should pass the `$locale` var that you've defined.

src/Extension/FluentVersionedExtension.php
```diff
@@ -595,4 +634,331 @@ public function updateLocalisationTabConfig(GridFieldConfig $config)
...
+    /**
+     * @param string|null $locale
+     * @return bool
+     */
+    public function isArchivedInLocale(?string $locale = null): bool
+    {
+        $locale = $locale ?: FluentState::singleton()->getLocale();
+
+        if (!$locale) {
+            return false;
```
chrispenny, line 664: I'm wondering if perhaps this should return `null`, rather than `false`?

`false` indicates that something is specifically **not** archived in a locale, but if there is no locale, then I think neither `true` or `false` is technically accurate.

This should also mean that `updateIsArchived()` does not update the flag at all - whatever was determined previously is left as is.

src/Extension/FluentVersionedExtension.php
```diff
@@ -595,4 +634,331 @@ public function updateLocalisationTabConfig(GridFieldConfig $config)
...
+        }
+
+        $owner = $this->owner;
+        $class = $owner->ClassName;
+        $id = $owner->ID ?: $owner->OldID;
+
+        if (!$id) {
+            return false;
```
chrispenny, line 672: Should this also be `null`, rather than `false`?

src/Extension/FluentVersionedExtension.php
```diff
@@ -595,4 +634,331 @@ public function updateLocalisationTabConfig(GridFieldConfig $config)
...
+        }
+
+        if ($owner->existsInLocale($locale)) {
+            return false;
```
chrispenny, line 676: Should this also be `null`, rather than `false`?

src/Extension/FluentVersionedExtension.php
```diff
@@ -595,4 +634,331 @@ public function updateLocalisationTabConfig(GridFieldConfig $config)
...
+     *
+     * @param string|null $locale
+     * @return bool
+     */
+    public function hasArchiveInLocale(string $locale = null): bool
+    {
+        $locale = $locale ?: FluentState::singleton()->getLocale();
+
+        if (!$locale) {
+            return false;
```
chrispenny, line 693: Should this also be `null`, rather than `false`?

src/Extension/FluentVersionedExtension.php
mfendeksilverstripe: I think this should be `false` as no `ID` indicates a new object which doesn't have an archive.

src/Extension/FluentVersionedExtension.php
mfendeksilverstripe: This should be `false` as a record which exists in locale is considered NOT archived.

mfendeksilverstripe:
@chrispenny Pushed up fixes but also one notable change - when a page has no content in locale (draft, live or archived) you can't view it anymore in that locale. The reason for this is that versions lookup is now full localised and it's really hard to make exceptions and not break other functionality.

tractorcow:
I would agree with that last statement; It's an archive so probably shouldn't show something that was never localised in that locale, right?

Good feedback too @chrispenny thanks :)

mfendeksilverstripe:
@tractorcow @chrispenny Unit tests added. Note that these test will break without the dependency (Versioned module).

clarkepaul:
This looks like an interesting set of changes, would be good to see them in action just to be on top of what the new experience is like. Maybe a mini demo for the UX team makes sense if the Product dev team review and merge.

mfendeksilverstripe:
@clarkepaul Sounds good, please book some time in the calendar and I'm happy to prepare a small demo ;-)

tractorcow:
@mfendeksilverstripe 5.0.0-rc1 is out, and I've created a 5 branch (alias 5.1) if you want to merge it there.

mfendeksilverstripe:
Thanks @tractorcow I'll pick it up 👍

mfendeksilverstripe:
Found some issues. Fixed one but there are still a couple more left.

mfendeksilverstripe:
@tractorcow @chrispenny I pushed up [new changes](https://github.com/tractorcow-farm/silverstripe-fluent/pull/639/commits/1b94fb4349f5f15d664d77a1eb317e89fd071e8e):

* added missing in memory cache flush
* updated behaviour of `isOnDraft()` and `isPublished()`
* the change above has a [new dependency](https://github.com/silverstripe/silverstripe-versioned/pull/297) which has to be merged before the unit tests can pass (but they are passing on my local)

The cache flush change is pretty straightforward but the change regarding the `isOnDraft()` and `isPublished()` requires some more discussion I think.

With the changes on this branch the version lookups done within `isOnDraft()` and `isPublished()` would be localised. I think that this is expected and correct behaviour but this is breaking too many things in the core modules. For example, the default CMS actions would not be created in some cases. Also, it seems we already have a paradigm of not overriding existing functions but rather adding locale specific functions for example `isPublished` vs `isPublishedInLocale()`.

So the new behaviour is:

`isOnDraft` - looks at the base record, you should use `isDraftedInLocale` for localised record
`isPublished` - - looks at the base record, you should use `isPublishedInLocale` for localised record

Which is the same as the old behaviour but the reason for this before was a bug (or missing feature) and now it's intentional (expected).

Pleas let me know what you think.

chrispenny:
@tractorcow @mfendeksilverstripe 

Personally, I'm struggling with this decision.

From a program point of view, I have how I expect these methods to be used, but then it seems like the reality of how they're actually used, is (in some instances) different.

My initial reaction though was that I think `isPublished`/`isOnDraft`/etc should be localised, and that we should set that expectation now (while we're already in a major version change) and move forward from there.

I had a quick chat with Mojmir about some of the issues (eg: like fields no longer being scaffolded), and some of them I think we've already mitigated (eg: through other interfaces, like the localisation management tool). There are other areas though (like viewing an archived page that only had inherited content), that might need additional improvements. Arguably... I still think that those areas are correct from data point of view, and I kinda feel like it's acceptable for these areas to be a "WIP".

On the other hand... we could be opening up a can of worms if we don't understand the extent of the improvements/fixes that would then be required.

mfendeksilverstripe:
@tractorcow @chrispenny Added some discussion points from the recent meeting.

mfendeksilverstripe:
@chrispenny @tractorcow Pushed up new changes:

* Versioned lookups are now fully localised
* Removed code which was patching up broken UI as it's no longer needed - the UI is built correctly now as the versioned state is propagated correctly (for example publish button state is now correct out of the box as it's using localised data lookup)
* Removed an option to have "Copy" buttons as these buttons won't get scaffolded - use locale manager tab instead


With this changeset I'm happy with the overall state of the Fluent and would be keen to line this up for release as `5.1`

chrispenny:
I talked through the latest commits with Mojmir IRL, and I'm happy with them 🥰 I think it makes everything much more predictable moving forward.

tractorcow:
Sorry I have not given this PR a lot of my time yet; @chrispenny @mfendeksilverstripe have your team got this to the point where it's reasonably unlikely to be substantially changed? I will give it a code quality review if you are happy with the functionality on your side.

mfendeksilverstripe:
@tractorcow Yes, we are happy with the current state and are unlikely to make any more changes. Please go ahead with your review 👍

mfendeksilverstripe:
@tractorcow @chrispenny If there is no further feedback on this one I'll merge it after the Christmas break. With Framework `4.7` out we have all needed dependencies. I would be keen to start testing this on a small but real world project.

tractorcow:
nice job!

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -43,15 +42,6 @@ class FluentSiteTreeExtension extends FluentVersionedExtension
...
     private static $locale_published_status_message = true;
 
-    /**
-     * Enable localise actions (copy to draft and copy & publish actions)
-     * these actions can be used to localise page content directly via main page actions
-     *
-     * @config
-     * @var bool
-     */
-    private static $localise_actions_enabled = true;
```
tractorcow, line 53: Oops, did this config get deleted? This has become a breaking change if so.

tractorcow:
@mfendeksilverstripe sorry I haven't really given this much of a review, but I have some concern that this merge removed a bunch of features (config, buttons, methods) that might make this an actual breaking change. Can you please give this a "post-mortem" review to confirm that nothing is broken, and justify the removed API to confirm none of these are going to break upgrades from 5.0?

src/Extension/FluentSiteTreeExtension.php
tractorcow: Also, I'm not 100% sure what this config is doing. It's possible I just need to be educated and nothing at all is wrong. :)

tractorcow:
@mfendeksilverstripe this looks like there might have just been a bad rebase; I will reserve judgement until I hear back from your team though.

Thanks again for your hard work on this feature!

mfendeksilverstripe:
Hey @tractorcow , this change-set fixes Version history lookup so as a consequence common version related methods such as

* `isOnDraft()`
* `isPublished()`
* `isArchived()`
* `stagesDiffer()`

Are now localised. As a consequence there are no buttons to augment as the core functionality which builds the form no longer produces those buttons if the page is not localised (drafted in locale) / published in locale.

This was discussed as a part of this PR and is expected behaviour. The intended user experience is to rely on locale manager tab instead of form buttons (copy to draft, copy & publish). The config was removed as there are no buttons to augment.

If we want to retain those buttons we would need to re-create them from scratch instead of augmenting them. This shouldn't be too much effort but there are other reasons why these buttons should not be used as they can produce unexpected behaviour in some cases (for example locales which do not inherit from other locales would get content from base record which is undesirable).

Please let me know if you want me to create some sort of compatibility patch PR.

tractorcow:
I think that's ok, I just wanted to make sure it was intentional. :) thanks for the explanation.

tractorcow:
hi @mfendeksilverstripe this PR forces us to upgrade to silverstripe 4.7. Is there any way to avoid this?

tractorcow:
@mfendeksilverstripe how do you feel about this:
 - current 5 branch (aliased as 5.1) - becomes fluent 6
 - branch current 5.0 branch and make 5 (new 5.1 alias)
 - keep concurrent releases of fluent 5 + 6. fluent 5 remains for older framework version, but fluent 6 requires 4.7

I feel like a forced upgrade (especially with new major version of graphql) is a bit too heavy.

mfendeksilverstripe:
Hey @tractorcow no, the `4.7` upgrade is mandatory as it contains essential fixes to Versioned (which is split across two core modules). I agree that the change-set in `5.1` might be too heavy and it makes sense to change it to `6`. Happy to go with your suggestion 👍 .

PR #653, opened 2020-10-01
Translation to Slovenian (sl.yml)

tomaszpirc:
New translation to Slovenian (translation is also available in https://www.transifex.com/silverstripe/silverstripe-fluent/).

### DIFF
--- a/lang/sl.yml
+++ b/lang/sl.yml
@@ -1,0 +1,58 @@
+sl:
+  TractorCow\Fluent\Control\LocaleAdmin:
+    MENUTITLE: Lokalizacije / Jeziki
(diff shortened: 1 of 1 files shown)


robbieaverill:
Thanks! We can merge this - it will be overwritten eventually by Transifex's version, but that's OK.

tomaszpirc:
Great. I did a PR because I was eagerly waiting for 5.0 to be released just to find out that transifex's lang files were not included. :)

tractorcow:
I haven't updated transifex in ages lol. I should do that...

When we do, we'll just be careful to do a pull, merge, push.

PR #657, opened 2020-10-14
Create es.yaml

estefania-vila-simplon:
translated into spanish


robbieaverill (commented):
Thanks for contributing. We manage non-English translations in Transifex. Would you mind submitting the translations there? https://www.transifex.com/silverstripe/silverstripe-fluent/language/es/

PR #659, opened 2020-11-11
ENHANCEMENT 404 errors when changing locales will redirect to the home page of the CMS

tractorcow:
E.g. when viewing a record (that may only be valid in one locale), if you change the locale dropdown, you could sometimes end up on a 404 page.

`That record was not found`

This middleware will detect those errors and safely redirect the user to the main admin page instead.

I've opted not to do this on the frontend because it could interfere with any custom home page routing the dev could have built.


tractorcow:
Merging up branches to get tests working

tractorcow:
Note: Im going to target this feature for 5.1, but don't want to merge until we sort out https://github.com/tractorcow-farm/silverstripe-fluent/issues/681

PR #671, opened 2021-01-14
MNT Use travis shared config, use sminnee/phpunit

emteknetnz:
Targets 5.0 branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,40 +1,4 @@
-language: php
+version: ~> 1.0
 
-dist: xenial
-
-services:
-  - mysql
-
-addons:
-  apt:
-    packages:
-      - tidy
-
-matrix:
-  include:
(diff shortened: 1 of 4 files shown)


tractorcow:
Can we justify why we have forked phpunit? I'm not aware of the background around that decision.

emteknetnz:
sminnee/phpunit is a fork of phpunit 5.7 that includes php8 compatibility

Idealy we'd be running phpunit ^9 though we haven't yet done the work to get silverstripe compatible with that

tractorcow:
sounds fair, I'll accept that. Hopefully we can upgrade phpunit and not need this. :)

dnsl48 (approved):
looks good to me

dnsl48:
Assuming the broken tests are caused by the increased code coverage and not this PR.

PR #684, opened 2021-02-26
NEW: Clickable links on page info for localisation tab.

aria5305:
# Clickable links on page info for localisation tab

Augment Localisation tab with clickable locale links to allow easy navigation between page localisations.

This feature is applied to only page via `FluentSiteTreeExtension`.

## Preview

Screen Shot 2021-02-26 at 3 05 02 PM

### DIFF
--- a/src/Extension/FluentSiteTreeExtension.php
+++ b/src/Extension/FluentSiteTreeExtension.php
@@ -622,7 +622,8 @@ class FluentSiteTreeExtension extends FluentVersionedExtension
                 $recordLocale = $object->RecordLocale();
                 $locale = $recordLocale->getLocale();
                 $localeLink = Controller::join_links($url, '?l=' . $locale);
-                $localeTitle = $recordLocale->getTitle();
+                $localeTitle = Convert::raw2xml($recordLocale->getTitle());
+
                 $render = sprintf('<a href="%s" target="_top">%s</a>', $localeLink, $localeTitle);
 
                 return DBField::create_field('HTMLVarchar', $render);
--- a/src/Extension/FluentSiteTreeExtension.php
+++ b/src/Extension/FluentSiteTreeExtension.php
@@ -581,6 +581,8 @@ class FluentSiteTreeExtension extends FluentVersionedExtension
     }
 
     /**
+     * Augment Localisation tab with clickable locale links to allow easy navigation between page localisations
+     *
      * @param $summaryColumns
      * @see FluentExtension::updateFluentCMSFields()
      */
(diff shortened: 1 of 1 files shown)


tractorcow:
So cool feature.

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -578,4 +579,54 @@ public function actionComplete($form, $message)
...
+        $summaryColumns['Title'] = [
+            'title' => 'Title',
+            'callback' => function (Locale $object) use ($url) {
+                if (!$object->RecordLocale()) {
+                    return;
+                }
+
+                $recordLocale = $object->RecordLocale();
+                $locale = $recordLocale->getLocale();
+                $localeLink = Controller::join_links($url, '?l=' . $locale);
```
tractorcow, line 624: I know this is a super pedantic, but we should probably use http_build_query to make param injection impossible :) I don't trust any unescaped querystring args.

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -578,4 +579,54 @@ public function actionComplete($form, $message)
...
+                $localeTitle = $recordLocale->getTitle();
+                $render = sprintf('<a href="%s" target="_top">%s</a>', $localeLink, $localeTitle);
```
tractorcow, line 626: HTML title is not escaped here either.

tractorcow:
I found a few tiny security risks; Happy to merge once those are resolved.

src/Extension/FluentSiteTreeExtension.php
aria5305: Hey Damian, thanks for the feedback :) 
It seems like that the `join_links` method already uses `http_build_query` in it
See https://github.com/silverstripe/silverstripe-framework/blob/4.7/src/Control/Controller.php#L702

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -578,4 +579,55 @@ public function actionComplete($form, $message)
...
+            'title' => 'Title',
+            'callback' => function (Locale $object) use ($url) {
+                if (!$object->RecordLocale()) {
+                    return;
+                }
+
+                $recordLocale = $object->RecordLocale();
+                $locale = $recordLocale->getLocale();
+                $localeLink = Controller::join_links($url, '?l=' . $locale);
+                $localeTitle = Convert::raw2xml($recordLocale->getTitle());
```
mfendeksilverstripe, line 625: @aria5305 Please import namespace for `Convert`

src/Extension/FluentSiteTreeExtension.php
tractorcow: Yes, but those strings are passed through parse_str first; That is the method you need to escape FOR. :) If you don't escape your `&` properly then parse_str will escape those variables incorrectly.

src/Extension/FluentSiteTreeExtension.php
tractorcow: You can use `Convert::raw2xml` to escape it by the way :)

aria5305:
Hi @tractorcow I've pushed some changes re the PR feedback for review again :)

tractorcow (approved):

tractorcow:
Thanks @aria5305 good job.

PR #685, opened 2021-03-01
BUG: Fix Travis setup to run with correct requirements.

mfendeksilverstripe:
## BUG: Fix Travis setup to run with correct requirements

* Fix Travis setup to run with correct requirements to match composer requirements
* Fix breaking unit tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
 version: ~> 1.0
 
+env:
+  global:
+    - COMPOSER_ROOT_VERSION="4.7.x-dev"
+    - REQUIRE_RECIPE="4.7.x-dev"
+
 import:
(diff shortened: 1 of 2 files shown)


mfendeksilverstripe:
The travis config is now updated but it looks like we have some `PGSQL` issues.

tractorcow:
Yeah there is code in there that's generating non-pgsql SQL :)

ansi-sql identifiers should always be double quoted. MySQL just doesn't complain about it, but it's still incorrect.

tractorcow:
See FluentVersionedExtension::hasArchiveInLocale(). It's using unquoted tables.

mfendeksilverstripe:
@tractorcow Tests are all green, please review.

chrispenny (commented):
The change seems fine to me (if these areas are covered by unit tests).

PR #688, opened 2021-03-15
Include action in locale links

### LINKED ISSUE #458: Action is NOT appended to the page Locale URL
Page action is not appended to the localization link, it only shows the page link with no actions.
```
class MyPage extends Page {
}

class MyPageController extends PageController {
    private static $allowed_actions = ['show'];
    private static $url_handlers = ['show/$ID' => 'show'];

    public function show() {
         ....
    }
}
```

Template:
```
<select class="select" onchange="window.location.href=this.value;">
    <% loop Locales %>
        <% if LinkingMode = current %>
            <option selected value="$Link.ATT">$Title.XML</option>
        <% else %>
            <option value="$Link.ATT">$Title.XML</option>
        <% end_if %>
    <% end_loop %>
</select>
```

View Source: www.example.com/en/mypage/show/5
```
<select class="select" onchange="window.location.href=this.value;">
     <option value="www.example.com/en/mypage/" selected="">English</option>
     <option value="www.example.com/ar/mypage/">العربية</option>
</select>
```
`show/5` Should be appended to the end of the locale link
```
<select class="select" onchange="window.location.href=this.value;">
     <option value="www.example.com/en/mypage/show/5" selected="">English</option>
     <option value="www.example.com/ar/mypage/show/5">العربية</option>
</select>
```

purplespider:
Fixes #458

### DIFF
--- a/src/Model/RecordLocale.php
+++ b/src/Model/RecordLocale.php
@@ -209,7 +209,8 @@ class RecordLocale extends ViewableData
         }
 
         // Get link from localised record
-        return $record->Link();
+        $action = Controller::curr()->getRequest()->params()['Action'] ?? null;
+        return $record->Link($action);
     }
 
     /**


src/Model/RecordLocale.php
```diff
@@ -209,7 +209,8 @@ public function getLink()
         }
 
         // Get link from localised record
-        return $record->Link();
+        $action = Controller::curr()->getRequest()->params()['Action'] ?? null;
```
tractorcow, line 212: The problem here is that you are making a model action dependant on controller state, which creates a risky and inverse dependency between model / controller.

Better solution: Allow getLink to take an action (like nearly every other link / relative link method) and conditionally join it.

E.g. 

```
public function getLink($action = null) {
}
```

and remove Controller::curr() in it's entirety. :)

Then you can use `$Link('show')` in your custom template.

tractorcow: You could also support this with `$record->LocaleLink($this->getLocale(), $action)`

tractorcow: The main reason I have a problem with this code is that Controller::curr() may not necessarily be the controller for the record in question. E.g. Controller::curr() being `Security`, but $record being a `HomePage` class.

It's risky to make this presumption.

tractorcow: Better solution:

- add $action to getLink() method
- add getCurrentLink() method, that assumes the current controller action, and calls getLink() with the current action / id / etc...

PR #691, opened 2021-03-23
BUG Fix issue with Security link being wrong

### LINKED ISSUE #689: Frontend locale menu doesn't work on login form (Security)
The locale menu (`LocaleMenu.ss`) doesn't appear to work when on the login form or password reset form (e.g. `Security`).

PR #688 fixes the issue with actions not being included in links, but there is a 2nd issue that results in the links for all Locales within the locale menu pointing to the "current" locale, instead of to the other relevant locale.

I believe the issue is `getRecordLocale()`([Link](https://github.com/tractorcow-farm/silverstripe-fluent/blob/5.0/src/Extension/FluentExtension.php#L813)) is failing to obtain the relevant local when on `Security` and is falling back to `getCurrentLocale()`.

But I can't get my head around how it could be resolved.

_Fluent v5.0.0_

tractorcow:
fixes #689

@purplespider  can you please help test this with your project? It should resolve the security url issue.

### DIFF
--- a/src/Model/RecordLocale.php
+++ b/src/Model/RecordLocale.php
@@ -80,15 +80,18 @@ class RecordLocale extends ViewableData
             return $this->record ?: null;
         }
 
-        // If record isn't saved, use draft record
-        if (!$this->originalRecord->isInDB()) {
-            return $this->originalRecord;
+        // If record isn't saved, clone the draft record, but mark it
+        // as belonging to the current locale
+        $originalRecord = $this->getOriginalRecord();
+        if (!$originalRecord->isInDB()) {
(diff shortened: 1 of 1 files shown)


purplespider:
Ace! Can confirm this fixes #689 in my project.

However, just to confirm, `LocaleMenu.ss` links still won't work on Security pages until #458 is also resolved.

tractorcow:
hm yes I think you are right.

tractorcow:
I have re-reviewed https://github.com/tractorcow-farm/silverstripe-fluent/pull/688 to add some suggestion for getting that fix over the line.

PR #693, opened 2021-04-16
Bugfix for version history viewer

aria5305:
When you navigate to version history viewer you get an exception from the MySQL query as `$localisedVersionTable` is a string that looks like `"SiteTree_Localised_Versions"` which means it currently then becomes `""SiteTree_Localised_Versions""`


tractorcow:
Perfect and good job; The phpdoc of course says "unquoted table name" :) Not sure why I didn't follow the rules.

Issue #702, opened 2021-04-23
feature request: only write default locale field changes to the base table(s)


pine3ree:
Would it be possible to avoid saving non default locale data to the base table?
e.g If I change and save the Page Title for locale `de_DE`, it would be nice to keep the original `en_US` Title in the base `SiteTree(_Live)` table. But if I am changing the Title in the default locale (`en_US`), I would that to be changed in the base table too.  In short: writing to the base table(s) should be done only when the fields in default locale are changed.
kind regards

PR #703, opened 2021-04-24
update manipulation to skip base table translate fields

pine3ree:
...if editing in non-default-locale.

ref: https://github.com/tractorcow-farm/silverstripe-fluent/issues/702

p.s.
A build failure happens for php-7.2 (PGSQL), but it seems unrelated since it happens for the original branch too.
original 5.0 branch failure: https://travis-ci.com/github/pine3ree/silverstripe-fluent/builds/224071758

kind regards

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -57,7 +57,6 @@ localisations, or being a part of the contributing team, please see
  * [Scenarios](docs/en/scenarios.md)
  * [Localised copy](docs/en/localised-copy.md)
  * [Migrating from Translatable](docs/en/migrating-from-translatable.md)
- * [Changelogs](CHANGELOG.md)
 
 ## Translations
 
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -12,6 +12,7 @@ use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\FormField;
 use SilverStripe\Forms\GridField\GridField_ActionMenuItem;
 use SilverStripe\Forms\GridField\GridFieldConfig;
+use SilverStripe\i18n\i18n;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataExtension;
 use SilverStripe\ORM\DataObject;
@@ -593,7 +594,8 @@ class FluentExtension extends DataExtension
         }
 
         $defaultLocaleObj = Locale::getDefault();
-        $defaultLocale = $defaultLocaleObj ? $defaultLocaleObj->getLocale() : i18n::config()->get('default_locale');
+        $defaultLocale    = $defaultLocaleObj ? $defaultLocaleObj->getLocale() : i18n::config()->get('default_locale');
+        $notDefaultLocale = $locale->getLocale() !== $defaultLocale;
 
         // Get all tables to translate fields for, and their respective field names
         $includedTables = $this->getLocalisedTables();
@@ -607,9 +609,9 @@ class FluentExtension extends DataExtension
                 $locale
             );
 
-            // Remove localised fields from base table if the editing locale is
+            // Remove localised fields from base tables if the editing locale is
             // not the global default locale
-            if ($locale->getLocale() !== $defaultLocale) {
+            if ($notDefaultLocale) {
                 foreach ($localisedFields as $localisedField) {
                     unset($manipulation[$table]['fields'][$localisedField]);
                 }
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -592,6 +592,9 @@ class FluentExtension extends DataExtension
             return;
         }
 
+        $defaultLocaleObj = Locale::getDefault();
+        $defaultLocale = $defaultLocaleObj ? $defaultLocaleObj->getLocale() : i18n::config()->get('default_locale');
+
         // Get all tables to translate fields for, and their respective field names
         $includedTables = $this->getLocalisedTables();
         foreach ($includedTables as $table => $localisedFields) {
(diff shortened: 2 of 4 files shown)


pine3ree:
...actually, this only works for unversioned dataobjects...

tractorcow:
I'm not sure if I 100% agree with this change. Part of the update to fluent 4 is to deprecate the idea of needing a default language. For instance, not all records need to ever exist in the "global default". Should those just be blank if created in any other locale?

pine3ree:
Hello @tractorcow,
thanks for looking into it. :-)

The main opinionated reason to only save the default locale data, if any, into the main table is that if we later decide to keep only one language it will most probably be the default lnguage (it happened to me in the past with a few clients, they wanted to start a multilang website with 3~4 language and after its done they only kept the default).

A better approach would be to complete remove data duplication (even base table columns) and keep the localised data only in the localised tables, but I am not sure this is possible with silverstripe without explicitely removing those fields from `$db` config prop and adding extension methods for localises fields in a ways similar to the one used for relations. 

Years ago when I was using Translatable, I did that for unversioned dataobjects:
1 DataObject without localised data + 1 related DataObject for localised fields (basically what I always do when using/building orms) and virtual props/fields that read from the translation object fields for a given locale. Data-objects and related translation for the curent locale were loaded and hydrated with a join query. (Localised form fields in tabs were also time-saving when editing records...)
In our case keeping empty db-columns in the base table for localised would still be an improvement (IMO) in terms of avoiding duplication.

So basically In my own opinionated order of preference I would like any of the following:
1. No localised fields columns in the base-table
2. Empty columns for localised fields in the base-table
3. Columns for localised fields in the base-table containing the defaul language data, if any

kind regards,
P.S. Feel free to close this PR at any time....

tractorcow:
Hey @pine3ree thanks for the reply.

I think the important thing to keep in mind is that the main way that fluent differs from modules like translatable is that we still have a single sitetree, where translatable has entirely different trees for each locale.

If a page is localised into the global default, it should not be affected by modifications to the base record, since the localised row will always be used.

An edge case that we need to consider is, if we create a page in a non-default locale, and then switch to any other locale, what text should we see is the sitetree view? If we don't write to the base record, that view will literally have a node without any text at all. If you click it and enter edit mode, the entire form would also be blank. It's not a very good user experience.

As I mentioned before, i don't really want to have the concept of a "default locale" affecting the CMS any more than it absolutely needs to. The only reason there is still a default at all, is because we need a fallback for when locale detection fails, or you enter the main home page without any url path prefix.

pine3ree:
Hello @tractorcow,

> Hey @pine3ree thanks for the reply.
> 
> I think the important thing to keep in mind is that the main way that fluent differs from modules like translatable is that we still have a single sitetree, where translatable has entirely different trees for each locale.

Yes, it always felt more natural to me having a single tree + locale tables for translated fields. The translatable module is one of the few exception I have ever found in my career as a developer. :-)

> An edge case that we need to consider is, if we create a page in a non-default locale, and then switch to any other locale, what text should we see is the sitetree view? If we don't write to the base record, that view will literally have a node without any text at all. If you click it and enter edit mode, the entire form would also be blank. It's not a very good user experience.

If the default text in a new locale is taken from the base table fields then this excludes the possibility of not having (or keeping blank) localised fields int  base table. But in this case I would prefer to start a new translation with the language I consider as default or my natural language, usually the first I write a page in....but since one of the goal of this module is to avoid the concept of a default language I guess this doesn't make sense as well in this context.
 
Thanks anyway for looking into this matter.
kind regards,
maks

tractorcow:
If we had to have everything perfectly, what we could institute is ensuring that the default row is always written at least once, but then 'project' that row so that it isn't modified except if the current locale is "global default" though. That is more complex though, but just to let you know what could possibly be in the future. :)

tomaszpirc:
Sorry to comment a bit late. We had to resolve this issue with using the default language for the base table because our editors had issues when changing to their language and getting a bunch of pages in the site tree in different languages as placeholders (until the first localization is done). Especially difficult was when last save for a particular page happened for a more exotic language and editors couldn't figure out what non-translated page was about. 
I discussed this with the team and though we understand that default doesn't mean necessarily primary language, we used this info rather than introduce a new flag for the language. So in the end we just write to the base table only when change is made for the default language.

pine3ree:
Hello @tomaszpirc , 

it's just about the case with my customers. In my case the default language (the global default set in the fluent locales admin) is the native language of the editors.

kind regards

tomaszpirc:
Getting back to @tractorcow comment:
> An edge case that we need to consider is, if we create a page in a non-default locale, and then switch to any other locale, what text should we see is the sitetree view? If we don't write to the base record, that view will literally have a node without any text at all. If you click it and enter edit mode, the entire form would also be blank. It's not a very good user experience.
> As I mentioned before, i don't really want to have the concept of a "default locale" affecting the CMS any more than it absolutely needs to. The only reason there is still a default at all, is because we need a fallback for when locale detection fails, or you enter the main home page without any url path prefix.

I understand what you are saying, but maybe we can find a middle ground:
- when page is created and there is no other translation, it always writes to the base table in that locale; 
- when page is created (or changed) in non-default language and there already exists a translation (no matter if it is default lang or not), we do not overwrite the content in the base table (first one takes precedence); 
- when page is created (or changed) in default language, we update also the content in the base table (default lang flag takes precedence).

The alternative - to update the base table only the first time (meaning when page is created not when first translation is made) - would also avoid our issue since editors usually create pages in default language, but would also mean that we have very old content in the base table and no way of updating base table.

What do you think?

pine3ree:
Hello @tomaszpirc,

I believe that in any case it would be good to avoid having mixed translations in the base table.
As explained by @tractorcow , since the global default locale may be removed in the future, we could ad a config static property like `$base_locale` to indicate which translation will be used a the initial field values for new translations.

Filling new translation localised form fields can be done in different ways:
1. Using the $translate fields' values in  the base table. This would require having the base table fields written to only when editing in the `base_locale`. Since new translations maybe added even years later when adding a new locale for the entire website we should keep the base table updated with the latest `base_locale` field values.
2. Using the`base_locale` translation field values in the Localised table, if existing. This would work even with empty $translate fields in the base table. I am not sure if and how empty translate fields in the base table may affect versioned dataobjects like when examining older versions (btw, I am still having issue in the History tab for site-tree with fluent) 

(What I did is option 1 using the deprecated global default locale setting as the base_locale)

kind regards

PR #704, opened 2021-04-29
BUG: GridField buttons UI improvements for versioned models.

mfendeksilverstripe:
# GridField buttons UI improvements for versioned models.

* `localise_actions_enabled` setting moved down to FluentExtension and now applied to all buttons UI cases (model, versioned model, site tree model)
* information panel enhancements moved to `FluentAdminTrait` so it's shared with versioned model
* save and delete action UI enhancements now also apply to versioned models

Screen Shot 2021-05-04 at 8 51 43 AM

### DIFF
--- a/src/Extension/Traits/FluentAdminTrait.php
+++ b/src/Extension/Traits/FluentAdminTrait.php
@@ -2,11 +2,13 @@
 
 namespace TractorCow\Fluent\Extension\Traits;
 
+use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\Form;
 use SilverStripe\Forms\FormAction;
+use SilverStripe\Forms\LiteralField;
 use SilverStripe\Forms\Tab;
 use SilverStripe\Forms\TabSet;
 use SilverStripe\ORM\DataObject;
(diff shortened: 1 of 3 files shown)


src/Extension/Traits/FluentAdminTrait.php
```diff
@@ -63,12 +67,19 @@ protected function updateFluentActions(
         // Flush data before checking actions
         $record->flushCache(true);
 
+        // These actions need to be executed before we bail out on archived check
```
tractorcow, line 70: This code should not be before the check below

```
            // Potentially no Locales have been created in the system yet.
            return;
```

src/Extension/Traits/FluentAdminTrait.php
```diff
@@ -663,4 +670,56 @@ protected function updateDeleteAction(FieldList $actions, DataObject $record): v
...
+            return;
+        }
+
+        if (!$record->hasExtension(FluentVersionedExtension::class)) {
+            // This is only relevant for versioned records (fluent)
+            return;
+        }
+
+        /** @var Tab $moreOptions */
+        $moreOptions = $actions->fieldByName('ActionMenus.MoreOptions');
```
tractorcow, line 694: My understanding is that MoreOptions is for sitetree only, right?

Do you want to soft-support non-sitetree models too?

tractorcow: Do non-sitetree objects have Information panel?

tractorcow:
General question: How well does this work with other versioned, but non-sitetree models?

src/Extension/Traits/FluentAdminTrait.php
mfendeksilverstripe: yeah, they do. It's a relatively recent addition though.

https://github.com/silverstripe/silverstripe-versioned/blob/1.7/src/VersionedGridFieldItemRequest.php#L308

mfendeksilverstripe:
@tractorcow I tested this via model admin on non-site tree models and the buttons UI looks good. I'll post some screenshots soon. One thing I noticed is that `VersionedGridFieldState` is not handled well. I'll raise that as a separate issue.

mfendeksilverstripe:
@tractorcow Added a screenshot to the description and raised a separate issue for the GridFieldState state https://github.com/tractorcow-farm/silverstripe-fluent/issues/706.

tractorcow:
In it goes then. :D

Issue #705, opened 2021-05-03
Fallbacks should only fallback to the correct locales


adrhumphreys:
Currently we have the following documentation for setting up fallbacks: https://github.com/tractorcow-farm/silverstripe-fluent/blob/5.0/docs/en/scenarios.md#content-inheritance-for-the-frontend

I think the extension should always be applied when `frontend_publish_required` is set to `false` or the config being updated. The reason for this is that I can't think of a situation where you would not want to only fallback to the locales you've selected as fallbacks

For example, if you have Maori, Japanese, Korea, and English then you could have a user viewing the Maori page and falling back to Japanese as it was the last written record. Then the next day it might fallback to English because that record got written to. Then the user refreshes the page and it's suddenly in Korean because someone updated the page

But maybe this is me mixing options and we need two configs one for fluent fallbacks (which I would assume is applied automatically when you have added a fallback (requiring no dev effort)) and then leaving `frontend_publish_required` alone?

tractorcow:
The default is meant to have frontend_publish_required set to true for fallbacks to work properly; It was added as an option for users to want to customise this behaviour though. Literally, it treats "last written" as the global fallback; Not really the best scenario, but that's why it's set to true by default. :)

Options are: 
 - leave frontend_publish_required as true
 - set frontend_publish_required but make sure that all locales fallback up the chain to one "best" default

Just to confirm:

> For example, if you have Maori, Japanese, Korea, and English then you could have a user viewing the Maori page and falling back to Japanese as it was the last written record. Then the next day it might fallback to English because that record got written to. Then the user refreshes the page and it's suddenly in Korean because someone updated the page

You are assigning the "fallback" properly in the locales section right? You aren't just leaving it blank for all your locales?

tractorcow:
I think I can see the problem,

What we really need is better granularity for frontend_publish_required:
 - true (must be published in this locale)
 - 'fallback' (must be published in this, or any fallback locale)
 - false (will fallback to last written record)

Alternatively:
 - true (must be published in this, or any fallback locale)
 - 'strict' (must be published in this exact locale)
 - false (will fallback to last written record)

tractorcow:
@mfendeksilverstripe @chrispenny what do you think about the above?

mfendeksilverstripe:
[...]

PR #711, opened 2021-05-20
NEW: Locale fallback configuration enhancement.

mfendeksilverstripe:
# NEW: Locale fallback configuration enhancement

* Fallback option added to both `cms_localisation_required` and `frontend_publish_required`
* `cms_publish_required` removed as it's deprecated
* change is not backwards compatible but transition is possible (mapping below):

`false` -> `any`
`true` -> `exact`

Implementation of https://github.com/tractorcow-farm/silverstripe-fluent/issues/705

### DIFF
--- a/_config/model.yml
+++ b/_config/model.yml
@@ -4,7 +4,7 @@ Name: fluentmodel
 SilverStripe\ORM\DataObject:
   # See FluentExtension
   cms_localisation_required: any
-  frontend_publish_required: none
+  frontend_publish_required: fallback
   # See FluentFilteredExtension
   apply_filtered_locales_to_stage: true
   # See FluentIsolatedExtension
--- a/_config/model.yml
+++ b/_config/model.yml
@@ -3,8 +3,8 @@ Name: fluentmodel
 ---
 SilverStripe\ORM\DataObject:
   # See FluentExtension
-  cms_localisation_required: false
-  frontend_publish_required: true
+  cms_localisation_required: any
+  frontend_publish_required: none
   # See FluentFilteredExtension
   apply_filtered_locales_to_stage: true
   # See FluentIsolatedExtension
@@ -12,4 +12,4 @@ SilverStripe\ORM\DataObject:
 
 # Prevent duplicate siteconfigs when localised
 SilverStripe\SiteConfig\SiteConfig:
-  frontend_publish_required: false
+  frontend_publish_required: any
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -66,6 +66,24 @@ class FluentExtension extends DataExtension
     const TRANSLATE_NONE = 'none';
 
     /**
+     * Content inheritance - content will be served from the following sources in this order:
+     * current locale
+     */
+    const INHERITANCE_MODE_NONE = 'none';
+
+    /**
+     * Content inheritance - content will be served from the following sources in this order:
+     * current locale, fallback locale
+     */
+    const INHERITANCE_MODE_FALLBACK = 'fallback';
+
+    /**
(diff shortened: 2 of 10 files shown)


src/Extension/FluentExtension.php
```diff
@@ -469,11 +487,24 @@ public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null)
...
+        $inheritanceMode = $this->getInheritanceMode();
+        if ($inheritanceMode === self::INHERITANCE_MODE_NONE) {
             $joinAlias = $this->getLocalisedTable($this->owner->baseTable(), $locale->Locale);
-            $where = "\"{$joinAlias}\".\"ID\" IS NOT NULL";
+            $where = sprintf('"%s"."ID" IS NOT NULL', $joinAlias);
             $query->addWhereAny($where);
+        } elseif ($inheritanceMode === self::INHERITANCE_MODE_FALLBACK) {
+            $conditions = [];
+
+            foreach ($this->owner->getLocalisedTables() as $table => $fields) {
```
tractorcow, line 499: instead of all tables, can we just use the base table please for each locale?

src/Extension/FluentExtension.php
```diff
@@ -1021,21 +1052,25 @@ public function updateFluentCMSField(FormField $field)
...
+    protected function getInheritanceMode(): string
     {
-        if (FluentState::singleton()->getIsFrontend()) {
-            return $this->owner->config()->get('frontend_publish_required');
-        }
-
-        if ($this->owner->config()->get('cms_publish_required') !== null) {
-            Deprecation::notice('5.0', 'Use cms_localisation_required instead');
+        $config = $this->owner->config();
+        $inheritanceMode = FluentState::singleton()->getIsFrontend()
```
tractorcow, line 1060: I suggest we detect is_boolean here, and map to the default values. It doesn't have to be a deprecation warning.

E.g. false == none, true === fallback

tractorcow (changes requested):
as commented

src/Extension/FluentExtension.php
```diff
@@ -65,6 +65,24 @@ class FluentExtension extends DataExtension
      */
     const TRANSLATE_NONE = 'none';
 
+    /**
+     * Content inheritance - content will be served from the following sources in this order:
+     * current locale
+     */
+    const INHERITANCE_MODE_NONE = 'none';
```
tractorcow, line 72: I feel like 'none' is a bit obtuse. How about "exact"? (no fallback, must be localised in exact locale)

_config/model.yml
```diff
@@ -3,13 +3,13 @@ Name: fluentmodel
 ---
 SilverStripe\ORM\DataObject:
   # See FluentExtension
-  cms_localisation_required: false
-  frontend_publish_required: true
+  cms_localisation_required: any
+  frontend_publish_required: none
```
tractorcow, line 7: Since this is master (new major version) happy to change default to `fallback` to give the best out of the box for the frontend.

tractorcow:
@mfendeksilverstripe great job :D

mfendeksilverstripe:
Thanks for feedback @tractorcow , I've pushed the requested changes, please review.

tractorcow (approved):

tractorcow:
@mfendeksilverstripe think we need a 6.0.0 tag, or still more breaking changes to go in? (Gut feel says it's a bit early)

mfendeksilverstripe:
I recommend waiting a bit @tractorcow as I have some more items on my list to go trough.

tractorcow:
@mfendeksilverstripe also, I don't mind this not being in 5.1 also (in a non-breaking way) but I can back-port it if someone really wants it.

PR #712, opened 2021-05-21
Feature: Allow timezones to be specified for each region

tractorcow:
Helpful for when a date needs to be shown in different timezones in any region.

Since DBDatetime doesn't internally support a timezone, we rely on storage in the server timezone, and only convert on rendering output. (otherwise, internal functions like strtotime are wrong).

### DIFF
--- a/src/Model/LocalDateTime.php
+++ b/src/Model/LocalDateTime.php
@@ -2,7 +2,7 @@
 
 namespace TractorCow\Fluent\Model;
 
-use DateTimeImmutable;
+use DateTime;
 use DateTimeZone;
 use Exception;
 use IntlDateFormatter;
@@ -122,9 +122,10 @@ class LocalDateTime extends DBDatetime
 
         // Parse from local timezone
         $timezone = $this->getTimezone() ?: date_default_timezone_get();
-        $localTime = new DateTimeImmutable($value, new DateTimeZone($timezone));
+        $localTime = new DateTime($value, new DateTimeZone($timezone));
         $localTime->setTimezone(new DateTimeZone(date_default_timezone_get())); // Store in server timezone
-        $this->value = $localTime->format('Y-m-d H:i:s');
+        $serverValue = $localTime->format('Y-m-d H:i:s');
+        $this->value = $serverValue;
         return $this;
     }
 }
--- a/src/Model/LocalDateTime.php
+++ b/src/Model/LocalDateTime.php
@@ -49,13 +49,6 @@ class LocalDateTime extends DBDatetime
         return $this;
     }
 
-    /**
-     * @param null $locale
-     * @param null $pattern
-     * @param int $dateLength
-     * @param int $timeLength
-     * @return IntlDateFormatter
-     */
     public function getCustomFormatter(
         $locale = null,
         $pattern = null,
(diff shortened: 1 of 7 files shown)


tractorcow:
@mfendeksilverstripe do you want to review this?

src/Model/LocalDateTime.php
```diff
@@ -0,0 +1,130 @@
...
+use SilverStripe\ORM\FieldType\DBDatetime;
+
+/**
+ * Stores dates in the same timezone as DBDateTime, but will format them
+ * based on the timezone of another locale.
+ *
+ * Note: getValue() date will NOT be in the specified timezone. Only dates formatted via ->Format()
+ * or ->getLocalValue() will be converted as expected.
+ */
+class LocalDateTime extends DBDatetime
```
mfendeksilverstripe, line 19: I'm wondering if it would be possible to keep this available via an extension. Generally, I prefer composition over inheritance as it's a more flexible approach. I'm thinking of a case where `DBDatetime` is subclassed by project's customisation. Is that case using `getLocalTime` effectively removes the customisation.

tractorcow: The problem is that there's no way to extend getCustomFormatter. The object has to be completely overwritten in order for a timezone to be injected.

tractorcow:
I've added a setter, so that dates from local times can also be parsed easily, and unit tests.

src/Model/LocalDateTime.php
tractorcow: I think this is why other frameworks use facades for extensions ;P It'd be easy then. Oh well!

mfendeksilverstripe: Right, all good then.

tests/php/Model/LocalDateTimeTest.php
```diff
@@ -0,0 +1,67 @@
...
+        // Convert to US timezone
+        $localisedDate = $date->getLocalTime();
+        $this->assertEquals('2021-02-18 06:59:59', $localisedDate->getLocalValue()); // US time, 5 hours before UTC
+
+        // Internal time is non-modified (original UTC timezone)
+        $this->assertEquals('2021-02-18 11:59:59', $localisedDate->getValue());
+
+        // Change to NZ timezone
+        $localisedDate->setTimezone('Pacific/Auckland');
+        $this->assertEquals('2021-02-19 00:59:59', $localisedDate->getLocalValue()); // NZ is 13 hours after UTC
```
mfendeksilverstripe, line 44: Can you please also add cases where we switch locale and see that the timezone changes? I see that you already have several locales set up with different timezones. You can use a dataProvider here to test several locales with just one test ;-)

anthonyvanbeek (approved):

tests/php/Model/LocalDateTimeTest.php
tractorcow: Good idea

tractorcow: Done :D

tractorcow:
postgres is having a dumb. Sorry not sure why (it's broken on merged branches too).

mfendeksilverstripe (approved):
This is a really nice feature ❤️ , feel free to merge this once tests pass.

Nice job 👍

mfendeksilverstripe:
I suggest to raise an issue for this and fix it later (maybe some missing double quotes somewhere?). I don't think this is a blocker for this PR though.

tractorcow:
Most of the tests pass, will need to debug postgres later.

Issue #713, opened 2021-05-27
v6 - Localised copy doesn't trigger when ClassName changes


mfendeksilverstripe:
# Localised copy doesn't trigger when ClassName changes

This may be an edge case but it's quite important for the Fluent + Elemental setup

## Test setup

`Page` -  "out of the box" model type
`CustomPage` - page with a has_one relation which uses `localised_copy` and this relation is automatically populated when a page is created (like ElementalArea)

* Create new `Page`, localise in `int` and `us`
* Current locale is `int`
* Change `Page` class to `CustomPage` and save
* This triggers the relation to be populated for the `int` locale but not for the `us` locale

The most common case is ElementalArea as this model is automatically attached when ClassName is changed (as the page relation becomes available).

## Possible solution

Something like if we have `localised_copy` setup then we wold search for all localised records of the current model and check if this relation is populated. If it's not populated we would duplicate the related model from the current locale and attach. This should happen `onAfterWrite()`. We should probably use writeWithoutVersion or raw query if needed to attach relations.

tractorcow:
> and this relation is automatically populated when a page is created (like ElementalArea)

Should localised_copy create a brand new record each time, or only copy an existing record to the new locale when it's localised? I'm not sure it's clear that's the expectation.

mfendeksilverstripe:
My understanding is that localised_copy means you always create a new record. For example if you copy to the same locale twice the second time will create a new record. I'll review how localised copy works currently in detail and ideally I'll put up a POC for review.

mfendeksilverstripe:
@tractorcow I've added a POC for this https://github.com/tractorcow-farm/silverstripe-fluent/pull/717

Please have a look and let me know your thoughts ;-)

PR #717, opened 2021-06-09
BUG: Localised copy support for ClassName changes.

mfendeksilverstripe:
# Localised copy support for ClassName changes.

Fix for https://github.com/tractorcow-farm/silverstripe-fluent/issues/713

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -603,9 +603,6 @@ class FluentExtension extends DataExtension
             return;
         }
 
-        // Detect class name change (used later to properly localised newly available relation data)
-        $owner->classNameChangeDetected = $owner->isInDB() && $owner->isChanged('ClassName');
-
         $this->makeLocalisedCopy();
 
         // If the record is not versioned, force change
@@ -637,7 +634,7 @@ class FluentExtension extends DataExtension
             return;
         }
 
-        if (!$owner->classNameChangeDetected) {
+        if (!$owner->isChanged('ClassName')) {
             // ClassName did not change so we can bail out
             return;
         }
@@ -654,6 +651,7 @@ class FluentExtension extends DataExtension
 
         // Get list of all localised instances of this model and duplicate relations if needed (if current one has it)
         $locales = $owner->getLocaleInstances();
+        $objectsToWrite = [];
 
         /** @var Locale $locale */
         foreach ($locales as $locale) {
@@ -678,14 +676,16 @@ class FluentExtension extends DataExtension
                 }
 
                 /** @var DataObject $localisedRecord */
-                $localisedRecord = FluentState::singleton()->withState(
-                    function (FluentState $state) use ($owner, $locale): ?DataObject {
-                        $state->setLocale($locale->Locale);
+                $localisedRecord = array_key_exists($locale->Locale, $objectsToWrite)
+                    ? $objectsToWrite[$locale->Locale]
+                    : FluentState::singleton()->withState(
+                        static function (FluentState $state) use ($owner, $locale): ?DataObject {
+                            $state->setLocale($locale->Locale);
 
-                        // Fetch localised record from other locale
-                        return DataObject::get_by_id($owner->ClassName, $owner->ID);
-                    }
-                );
+                            // Fetch localised record from other locale
+                            return DataObject::get_by_id($owner->ClassName, $owner->ID);
+                        }
+                    );
 
                 if (!$localisedRecord) {
                     // This is just a sanity check
@@ -696,7 +696,6 @@ class FluentExtension extends DataExtension
                 $localisedRelationID = $localisedRecord->{$relationIdField};
                 $localisedRelation = $localisedRecord->{$relation}();
 
-                // TODO decide if we want to override existing relation or not
                 if ((int) $localisedRelationID !== (int) $owner->{$relationIdField}
                     && $localisedRelation instanceof DataObject
                     && $localisedRelation->exists()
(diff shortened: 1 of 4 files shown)


mfendeksilverstripe:
@tractorcow @chrispenny Please review and let me know your feedback.

src/Extension/FluentExtension.php
```diff
@@ -593,16 +594,18 @@ public function augmentSQL(SQLSelect $query, DataQuery $dataQuery = null)
...
-
-        /** @var string $currentLocale */
         $currentLocale = FluentState::singleton()->getLocale();
+
         if (!$currentLocale) {
             return;
         }
 
+        // Detect class name change (used later to properly localised newly available relation data)
+        $owner->classNameChangeDetected = $owner->isInDB() && $owner->isChanged('ClassName');
```
tractorcow, line 607: you can detect isChanged in onAfterWrite. I prefer to do this there than adding an unnecessary temp variable.

You can also use isChanged('ID') to see if a record was created.

src/Extension/FluentExtension.php
```diff
@@ -617,6 +620,114 @@ public function onBeforeWrite()
...
+                    }
+                );
+
+                if (!$localisedRecord) {
+                    // This is just a sanity check
+                    continue;
+                }
+
+                $relationIdField = $relation . 'ID';
+                $localisedRelationID = $localisedRecord->{$relationIdField};
```
tractorcow, line 696: what if relation is polymorphic? (ID + ClassName field both).

src/Extension/FluentExtension.php
```diff
@@ -617,6 +620,114 @@ public function onBeforeWrite()
...
+                // Attach the duplicated relation to localised record
+                FluentState::singleton()->withState(
+                    function (FluentState $state) use ($locale, $localisedRecord, $relation, $duplicate): void {
+                        $state->setLocale($locale->Locale);
+
+                        $localisedRecord->setComponent($relation, $duplicate);
+
+                        if ($localisedRecord->hasExtension(Versioned::class)) {
+                            /** @var DataObject|Versioned $localisedRecord */
+                            $localisedRecord->writeWithoutVersion();
```
tractorcow, line 719: Why write without versioned? Just a ->write() in all cases would be enough?

src/Extension/FluentExtension.php
```diff
@@ -617,6 +620,114 @@ public function onBeforeWrite()
...
+        if ($stage !== Versioned::DRAFT) {
+            // Only draft stage is relevant for the duplication
+            return;
+        }
+
+        // Get list of all localised instances of this model and duplicate relations if needed (if current one has it)
+        $locales = $owner->getLocaleInstances();
+
+        /** @var Locale $locale */
+        foreach ($locales as $locale) {
```
tractorcow, line 659: maybe if you loop over $this->Locales() instead, you can re-use the logic in RecordLocale to simplify some of this code.

E.g. 


```php
$localisedRecord = $recordLocale->getRecord();
```

tractorcow (changes requested):

tractorcow:
@mfendeksilverstripe looks good; Asked some questions, and added some suggestions that can simplify the code a bit and re-use existing work :)

src/Extension/FluentExtension.php
```diff
@@ -617,6 +620,114 @@ public function onBeforeWrite()
...
+            if ($locale->Locale === $currentLocale) {
+                // We only need to handle other locale instances, current locale doesn't need updates
+                continue;
+            }
+
+            $relations = (array) $owner->config()->get('localised_copy');
```
tractorcow, line 665: localised_copy is exclusively for has_one right? just checking :)

mfendeksilverstripe: That's correct.

src/Extension/FluentExtension.php
mfendeksilverstripe: I had a look at this but it I couldn't find a way to simplify this further - the goal is to iterate over all locales which have localised record except the current locale. I'm already using the existing functionality to search for such locales via `getLocaleInstances`. I can simplify how the data changes are made though (one write per model).

src/Extension/FluentExtension.php
mfendeksilverstripe: I think it's sufficient to check only ID for the purpose of validating if we need to do an update or not. Note that later, when we actually assign the new relation we use `setComponent` which also covers the ClassName.

src/Extension/FluentExtension.php
tractorcow: you could put those into a condition and `continue` in that case. still less code, and you can use RecordLocale to check if the localisation exists.

mfendeksilverstripe: Not sure I follow - we are already looping a list of locales in which the model is localised (has localised record). I'm only skipping the current locale as I can't exclude it from the list as `getLocaleInstances` returns an `array` not an `ArrayList`.

mfendeksilverstripe:
@tractorcow I've pushed up new changes, please review.

src/Extension/FluentExtension.php
tractorcow: I think the problem I have is that getLocaleInstances() is solving the same problem that the Locales() method had already solved, so we end up with a lot of duplication of effort.

without this method, we could do this


```
foreach ($this->Locales() as $recordLocale) {
  // simpler "saved in this locale?" check
  if (!$recordLocale->IsDraft()) {
    continue;
  } 
$localisedRecord  = $recordLocale->getRecord();
// rest of code
}
```

src/Extension/FluentExtension.php
tractorcow: the ID of two different polymorphic classes can be the same, but those are different objects. :)

E.g. polymorphic relation to a File ID = 5 could be replaced with a Page ID = 5.

tractorcow:
Hi @mfendeksilverstripe want to say again what a great job you have done on this PR so far. I am being pedantic I know, but I want to check I have covered every base to save issues later on down the line. :)

Let me know if you want me to do a suggested patch to this PR before we merge; I want to do a semi-review of the codebase before 6.0.0 anyway.

tractorcow:
hi @mfendeksilverstripe I had more of a think about it, and I think we can probably worry about optimisation / code cleanup later. I have to spend a bit of time to do a review of master before releasing 6.0.0 anyway.

The only fix I think is blocking is the polymorphic relation, which is a real bug still.

src/Extension/FluentExtension.php
tractorcow: As mentioned below, I can tidy this up later; better to get what's already done in the code for now. :)

src/Extension/FluentExtension.php
tractorcow: you could just query the objects directly without trying to check the fields

```php
$ownerRelation = $owner->{$relation}();
$localisedRelation = $localisedRecord->{$relation}();
if ($ownerRelation->ID !== $localisedRelation->ID || get_class($ownerRelation) !== get_class($localisedRelation)) {
  continue;
}
```

tractorcow:
I left a suggestion above on a quick and easy way to solve that :D

mfendeksilverstripe:
Thanks for help @tractorcow , I added the small changes you requested. Please review.

tractorcow (approved):

tractorcow:
it's done :D

tractorcow:
I made my suggested cleanup changes in https://github.com/tractorcow-farm/silverstripe-fluent/pull/719

PR #719, opened 2021-06-27
Consolidate some redundant logic in code

tractorcow:
@mfendeksilverstripe these are the suggested changes I had in mind to follow up to #717

I switched out some of the logic in your loopy for checks / queries that were already handled in RecordLocale.

getLocaleInstances was also simplified slightly by re-using this logic.

I have another PR coming up (will take some time to progress) that adjusts the internal storage for localisations to use the localeID

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -666,7 +666,6 @@ class FluentExtension extends DataExtension
                 continue;
             }
 
-            $locale = $recordLocale->getLocaleObject();
             if ($recordLocale->getLocale() === $currentLocale) {
                 // We only need to handle other locale instances, current locale doesn't need updates
                 continue;
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -622,6 +622,16 @@ class FluentExtension extends DataExtension
      */
     public function onAfterWrite(): void
     {
+        $this->handleClassChanged();
+    }
+
+    /**
+     * If an object is changed to another class, we should trigger localised copy
+     *
+     * @throws ValidationException
+     */
+    protected function handleClassChanged(): void
+    {
         $owner = $this->owner;
         $stage = Versioned::get_stage() ?: Versioned::DRAFT;
         $currentLocale = FluentState::singleton()->getLocale();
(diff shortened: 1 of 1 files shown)


src/Extension/FluentExtension.php
```diff
@@ -650,55 +660,42 @@ public function onAfterWrite(): void
...
+        foreach ($this->owner->Locales() as $recordLocale) {
+            // Skip locales this record isn't localised in
+            if (!$recordLocale->IsDraft()) {
+                continue;
+            }
 
-        /** @var Locale $locale */
-        foreach ($locales as $locale) {
-            if ($locale->Locale === $currentLocale) {
+            $locale = $recordLocale->getLocaleObject();
```
mfendeksilverstripe, line 669: [nitpick] This looks like unused variable.
```suggestion
```

src/Extension/FluentExtension.php
```diff
@@ -707,20 +704,18 @@ static function (FluentState $state) use ($owner, $locale): ?DataObject {
...
 
-        foreach ($objectsToWrite as $localeCode => $localisedRecord) {
-            FluentState::singleton()->withState(
-                static function (FluentState $state) use ($localeCode, $localisedRecord): void {
-                    $state->setLocale($localeCode);
-                    $localisedRecord->write();
-                }
-            );
+            // Update localised record if any localised copies took place
+            if ($hasRecordsToWrite) {
```
mfendeksilverstripe, line 711: [nitpick] Would prefer early exit approach here to reduce code nesting.

mfendeksilverstripe (approved):
Looks really great 👍 , the code is much cleaner and easier to understand. I left two nitpicks but these are very minor things. The tests are passing on my local but I noticed that the Travis test suite is not running.

src/Extension/FluentExtension.php
tractorcow: ah good spotting.

src/Extension/FluentExtension.php
tractorcow: you mean 

```php
if (!$hasRecordsToWrite) {
  continue
}
FluentState::singleton()->withState(
    static function (FluentState $state) use ($recordLocale, $localisedRecord): void {
        $state->setLocale($recordLocale->getLocale());
        $localisedRecord->write();
    }
);
```

mfendeksilverstripe: That's correct ;-)

tractorcow: I'll update when I got some time :D

PR #720, opened 2021-06-27
[WIP] Localise records by locale ID not by code

tractorcow:
Still in progress, needs to be tested, and I need to come up with a migration plan. Ideally upgrading to 6.0 should just be a dev/build, and missing records would be fixed.

The advantage of this is that now, if you have localised your site in one locale, you can actually tweak or change the locale code without having to re-enter all of your content. E.g. if your locale changes from en_NZ to en_AU, your localisations for en_NZ don't disappear, but just appear as en_AU.

Theoretically, in the future we could also allow multiple locales to share the same locale code (since we would be keying by locale ID).

However, I'm not 100% sure we should even merge this change. The downside to this is that localisations are now tied to locale records.

Also, it's possible that you may not always want to tie a localisation row to a locale record; The localisations now follow that record, but maybe it would make switching localisations to ANOTHER locale object harder?

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -622,6 +622,16 @@ class FluentExtension extends DataExtension
      */
     public function onAfterWrite(): void
     {
+        $this->handleClassChanged();
+    }
+
+    /**
+     * If an object is changed to another class, we should trigger localised copy
+     *
+     * @throws ValidationException
+     */
+    protected function handleClassChanged(): void
+    {
         $owner = $this->owner;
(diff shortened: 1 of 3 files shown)


tractorcow:
Sorry, I did base this PR on https://github.com/tractorcow-farm/silverstripe-fluent/pull/719 as well, so ignore the first commit. ;)

tractorcow:
As discussed, we might want to store both locale and locale ID (instead of only one) and allowing the key to be chosen conditionally by config.

PR #722, opened 2021-07-12
feat(locale): add hook to update url segment

oliver-norden:
Allow for headless usage where the locale/url segment in fluent differ from the front-end locale routing.

### DIFF
--- a/src/Model/Locale.php
+++ b/src/Model/Locale.php
@@ -237,12 +237,15 @@ class Locale extends DataObject implements PermissionProvider
     public function getURLSegment()
     {
         $segment = $this->getField('URLSegment');
-        if ($segment) {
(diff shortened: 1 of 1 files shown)


tractorcow:
Hi @oliver-norden , shouldn't the URLSegment be the same as the frontend? I use fluent in headless CMS, but still configure these via the CMS.

E.g. URLSegment = 'au-en/cairns' for my nextjs routes mapped to [locale]/[region].js

tractorcow:
That being said, I think that this PR is still good, but it may not be needed.

oliver-norden:
Hello, yes you are correct. I just had some local issues where it would seem like the url-segment would revert back after i changed it to what I wanted it to be

Issue #737, opened 2021-11-08
Fallback locales sorting doesn't work correctly


shchedrakov:
In my project fallback locales should be organized like a tree structure. Where `en_US` is a global international locale and each object must be created on the locale at first.

If a user creates a new page on `fr_BE` the application should fill the page fields with `fr_CA` localization strings at first. If the page doesn't have `fr_CA` localization then It should be filled with `fr_FR`, then `en_US`.

![image](https://user-images.githubusercontent.com/9498638/140803656-37a3d624-8659-40bf-81a4-ec3ce749355f.png)

So, to set up such behavior I created the next fallback locales list:

![image](https://user-images.githubusercontent.com/9498638/140806131-a959c7c8-5948-402f-be41-dcf62246f10e.png)

But I found that it doesn't work as I expected. The first fallback locale was `en_US` instead of `fr_CA`. I tried to put `en_US` at the top of the list but it did work too. 

During debugging I found out that the `Fluent_FallbackLocale.Sort` field is ignored during the fallback locales chain creation. And I managed to fix it by editing `getChain()` method of `TractorCow\Fluent\Model\Locale` object.

Origin method snippet:
```php
/**
 * Get chain of all locales that should be preferred when this locale is current
 *
 * @return ArrayList
 */
public function getChain()
{
    if ($this->chain) {
        return $this->chain;
    }

    // Build list
    $this->chain = ArrayList::create();
    $this->chain->push($this);
    $this->chain->merge($this->Fallbacks());

    return $this->chain;
}
```

Edited method snippet:
```php
/**
 * Get chain of all locales that should be preferred when this locale is current
 *
 * @return ArrayList
 */
public function getChain()
{
    if ($this->chain) {
        return $this->chain;
    }

    $this->chain = ArrayList::create();

    // Push the current locale as the first fallback.
    $this->chain->push($this);

    // Get the current locale fallbacks and sort them by "Sort" field.
    $fallbacks = $this->FallbackLocales()->sort('Sort');
    foreach ($fallbacks as $fallback) {
        $this->chain->push($fallback->Locale());
    }

    return $this->chain;
}
```

tractorcow:
Fixed in #738

PR #738, opened 2021-11-08
Fix fallback locales chain

shchedrakov:
See https://github.com/tractorcow-farm/silverstripe-fluent/issues/737


tractorcow (approved):

tractorcow:
Thanks for this, it's a good catch!

Issue #740, opened 2022-01-10
BUG: Locale link redirects to empty page after initial localisation of a page


mikenuguid:
If I localise a page on the default locale and try to localise that to another locale say Te reo and used `copy to` on the localisation tab, I noticed that when I click on the Locale link, it doesn't redirect me to the locale but to an empty page. What I noticed is that the end of URL is changed to `RecordLocales` after using copy to. This goes away when I refresh the CMS page though

**Screenshots:**

![image](https://user-images.githubusercontent.com/44601742/148708173-8601c8b8-3020-44a8-a4fc-d7ceff0ef90c.png)

This is the page I get redirected to if I click that Te reo Maori New Zealand link (after using copy to).

![image](https://user-images.githubusercontent.com/44601742/148708196-ef3523f5-a011-4110-9f49-fb18e6746eda.png)

mikenuguid:
@mfendeksilverstripe fyi

tractorcow:
Does the actual copy work though? What versions of fluent + silverstripe are you using?

mikenuguid:
@tractorcow yes it does. Fluent: `dev-master`, Silverstripe: `4.8.0`

tractorcow:
ok that's cool. it's probably just an ajax action that's not triggering properly in the admin.

mfendeksilverstripe:
@mikenuguid @tractorcow Fixed in https://github.com/tractorcow-farm/silverstripe-fluent/pull/742

tractorcow:
fixed with #742

mfendeksilverstripe:
^ @mikenuguid

mikenuguid:
thanks @mfendeksilverstripe 🙂

PR #742, opened 2022-01-13
BUG: Localisation manager / Locale link fix for AJAX requests.

mfendeksilverstripe:
# BUG: Localisation manager / Locale link fix for AJAX requests.

* Now using CMS edit link instead of current URL as the latter can change in some situations like AJAX requests

Fix for https://github.com/tractorcow-farm/silverstripe-fluent/issues/740


tractorcow (approved):

tractorcow:
Thanks, agree this makes the behaviour much more reliable.

PR #743, opened 2022-01-27
NEW: Copy to locale enhancements.

mfendeksilverstripe:
# NEW: Copy to locale enhancements

It turns out that copy to locale feature is really good! This enhancement contains no functional changes but it opens this feature to be used via public API. This is useful in situations where GridField UI is not appropriate solution like migrations, custom UI or similar.

* generic API is available via a service
* `FluentExtension` provides a simplified API intended for the most common case


mfendeksilverstripe:
@chrispenny Might be worth checking this out ;-)

tractorcow (approved):

mfendeksilverstripe:
We don't have tests running here but I ran full test suite on my local (vanilla instal) and also project integration. Both are passing.

Screen Shot 2022-01-28 at 7 02 00 AM

Screen Shot 2022-01-28 at 8 14 05 AM

chrispenny:
Thanks so much for this! Great new addition.

PR #744, opened 2022-02-04
BUG: Localisation manager fixes

mfendeksilverstripe:
# BUG: Localisation manager fixes

* BUG: `FluentVersionedExtension:stagesDifferInLocale()` now uses provided `$locale` and falls back to current locale instead of using current locale in all cases
*  BUG: `RecordLocale:: IsPublished()` corrected - removed legacy base record check and now checks the published state of the source locale instead
* ENH: `updateRecordLocaleStagesDiffer` extension point added to provide some customisation options

## Tests

Tests are passing on my local.

Screen Shot 2022-02-04 at 1 52 21 PM

### DIFF
--- a/src/Extension/FluentVersionedExtension.php
+++ b/src/Extension/FluentVersionedExtension.php
@@ -401,13 +401,13 @@ class FluentVersionedExtension extends FluentExtension implements Resettable
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


tractorcow (approved):

tractorcow:
Seems ok, doesn't break anything. :P

PR #750, opened 2022-02-28
BUG Fix copyFluent action writing all locales including original

renskorswagen:
Copy 'Locale' to other locales now writes to every locale. I think this is unnecessary and in some cases causes bugs.

Now skipping the original locale.

### DIFF
--- a/src/Extension/Traits/FluentAdminTrait.php
+++ b/src/Extension/Traits/FluentAdminTrait.php
@@ -277,7 +277,14 @@ trait FluentAdminTrait
         $record = $form->getRecord();
         $record->flushCache(true);
 
-        $this->inEveryLocale(function () use ($record) {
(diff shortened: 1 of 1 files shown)


src/Extension/Traits/FluentAdminTrait.php
```diff
@@ -277,7 +277,14 @@ public function copyFluent($data, $form)
...
         $record->flushCache(true);
 
-        $this->inEveryLocale(function () use ($record) {
+        $originalLocale = Locale::getCurrentLocale();
+
+        $this->inEveryLocale(function () use ($record, $originalLocale) {
+            // Skip original locale
+            if ($locale->ID == $originalLocale->ID) {
+                return;
+            }
```
tractorcow, line 286: Where is `$locale` defined?

tractorcow:
Thanks @renskorswagen

PR #752, opened 2022-03-09
NEW: Locale performance / extensibility enhancements.

mfendeksilverstripe:
# NEW: Locale performance / extensibility enhancements.

* **Performance**: All instances of `Locale::get()` replaced with `Locale::getCached()` to take advantage of in-memory cache
* **Extensibility**: [Cachable model injection point](https://github.com/tractorcow-farm/silverstripe-fluent/blob/master/src/Model/CachableModel.php#L25) now covers all locale calls which allows consistency in case this is used to customise cached locales data
* **Cleanup**: In-memory cache for locale code lookup removed as this was already operating on top of another memory cache (cached models), the performance gain is negligible and is not worth keeping the additional complexity, also this in-memory cache was not flushed between unit tests and could interfere with [Cachable model injection point](https://github.com/tractorcow-farm/silverstripe-fluent/blob/master/src/Model/CachableModel.php#L25)

Tests passing on my local.

Screen Shot 2022-03-10 at 12 32 25 PM

### DIFF
--- a/src/Extension/FluentFilteredExtension.php
+++ b/src/Extension/FluentFilteredExtension.php
@@ -39,7 +39,7 @@ class FluentFilteredExtension extends DataExtension
 
     public function updateCMSFields(FieldList $fields)
     {
-        $locales = Locale::get();
+        $locales = Locale::getCached();
 
         // If there are no Locales, then we're not adding any fields.
         if ($locales->count() === 0) {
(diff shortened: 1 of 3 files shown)


tractorcow (approved):

tractorcow:
Nice job,

Issue #751, opened 2022-03-08
No content in Localisation tab for user without full admin permission


helenaerni:
We have noticed that in the localization tab no content is displayed for the content editors, although they have all language permissions. 
The easiest way to explain this is to look at the screenshots, above full admin permissions, below content editor permissions:
![locale-tab](https://user-images.githubusercontent.com/57763604/157206139-682885ef-2ed8-48ad-a3e8-ceccd5c3b2b3.jpg)
.

tractorcow:
that's odd, users should be able to see that table if they have any CMS section permission at all (e.g. pages).

Which version of fluent are you on?

tractorcow:
I'm actually looking at the permission handling today; I'll see if I can track down the cause.

helenaerni:
> that's odd, users should be able to see that table if they have any CMS section permission at all (e.g. pages).
> In your second example, I assume "Access to pages section" is still on?
Yes the pages permission is on and works, they can also translate the pages with the translations button on the bottom section or copy the page into a translation. It's just the overview not showing up.

> Which version of fluent are you on?
Fluent 5 (update befor I posted), Silverstripe 4.9

tractorcow:
One issue I've had is that the "administrator" checkbox force-checks every checkbox, which isn't what you always want (e.g. if you want administrator, but only in one region). I'm trying to find a solution to that. I might need to block / replace that entire script in the backend.

tractorcow:
hi @helenaerni I couldn't quite get the same symptoms you did, but did address another issue over at https://github.com/tractorcow-farm/silverstripe-fluent/pull/753

To demonstrate an example of a non-admin being able to see the localisation records.

![image](https://user-images.githubusercontent.com/936064/157588891-7105cd84-a80f-4b62-942f-b91d516b6801.png)

![image](https://user-images.githubusercontent.com/936064/157588906-61686b4c-e49f-4905-a6df-4db3128cb92d.png)

![image](https://user-images.githubusercontent.com/936064/157588940-6de15ab9-7fb6-405e-bdaa-cc2b3715f3ac.png)

tractorcow:
Ok, i think I know what's happened.

https://github.com/tractorcow-farm/silverstripe-fluent/pull/698

This was fixed, but not tagged. I'll tag a new 5.0.x for you :)

tractorcow:
There you go 

[...]

PR #753, opened 2022-03-10
fix: issue with "all cms sections" permission erroneously handling fluent permissions

tractorcow:
I attempted to reproduce https://github.com/tractorcow-farm/silverstripe-fluent/issues/751 but I had a few problems getting the same symptoms.

In any case, I fixed the permissions code so that the locale-permissions no longer get set to readonly if you check "all cms permissions".


tractorcow:
todo: Once this is merged, I'll get 5.1.0

anthonyvanbeek (approved):

PR #755, opened 2022-03-24
MNT Use v0.1 of ci workflow (4.6 branch)

emteknetnz:
Use v0.1 version of the ci.yml file, which is a special version of the workflow that is re-released whenever a new tag matching 0.1.x is created.  Effectively simulates ^ caret support for github workflows which do not support carets.


tractorcow:
thanks @emteknetnz

PR #758, opened 2022-03-25
BUG: PHP 8 compatibility patch.

mfendeksilverstripe:
# BUG: PHP 8 compatibility patch.

* PHP 8 compatibility patch

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "issues": "http://github.com/tractorcow-farm/silverstripe-fluent/issues"
     },
     "require": {
-        "php": "^7.4 || ^8",
+        "php": "^7.1 || ^8",
         "silverstripe/vendor-plugin": "^1",
         "silverstripe/framework": "^4.7",
         "silverstripe/admin": "^1",
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "issues": "http://github.com/tractorcow-farm/silverstripe-fluent/issues"
     },
     "require": {
-        "php": "^7.1 || ^8.0",
+        "php": "^7.4 || ^8",
         "silverstripe/vendor-plugin": "^1",
         "silverstripe/framework": "^4.7",
         "silverstripe/admin": "^1",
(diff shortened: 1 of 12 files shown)


composer.json
```diff
@@ -30,7 +30,7 @@
         "issues": "http://github.com/tractorcow-farm/silverstripe-fluent/issues"
     },
     "require": {
-        "php": "^7.1 || ^8.0",
+        "php": "^7.4 || ^8",
```
tractorcow, line 33: these changes don't require php 7.4 do they?

mfendeksilverstripe: No, I updated this to have the versions consistent with other modules. I think older version than 7.4 is no longer supported.

tractorcow: Sure, we can avoid supporting those versions, but let's leave the dependency as low as possible to prevent unhappy people. :)

If we explicitly need syntax we know isn't supported on those versions, it's a good reason to bump it.

tractorcow: Ah, it seems framework 4.10 uses php 7.3 minimum, but 4.9 and below is still 7.1. Good to know.

mfendeksilverstripe: Ok, I'll update the minimum versions and let's see if the tests pass.

mfendeksilverstripe:
@tractorcow Looks like there are no issues with PHP 7.1. Happy to keep the minimum version unchanged.

tractorcow (approved):

tractorcow:
Feel free to merge if you are happy you are done 😊

composer.json
maxime-rainville: Just to clarify, Silverstripe CMS 4.10 requires PHP7.3 or above. The reason for that is that we can't test PHP 7.2 or below because they don't work with PHPUnit 9.5.

It's probably not the end of the world if Fluent can be installed with PHP7.1 ... just be ware that it's possible that you could introduce a PHP7.1 bug that won't be detected by your CI.

mfendeksilverstripe: Good to know, thanks.

Issue #759, opened 2022-03-27
BUG: Legacy permission code migration


mfendeksilverstripe:
# BUG: Legacy permission code migration

This is what I get when running dev/build.

```
ERROR [Notice]: Undefined variable: 1
IN GET dev/build
Line 148 in /var/www/mysite/www/vendor/tractorcow/silverstripe-fluent/src/Model/Locale.php

Source
======
  139: 
  140:     public function requireDefaultRecords()
  141:     {
  142:         parent::requireDefaultRecords();
  143: 
  144:         // Migrate legacy permission codes to new codes
  145:         $permissions = Permission::get()->filter('Code:StartsWith', 'CMS_ACCESS_Fluent_');
  146:         $count = $permissions->count();
  147:         if ($count) {
* 148:             DB::alteration_message("Migrating ${$count} old fluent permissions", 'changed');
  149:         }
  150:         foreach ($permissions as $permission) {
  151:             $permission->Code = str_replace('CMS_ACCESS_Fluent_', 'Fluent_', $permission->Code);
  152:             $permission->write();
  153:         }
  154:     }
```

## Related pull requests

https://github.com/tractorcow-farm/silverstripe-fluent/pull/760

mfendeksilverstripe:
Fixed in https://github.com/tractorcow-farm/silverstripe-fluent/pull/760

PR #760, opened 2022-03-27
BUG: Legacy permissions migration fix.

mfendeksilverstripe:
# BUG: Legacy permissions migration fix.

Fix for https://github.com/tractorcow-farm/silverstripe-fluent/issues/759

Looks like this issue is caused by the log message which is shown during dev/build when legacy permissions are migrated.


tractorcow (approved):

tractorcow:
Ah, sorry about that, i'll backport to the older branch too.

PR #763, opened 2022-04-13
ENH PHP 8.1 compatibility

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10250

### DIFF
--- a/src/Extension/FluentDirectorExtension.php
+++ b/src/Extension/FluentDirectorExtension.php
@@ -123,7 +123,7 @@ class FluentDirectorExtension extends Extension
             $url = $localeObj->getURLSegment();
 
             // apply encode so we could route urls that contain multi-byte charaters
-            $url = urlencode($url);
+            $url = urlencode($url ?? '');
 
             // Apply to nested page url
             $controller = $this->getRuleController($originalRules['$URLSegment//$Action/$ID/$OtherID'], $localeObj);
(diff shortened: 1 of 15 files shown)


src/Extension/FluentDirectorExtension.php
```diff
@@ -166,9 +166,9 @@ protected function getRuleController($existingRule, $localeObj)
      */
     protected function insertRuleBefore(array $rules, $key, array $rule, $prependIfMissing = true)
     {
-        $i = array_search($key, array_keys($rules));
+        $i = array_search($key, array_keys($rules ?? []));
```
tractorcow, line 169: I assume `array` param type precludes null values. Is that not true in 8.1 anymore?

tractorcow: ok,  so it's more to do with changes to internal function arguments than typehints, apparently. I think in this case we can rely on the typehint. :)

tractorcow:
In general, not harmless, but lots of unnecessary nullable checks. I can approve if you are ready.

emteknetnz:
PHP 8.1 throws deprecation warnings if you pass null into non-nullable native functions now.  A whole lot of fun.

Solution was a code writer which does create lots of unnecessary nullable checks.  We've done the same on framework and all other supported modules

Happy to have approved, though I'd like to keep draft for now and unmerged on the chance that there's something that was missed

emteknetnz:
@tractorcow This is ready for review now. Thanks.

GuySartorelli:
@tractorcow there's some discussion in silverstripe/silverstripe-admin#1294 about the general approach which may be useful context.

tractorcow (approved):

emteknetnz:
@tractorcow could you please tag 4.7.0, we'll need this to ensure the next CMS release is PHP 8.1 compatible. Thanks

PR #765, opened 2022-04-28
Minor grammatical correction

oddnoc:
Had "you" rather than "to" to introduce the infinitive "to make".


tractorcow:
@oddnoc did you merge from / to the wrong branch? Looks like there's a lot of diffs in there :)

oddnoc:
Definitely. Github incorrectly guessed 5 at first.

PR #777, opened 2023-02-07
wip: silverstripe 5 compatibility 

tractorcow:
This is just a preview PR to be replaced once the current WIP is ready to merge.

Issue #776, opened 2023-02-04
SilverStripe 5 compatibility


tractorcow:
Please comment on this ticket for adding features / development effort to go into the SilverStripe 5 compatibility roadmap.

This release will target fluent 7.0

maxime-rainville:
FYI we created a temporary fork so we could devolped a CMS 5 compatible of Fluent: https://github.com/silverstripe/silverstripe-fluent

If OK with you, I'll just create a `7` branch from the `6` branch and open a PR from our fork.

Also part of our post beta work will be to draft an upgrade guide form CMS4 to CMS 5. That will involved some Fluent upgrade docs as well. @mfendeksilverstripe as already brief us.

tractorcow:
Sure thing, I'll try to respond promptly to updates on github to make sure you get the feedback you need.

tractorcow:
Master is now aliased to 7.x-dev, and I have updated the default branch + branch protection rules. 

https://github.com/tractorcow-farm/silverstripe-fluent/blob/master/composer.json

tractorcow:
I created a preview PR to review the current diff https://github.com/tractorcow-farm/silverstripe-fluent/pull/777, but will close it once you have an updated PR for merge.

maxime-rainville:
@tractorcow Thanks. I'll try to have the CMS Squad prioritise the work on Fluent this week.

GuySartorelli:
New PR at https://github.com/tractorcow-farm/silverstripe-fluent/pull/779

PR #779, opened 2023-02-16
DEP Silverstripe CMS 5 compatibility

GuySartorelli:
Replacement for https://github.com/tractorcow-farm/silverstripe-fluent/pull/777
Rebased on master. This is the same code that is in the beta, but with references to `silverstripe/silverstripe-fluent` reverted back to `tractorcow/silverstripe-fluent`

CI is green in our fork: https://github.com/creative-commoners/silverstripe-fluent/actions/runs/4190234023

## After merging
Ideally this would be tagged `7.0.0-beta1` - we can then (potentially) update our recipes to point back here instead of the fork - but even if we don't, it gives people the option of installing the beta of this module with the correct composer name.

## Parent issue
- https://github.com/tractorcow-farm/silverstripe-fluent/issues/776


GuySartorelli:
Added one last commit to remove deprecated methods which were deprecated in fluent 5 and aren't called in any of the code in 6 or 7.

GuySartorelli:
Not sure why scrutinizer is angry (can't install npm 18 for some reason?), but @tractorcow unless you've been regularly checking it I'd recommend either removing it entirely or, as a less drastic approach, accept the scrutinizer failure for this PR and handle that separately.

package.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "tractorcow-fluent",
-  "version": "4.0.0",
   "description": "Fluent localisation module for SilverStrip CMS",
   "engines": {
-    "node": "^10.x"
+    "node": "^18.x"
   },
   "scripts": {
-    "build": "NODE_ENV=production webpack -p --bail --progress",
+    "build": "yarn && yarn lint && rm -rf client/dist/* && NODE_ENV=production webpack --mode production --bail --progress",
```
mfendeksilverstripe, line 8: Why is the `rm` command present? I don't see this in any other modules, for example https://github.com/silverstripe/silverstripe-linkfield/blob/1/package.json#L20

tests/php/Extension/FluentExtensionTest.php
```diff
@@ -289,12 +289,16 @@ public function testLocalisedCopyClassNameChange(): void
...
      */
-    public function testLocalisedFieldsCanBeSorted($locale, array $sortArgs, $expected)
+    public function testLocalisedFieldsCanBeSorted($locale, array $sortArgs, $expected, $useOrderBy = false)
     {
-        FluentState::singleton()->withState(function (FluentState $newState) use ($locale, $sortArgs, $expected) {
+        FluentState::singleton()->withState(function (FluentState $newState) use ($locale, $sortArgs, $expected, $useOrderBy) {
             $newState->setLocale($locale);
 
-            $records = LocalisedParent::get()->sort(...$sortArgs);
+            if ($useOrderBy) {
```
mfendeksilverstripe, line 297: [nitpick] Use ternary operator.

mfendeksilverstripe:
Left some comments but the changes look good so far 👍

package.json
GuySartorelli: linkfield isn't a supported module.
Look at any supported module's CMS 5 compatible version, e.g. https://github.com/silverstripe/silverstripe-subsites/blob/3.0/package.json#L4

It's to protect against old files from earlier builds sticking around (e.g. as a result of webpack copy), or from people putting things in dist manually. We should always be able to blow away dist and rebuild and nothing will be lost - this ensures that's always the case.

tests/php/Extension/FluentExtensionTest.php
GuySartorelli: That's subjective, so unless there's some specific standard that this module adheres to which requires the ternary here I'm inclined to not make this change in this PR.

mfendeksilverstripe: Agree, hence the nitpick flag. Happy to keep it as is.

maxime-rainville:
Just approved the build to run.

tractorcow (approved):

tractorcow:
Merge at will. let me know if you need help releasing /etc.

maxime-rainville:
I just:
- merged the PR, 
- created  `7.0` and `7` branch from master,
- Tag a 7.0.0-beta1 from 7.0.

tractorcow:
Nice job :D

Issue #775, opened 2023-01-23
CMSPreview should be considered "frontend"


chrispenny:
A common complaint that we're receiving from authors is that when they use the CMS Preview window, they are seeing things that they don't expect, EG: content that is not localised, or content that should have been filtered out with their assignment of "Filtered Locales". If the author browses to the frontend with `?stage=Stage`, then they will receive the expected behaviour.

Currently, `CMSPreview` is treated as "backend" (`isFrontend = false`). This means that localisation and Filtered Locale logic is applied as if you were browsing the CMS. Which, I get that we are "in the CMS", but the practical purpose of the Preview window is to present the frontend experience that users will receive once the page is published.

`InitStateMiddleware::getIsFrontend()` currently contains a specific condition to treat `CMSPreview` as "backend".

```php
// If using the CMS preview, do not treat the site as frontend
if ($request->getVar('CMSPreview')) {
    return false;
}
```

I would like to propose removing this condition, so that the Preview window provides the same experience as browsing to the frontend with `?stage=Stage`.

tractorcow:
Yes I think you are right. preview should be "as frontend", so admin filters don't make sense.

The only point of having an admin filter is so that you could see potential items (unfiltered) in order to make them available on the frontend.

chrispenny:
I was able to find the reason why we (probably) made CMS preview be treated as "backend".

If you are browsing the CMS in Split or Preview mode, and you click on a page that is not localised, the Preview window redirects you to the 404 page - both in the Preview window (which would be expected), but also in the CMS edit form (which is unexpected).

https://user-images.githubusercontent.com/505788/214195166-2dad12a0-ba49-4dc2-8f8b-692969b7ef35.mov

I'm not entirely sure what the path forward would be. I do think that the expected behaviour for the Preview window would be to receive the 404 page, but it's not ideal that the CMS edit form also loads the 404 page.

tractorcow:
We could have a message to localise the page in draft before you can preview it?

chrispenny:
It's maybe just a bit tricky because the localisation interface is in that page (which you can't get to). Yea they could switch to a locale that has content, but which localised have a localisation might not even be immediately obvious to authors.

I don't suppose there's an easy way to force the CMS edit form to load with preview closed? Like a GET param or something that we could attach to the edit link if we see the page is unlocalised?

tractorcow:
middleware for 404 that detects draft querystring, and renders the message ?

[...]

PR #781, opened 2023-02-17
FIX Don't try to preview unlocalised objects

GuySartorelli:
I don't think there's ever a scenario where you'd _want_ to consider previews as not frontend so I don't think this is a breaking change.
I'm on the fence about whether `updatePreviewLink()` belongs here or in `FluentExtension` directly. _Probably_ it belongs on `FluentExtension` so it can affect anything else which have a preview (e.g. elemental blocks which aren't inline-editable) - but I'll leave it like this for now (it is still an improvement and solves the specific exact scenario mentioned in the issue) until someone opines.

## Parent issue
- https://github.com/tractorcow-farm/silverstripe-fluent/issues/775

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -1137,6 +1137,22 @@ class FluentExtension extends DataExtension
     }
 
     /**
+     * Update preview link to null if the object isn't in the current locale
+     * and we can't fallback cleanly.
+     *
+     * @param ?string $link
+     */
+    public function updatePreviewLink(&$link): void
+    {
+        $owner = $this->owner;
+        $info = $owner->LocaleInformation(FluentState::singleton()->getLocale());
+
+        if (!$info->getSourceLocale()) {
+            $link = null;
+        }
+    }
+
+    /**
      * Require that this record is saved in the given locale for it to be visible
      *
      * @return string
--- a/src/Middleware/InitStateMiddleware.php
+++ b/src/Middleware/InitStateMiddleware.php
@@ -70,11 +70,6 @@ class InitStateMiddleware implements HTTPMiddleware
             }
         }
 
-        // If using the CMS preview, do not treat the site as frontend
-        if ($request->getVar('CMSPreview')) {
-            return false;
-        }
-
         return true;
     }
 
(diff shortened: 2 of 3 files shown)


GuySartorelli:
Also happy to change the commit name - when I wrote it I was thinking of the _last_ piece of the chain that I was solving, not the first part as described in the issue itself.

tractorcow:
The only risk this introduced, as we identified in a prior conversation, is that some pages visible in the admin won't appear at all (i.e. 404) in preview. I still this this fix is a more consistent and better functionality, rather than showing a false view of content that then breaks after publish. :)

GuySartorelli:
@tractorcow I'm not sure if your comment is in response to something I've said, or to the changes I've introduced with this PR.... if the latter, then there _is no 404 in the preview (for pages specifically)_ with this PR. That's what setting the preview url to `null` ensures. Instead, the preview panel is collapsed by default for those pages and a "there is no preview available" message is displayed if the user explicitly swaps to split or preview mode.

tractorcow (approved):

GuySartorelli:
@tractorcow @chrispenny do either of you have an opinion as to whether this change belongs on the `FluentSiteTreeExtension` or directly on `FluentExtension`?

chrispenny:
@GuySartorelli Good question... Can any DataObject implement a Preview window? I don't think I've ever seen it outside of SiteTree, but I vaguely remember reading that it was possible.

If any DO can use Preview, then I guess it makes sense to be on `FluentExtension`.

CC: @mfendeksilverstripe

mfendeksilverstripe:
I think the ability to have a preview is not limited to only `SiteTree`. I think this belongs more to the FluentExtension. Alternatively, this can be a separate extension which is enabled by default but if needed can be disabled / overwritten via yaml config. This could give projects more flexibility over how to use this feature.

GuySartorelli:
> Can any DataObject implement a Preview window?

Yup. Any `DataObject` that implements the `CMSPreview` interface and returns some valid link path from `PreviewLink()` will have a preview available in gridfield edit forms and modeladmins. You can even slap a preview in the siteconfig if you really want.

GuySartorelli:
Oops I broke some tests - will look at that now

mfendeksilverstripe (approved):
Nice work 👍

PR #782, opened 2023-02-17
API Deprecate API removed in fluent 5

GuySartorelli:
Also updated to use the new gha-ci instead of the old experimental one

Because we're using `Deprecation::withNoReplacement()` this cannot be used with framework < 4.13

## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/666

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -1,5 +1,0 @@
-<?php
-
-use SilverStripe\Dev\Deprecation;
-
-Deprecation::notification_version('4.0.0', 'tractorcow/silverstripe-fluent');
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
     "require": {
         "php": "^7.3 || ^8.0",
         "silverstripe/vendor-plugin": "^1",
-        "silverstripe/framework": "^4.10",
+        "silverstripe/framework": "^4",
         "silverstripe/cms": "^4",
         "symbiote/silverstripe-gridfieldextensions": "^3.1"
     },
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -8,6 +8,7 @@ use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Extension;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\i18n\i18n;
 use SilverStripe\ORM\ArrayList;
@@ -826,10 +827,10 @@ class FluentExtension extends DataExtension
         $locale = $localeObj->getLocale();
 
         // Check linking mode
-        $linkingMode = $this->getLinkingMode($locale);
+        $linkingMode = $this->getLinkingModeValue($locale);
 
         // Check link
-        $link = $this->LocaleLink($locale);
+        $link = $this->getLocaleLinkValue($locale);
 
         // Store basic locale information
         return ArrayData::create([
@@ -852,6 +853,12 @@ class FluentExtension extends DataExtension
      * @return string
      */
     public function getLinkingMode($locale)
+    {
+        Deprecation::notice('4.13.0', 'Use LocaleInformation() instead');
+        return $this->getLinkingModeValue($locale);
+    }
+
+    private function getLinkingModeValue($locale)
     {
         if ($this->owner->hasMethod('canViewInLocale') && !$this->owner->canViewInLocale($locale)) {
             return 'invalid';
@@ -884,6 +891,12 @@ class FluentExtension extends DataExtension
      * @return string
      */
     public function LocaleLink($locale)
+    {
+        Deprecation::notice('4.13.0', 'Use LocaleInformation() instead');
+        return $this->getLocaleLinkValue($locale);
+    }
+
+    private function getLocaleLinkValue($locale)
     {
         // Skip dataobjects that do not have the Link method
         if (!$this->owner->hasMethod('Link')) {
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,9 +3,9 @@ name: Module CI
 on:
   push:
   pull_request:
+  workflow_dispatch:
 
 jobs:
   ci:
-    uses: silverstripe/github-actions-ci-cd/.github/workflows/ci.yml@v0.1
-    with:
-      run_phplinting: false
+    name: CI
+    uses: silverstripe/gha-ci/.github/workflows/ci.yml@v1
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -2,6 +2,9 @@
 <ruleset name="SilverStripe">
     <description>CodeSniffer ruleset for SilverStripe coding conventions.</description>
 
+	<file>src</file>
+	<file>tests</file>
+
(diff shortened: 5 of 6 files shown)


_config.php
```diff
@@ -1,5 +0,0 @@
-<?php
-
-use SilverStripe\Dev\Deprecation;
-
-Deprecation::notification_version('4.0.0', 'tractorcow/silverstripe-fluent');
```
GuySartorelli, line 5: Deprecation notices don't use this property anymore

src/Extension/FluentExtension.php
```diff
@@ -611,9 +612,13 @@ protected function localiseManipulationTable(&$manipulation, $table, $localeTabl
      *
      * @param SQLSelect $query
      * @param DataQuery $dataQuery
+     * @deprecated 4.8.0 Will be removed without equivalent functionality
      */
     public function augmentDataQueryCreation(SQLSelect $query, DataQuery $dataQuery)
     {
+        Deprecation::withNoReplacement(function () {
+            Deprecation::notice('4.8.0', 'Will be removed without equivalent functionality');
+        });
```
GuySartorelli, line 621: This needs to be wrapped with `Deprecation::withNoReplacement()` because this method will be called by `DataQuery` and we don't want to (by default) output notices about things people can't control in their own projects.

tractorcow: are we keeping data query params or are they going away?

GuySartorelli: All of those parameters are still valid - but this method is deprecated _for this module specifically_ for this extension because the method isn't in fluent 5. The actual hook is still valid in CMS 5, it's just not in use _for this module specifically_ from fluent 5 onwards.

tractorcow: Oh, it was removed at some point. I don't think I noticed it lol.

tractorcow: It's in fluentobjecttrait :) it's the mystery moving code,

composer.json
```diff
@@ -32,7 +32,7 @@
     "require": {
         "php": "^7.3 || ^8.0",
         "silverstripe/vendor-plugin": "^1",
-        "silverstripe/framework": "^4.10",
+        "silverstripe/framework": "^4.13",
```
tractorcow, line 35: This is going to cause problems for people pinned to older versions of both fluent and framework. fluent 4 is very old at this point (most people are using 5 or 6). 

When there were risky updates (although strictly non-breaking changes) I took the conservative approach and just released a new major fluent version (hence fluent 5 and 6 are both silverstripe 4).

Yes to this change, however, on the 6 branch and above, since it's the current "active" feature branch, and users can realistically expect more active development on these branches.

I propose we leave fluent 4 as is. I also regret updating 4 to require 4.10. I think 5.x is still pinned at ^4 framework.

src/Extension/FluentExtension.php
GuySartorelli: Oohhhhh nevermind then I'll undeprecate that one.

tractorcow:
@GuySartorelli as per my comment above, I propose we stop updating 4 and 5 branches with any more dependency bumps for the time being, and we focus on 6 and 7 for now. If you agree we will close this PR, unless you have an alternative. :)

composer.json
GuySartorelli: > This is going to cause problems for people pinned to older versions of both fluent and framework. fluent 4 is very old at this point (most people are using 5 or 6).

The whole point of this is for people who are doing upgrades. For anyone _not_ upgrading to a more recent version of silverstripe/framework, they have no need of tractorcow/silverstripe-fluent 4.8 anyway. When they run `composer update tractorcow/silverstripe-fluent` this version won't even be presented to them as an option, so it won't cause any problems. I can't see any way in which this will cause problems for anyone, given composer has really good dependency resolution.

maxime-rainville:
@tractorcow Just a bit of a wider context. Our official guidance for people upgrading will be to get their CMS project upgraded to the latest release of all the CMS 4 module and turn on deprecation warnings. 

Right now, even framework 4.10 is out of support. By the end of this year, only framework 4.13 will be supported. We've generally not gone out of our way to make it easy for project to run older version of framework with newer supported module. I'm 

If you think it's worthwhile allowing older version of framework to install the latest version of Fluent, we could do an existence check for the `withNoReplacement` method before calling it. So people could still choose to install Fluent with framework 4.10 if they want and get the deprecation warning if they are running framework 4.13.

@GuySartorelli I see that there's some methods that are throwing deprecation warnings in Fluent 5 that you haven't touch in this PR. I'm wondering if it would be better getting those methods to emit deprecation warning in Fluent 4 as well?

GuySartorelli:
> I see that there's some methods that are throwing deprecation warnings in Fluent 5 that you haven't touch in this PR. I'm wondering if it would be better getting those methods to emit deprecation warning in Fluent 4 as well?

Could do. The message would have to be "Will be replaced with x" instead of "Use x instead" since the replacements aren't actually available in 4. I guess they _are_ technically deprecated though since they're removed in _a future major release_ even if that isn't the release immediately following 4.

Or, in short: Yes that would be sensible, I'll do that (tomorrow).

tractorcow:
> If you think it's worthwhile allowing older version of framework to install the latest version of Fluent, we could do an existence check for the withNoReplacement method before calling it. So people could still choose to install Fluent with framework 4.10 if they want and get the deprecation warning if they are running framework 4.13.

Sorry, let me clarify. I do agree with your point. The current version of fluent should work on the latest version of framework. I have no disagreement there.

The problem is that this PR is against the version 4 of fluent, that's a very very old branch of fluent. There's no obligation for either of us to support an outdated version of fluent (or framework for that matter).

The point I'm making is that upgrading work should be done on newer branches, and we should not waste effort trying to back-port changes to branches that aren't actively developed anymore.

GuySartorelli:
> There's no obligation for either of us to support an outdated version of fluent

Unfortunately the only version of fluent we've had in recipes in version 4, which means that's the version that was commercially supported with CMS 4.

tractorcow:
>  This pr is purely to add deprecation notices, so the upgrade process is easier for anyone who's still using v4 of fluent.

Yeah but the problem is that users on old versions of fluent and framework won't get deprecation notices. they'll get crashes.

> Because we're using Deprecation::withNoReplacement() this cannot be used with framework < 4.13

If you can do this another way without using new API I can accept. Just use the old Deprecation::notice() api and call it a day?

tractorcow:
I'll be fair and try to meet you half way. :) https://github.com/tractorcow-farm/silverstripe-fluent/pull/784

GuySartorelli:
I've updated this PR with the changes from #784, and applied the changes recommended by Max in https://github.com/tractorcow-farm/silverstripe-fluent/pull/782#issuecomment-1437829335

@tractorcow Adding the `private static` for `cms_publish_required` is necessary for the "correct" way to deprecate config - but I understand that's a departure from how those config vars are added in this module so I'd be happy to remove that change if you find it a bit strange.

tractorcow (approved):

tractorcow:
Happy to merge if you are.

Do you want this merged up, or do you want to do 5, 6 etc... separately?

GuySartorelli:
Best to not merge up - resolving the merge conflicts would literally just be removing these changes anyway since this API is removed in those majors 😅

PR #783, opened 2023-02-17
API Standardise deprecations

GuySartorelli:
Standardises deprecation notices to be the same as other supported modules

## NOTE
After merging this PR, it should be merged up to or cherry-picked into the `6` branch.

## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/666

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -1,5 +1,0 @@
-<?php
-
-use SilverStripe\Dev\Deprecation;
-
-Deprecation::notification_version('4.0.0', 'tractorcow/silverstripe-fluent');
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -923,11 +923,11 @@ class FluentExtension extends DataExtension
      *
      * @param string $locale
      * @return string
-     * @deprecated 5.0 use LocaleInformation() instead
+     * @deprecated 5.0.0 Use LocaleInformation() instead
      */
     public function getLinkingMode($locale)
     {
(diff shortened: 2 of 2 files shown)


_config.php
```diff
@@ -1,5 +0,0 @@
-<?php
-
-use SilverStripe\Dev\Deprecation;
-
-Deprecation::notification_version('4.0.0', 'tractorcow/silverstripe-fluent');
```
GuySartorelli, line 5: Deprecation notices don't get filtered by version number anymore, so this isn't doing us any good being here.

src/Extension/FluentExtension.php
```diff
@@ -1021,7 +1021,11 @@ protected function requireSavedInLocale()
         }
 
         if ($this->owner->config()->get('cms_publish_required') !== null) {
-            Deprecation::notice('5.0', 'Use cms_localisation_required instead');
+            Deprecation::notice(
+                '5.0.0',
+                'The "' . self::class . '.cms_publish_required" configuration option is deprecated. Use "'
+                    . self::class . '.cms_localisation_required instead'
+            );
```
GuySartorelli, line 1028: The original message implies we should use something called `cms_localisation_required` instead of the `requireSavedInLocale()` method, which is misleading.

The correct way to deprecate config is to add a `@deprecated` annotation to the phpdoc for the `private static` config variable... but since this module doesn't declare those, this will probably be fine.

tractorcow (approved):

tractorcow:
Let me know when you are happy and I can merge.

PR #784, opened 2023-02-21
 API Deprecate API removed in fluent 5 (suggested)

tractorcow:
Suggested replacement PR for https://github.com/tractorcow-farm/silverstripe-fluent/pull/782.

My motivation is to allow a high degree of support for users of new framework 4.x (e.g. 4.12/4.13) that are using fluent 4 / 5 opting to use soft-support for api that may or may not be present.

For versions of fluent that require new framework features (e.g. versioned-admin, new deprecations, php version bump) I will be happy to support dependency changes on the 6 branch, which will be the last feature branch for framework 4.x dependencies.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,9 +3,9 @@ name: Module CI
 on:
   push:
   pull_request:
+  workflow_dispatch:
 
 jobs:
   ci:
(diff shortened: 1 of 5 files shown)


tractorcow:
Superseded by https://github.com/tractorcow-farm/silverstripe-fluent/pull/782

Issue #780, opened 2023-02-16
Bug: CMS Language selector disappearing


bumbus:
There seems to be a bug with the language selector in the CMS, after certain actions it just disappears. This happened after Updating to SilverStripe 4.12. It seems to affect both Fluent versions 5 and 6.
For now I have not figured out what exact actions trigger the dissapearance.

https://user-images.githubusercontent.com/1391103/219429726-32d445ca-b69e-44ec-afe2-f4f0ab200efe.mov

kevingroeger:
Same issue here, happening since I upgraded to silverstripe 4.12

tractorcow:
This is the CSS selector which is having the fluent locale selector appended to it.

```
$('.cms > .cms-container > .cms-menu > .cms-panel-content').entwine({
```

If this has changed the selector might break.

kinglozzer:
I’m also seeing this in 5.0.0-beta3 and Fluent 7.0.0-beta1, when the dropdown disappears the class names in the selector mentioned above are all there so everything looks to be correct... I’m also having real trouble nailing steps to reproduce though!

kinglozzer:
Quite possibly related to this: https://docs.silverstripe.org/en/4/changelogs/4.12.0/#jquery-entwine-change

jules0x:
Occurring for me as well.

composer.lock
```
"name": "silverstripe/framework",
"version": "4.12.3",
"name": "tractorcow/silverstripe-fluent",
"version": "5.1.1",
```

RVXD:
Hope this can be fixed. Seems to be a timing issue.

jules0x:
Thanks @kinglozzer a `composer update` pulls **silverstripe/admin** 1.12.5 and your fix has resolved the issue.

GuySartorelli:
No longer a problem

PR #788, opened 2023-03-22
ensure config is ready before loading menu

blueo:
This is a small change to allow wait for the CMS config to be loaded before running the onmatch to add the local menu to the side bar.

We found that the fluent JS was sometimes added before the CMS config script and in some cases this would mean it ran too early and would never add the Menu.


blueo:
Turns out there is a better fix in #780 for silverstripe admin so closing this one :)

PR #799, opened 2023-05-31
Update FluentExtension.php

mikey-harveycameron:
skip empty 'where' clauses.
just trying to make this available for SS4.

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -583,6 +583,10 @@ class FluentExtension extends DataExtension
             // Apply substitutions
             $localisedPredicate = str_replace($conditionSearch, $conditionReplace, $predicate);
 
+            if (empty($localisedPredicate)) {
+                continue;
+            }
+            
             $where[$index] = [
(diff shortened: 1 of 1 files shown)


tractorcow:
Do you know when this value is ever blank? Is there another error that we are ignoring in this case? further upstream? My worry is that there's a condition we are handling wrong, and just ignoring it instead.

tractorcow:
So this could occur if you had a condition like

```php
$query->where(['' => 3]); // Not a valid condition, but is accepted anyway
```

tractorcow (approved):

tractorcow:
No harm in the merge, but it'd be good to track down the cause please if you can add some more context in your own time. :)

mikey-harveycameron:
Could you update this branch with a new tag or something please, so that composer update can reach it... maybe like 6.0.1? (Not quite sure how this works, my apologies.)

mikey-harveycameron:
FYI I was able to target it with composer using "6.x-dev"

PR #805, opened 2023-08-09
MNT Run module-standardiser

emteknetnz:
This pull-request was created automatically by [module-standardiser](https://github.com/silverstripe/module-standardiser)


tractorcow:
I have some questions about silverstripe/gha-merge-up.  Is there some human readable description of what https://github.com/silverstripe/gha-merge-up does exactly?

tractorcow:
This isn't going to go over all the old branches from ~10 years ago and try to merge these up is it?

tractorcow (changes requested):
Temporarily blocking merge until we can confirm what this new process will do.

emteknetnz:
gha-merge-up will attempt to automatically merge-up from the latest minor branch on the previous supported major (in the case of fluent that will be 6) all the way through to the next-minor branch on the current major.  It runs on a weekly cron.  It's pretty conservative and will fail the github action if there are merge-conflicts, unless they are in client/dist in which case it will run yarn build

In the case of silverstripe/framework it will currently merge-up 4.13 -> 4 -> 5.0 -> 5

You can have a look to see what the job looks like in action [here](https://github.com/silverstripe/silverstripe-framework/actions/runs/5815166792)

> This isn't going to go over all the old branches from ~10 years ago and try to merge these up is it?

No it won't.

I'm not sure what the scope of differences are between fluent 6 and 7.  If you think this is unrealistic or just a bad idea for fluent then we would add an exception to gha-merge-up to only merge-up on the most recent supported major i.e. currently only merge-up from 7.0 -> 7

Does this new process seem OK to you?

tractorcow:
Lets give it a go and I'll keep an eye on it :)

GuySartorelli:
IMPORTANT: `7` isn't the default branch for this module, so this new workflow won't run unless:
1. the default branch is swapped to be `7`, OR
2. this change is merged up to the current default branch
@tractorcow  FYI

PR #808, opened 2023-08-29
MNT Run module-standardiser

emteknetnz:
This pull-request was created automatically by [module-standardiser](https://github.com/silverstripe/module-standardiser)


emteknetnz:
This is functionally the same PR as https://github.com/tractorcow-farm/silverstripe-fluent/pull/805 - it's just it targets `7.0` rather than `7`.  We found out that in practice we should have targetted the "next-patch" branches rather than the "next-minor" branches in the first place as some of the github workflows / actions we run need to have workflow's on the next-patch branches.

tractorcow:
Sure sounds good.

PR #812, opened 2023-10-17
MSSQL database doesn't allow aliases in GROUP BY

t3hn0:
This part of code is preventing SS5 to be run on MSSQL database.

Fix for #813


tractorcow:
You are correct, the original code was in bad style anyway. good on mssql for having high standards :)

tractorcow (approved):

PR #824, opened 2024-02-08
FIX allows i18nTextCollector to concatenate keys

### LINKED ISSUE #823: TextCollector Task throws many Warnings
### Module version(s) affected

5,6,7

### Description

The output of the i18nTextCollectorTask returns a number of warnings coming from this module. All are of two types:

```
[User Warning] Error concatenating localisation key
GET /dev/tasks/i18nTextCollectorTask?

Line 737 in /var/www/html/vendor/silverstripe/framework/src/i18n/TextCollection/i18nTextCollector.php
```
```
[User Notice] Missing localisation default for key Localisation
GET /dev/tasks/i18nTextCollectorTask

Line 815 in /var/www/html/vendor/silverstripe/framework/src/i18n/TextCollection/i18nTextCollector.php
```

Each instance of the second appears to be a direct result of the first. 

### How to reproduce

Run `/dev/tasks/i18nTextCollectorTask` with the environment set to `dev` or error reporting otherwise set to display PHP warnings.

### Possible Solution

The issue appears to be due to the use of the `__TRAIT__` magic constant w/in `_t()` function calls in FluentAdminTrait, FluentBadgeTrait and FluentObjectTrait. The backslashes in the fully namespaced trait name cannot be parsed.

It should be possible to resolve by replacing all instances of `__TRAIT__` with the string name of the class, or a function call which strips the namespace from the result.

### Additional Context

_No response_

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

davejtoews:
## Description

fixes #823 - The magic constant __TRAIT__ was inserting namespaces into localization keys, containing slashes that could not be parsed by the TextCollector task. These have been replaced w/ the Trait name as a string

## Manual testing steps
Run `/dev/tasks/i18nTextCollectorTask` with the environment set to Dev or error reporting otherwise configured to display warnings.

## Issues
#823

### DIFF
--- a/src/Extension/Traits/FluentAdminTrait.php
+++ b/src/Extension/Traits/FluentAdminTrait.php
@@ -107,7 +107,7 @@ trait FluentAdminTrait
         // Add menu button
         $moreOptions = Tab::create(
             'FluentMenuOptions',
-            _t(__TRAIT__ . '.Localisation', 'Localisation')
+            _t('FluentAdminTrait.Localisation', 'Localisation')
         );
         $moreOptions->addExtraClass('popover-actions-simulate');
         $rootTabSet->push($moreOptions);
@@ -117,7 +117,7 @@ trait FluentAdminTrait
             FormAction::create(
                 'clearFluent',
                 _t(
-                    __TRAIT__ . '.Label_clearFluent',
+                    'FluentAdminTrait.Label_clearFluent',
                     "Clear from all except '{title}'",
                     [
                         'title' => $locale->getTitle()
@@ -129,7 +129,7 @@ trait FluentAdminTrait
             FormAction::create(
                 'copyFluent',
                 _t(
-                    __TRAIT__ . '.Label_copyFluent',
+                    'FluentAdminTrait.Label_copyFluent',
                     "Copy '{title}' to other locales",
                     [
                         'title' => $locale->getTitle()
@@ -144,7 +144,7 @@ trait FluentAdminTrait
                 FormAction::create(
                     'unpublishFluent',
                     _t(
-                        __TRAIT__ . '.Label_unpublishFluent',
+                        'FluentAdminTrait.Label_unpublishFluent',
                         'Unpublish (all locales)'
                     )
                 )->addExtraClass('btn-secondary')
@@ -153,7 +153,7 @@ trait FluentAdminTrait
                 FormAction::create(
                     'archiveFluent',
                     _t(
-                        __TRAIT__ . '.Label_archiveFluent',
+                        'FluentAdminTrait.Label_archiveFluent',
                         'Unpublish and Archive (all locales)'
                     )
                 )->addExtraClass('btn-outline-danger')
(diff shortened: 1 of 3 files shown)


davejtoews:
I wasn't 100% sure which branch to submit a PR to patch version 6. Let me know if this belongs on another branch.

tractorcow:
That's not actually correct, textcollector has always been able to support namespaces with slashes in them. Here is the currently extracted strings.

Are you running an older or outdated version of text collector?


```yaml
  TractorCow\Fluent\Extension\Traits\FluentAdminTrait:
    ClearAllNotice: 'All localisations have been cleared for ''{title}''.'
    CopyNotice: 'Copied ''{title}'' to all other locales.'
    DeleteNotice: 'Deleted ''{title}'' and all of its localisations.'
    UnpublishNotice: 'Unpublished ''{title}'' from all locales.'
    ArchiveNotice: 'Archived ''{title}'' and all of its localisations.'
    PublishNotice: 'Published ''{title}'' across all locales.'
```

tractorcow (changes requested):
To investigate possible regression in text collector

tractorcow:
It could also matter maybe what PHP version you're testing it on.

davejtoews:
Yeah, I after a bit more digging it does appear to not be about the namespaces but something about the __TRAIT__ magic constant itself. Passing the slashes in as a string works, but not w/ __TRAIT__

I'm running Silverstripe Framework 4.13.29 on PHP 8.1

tractorcow:
Ok, we should be fine to replace __TRAIT__ with the literal, but we probably should keep the slashes otherwise all the existing localisations will need to be migrated. :) Hope that's a reasonable resolution.

davejtoews:
Yes, I was coming to the conclusion that was the correct solution already but hadn't yet updated the PR. Thanks.

davejtoews:
Strings updated to include full namespace. Presumably this change needs doing on Fluent v7 as well but I am not running Silverstripe 5 and haven't tested it.

GuySartorelli:
I honestly think this is being addressed in the wrong place - the textcollector should instead be updated to correctly handle `__TRAIT__`.

tractorcow (approved):

tractorcow:
Can be merged when tests pass.

davejtoews:
@GuySartorelli It looks like this may have been fixed in framework 5 but not applied to 4? https://github.com/silverstripe/silverstripe-framework/commit/f4e0c768bdeeb3c2a379320eb219a6020c80ce08

GuySartorelli:
@davejtoews oh, nice find! `__TRAIT__` is *well* within our support commitments for CMS 4 so if you want to cherry pick that on top of framework `4.13` and raise a PR that'll be an easy merge.

PR #830, opened 2024-03-20
ENH Add ability to ignore missing fields

lekoala:
## Description

SHOW FULL FIELDS are costly queries. Maybe it's not necessary to do this on every request

## Manual testing steps

Using debugbar, you can check the queries
With this new feature, this prevents all queries: SHOW FULL FIELDS IN xxx

## Issues
https://github.com/tractorcow-farm/silverstripe-fluent/issues/828

## Pull request checklist

- [ ] The target branch is correct
    - See [picking the right version](https://docs.silverstripe.org/en/contributing/code/#picking-the-right-version)
- [ ] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
    - Small amounts of additional linting are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [ ] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [ ] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [ ] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [ ] This change is covered with tests (or tests aren't necessary for this change)
- [ ] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [ ] CI is green

PR #831, opened 2024-03-20
fix: don't invoke db for ready check outside of dev urls

### LINKED ISSUE #828: Middleware is causing extra queries
### Module version(s) affected

7.0.0

### Description

Here is a debugbar screenshot

image

We see that we have two SHOW FULL FIELDS. One from the Domain class (even if you have a single domain and not using domain mode) and one for the Locale table (due to the databaseIsReady check). 

In many cases, this should not be made PER REQUEST. It could be useful to avoid errors in case the database is not available, but it should not lead to extra queries being applied for every request.

Unused features should not cause extra queries either.

### How to reproduce

Install the module
Install the debugbar
Check that indeed we have these queries

### Possible Solution

- Have a env var to disable domain feature if not used and avoid extra queries
- Allow forcing locale per dev config, instead of db config
- Take an optimistic approach for db availability and avoid extra checks per requests

### Additional Context

_No response_

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

tractorcow:
fixes #828

This maintains the check for DB, but it only invokes it during dev/build, which is when this check is necessary. Otherwise we assume the DB is ready for day to day use.

Replaces https://github.com/tractorcow-farm/silverstripe-fluent/pull/830

### DIFF
--- a/src/Model/CachableModel.php
+++ b/src/Model/CachableModel.php
@@ -2,6 +2,8 @@
 
 namespace TractorCow\Fluent\Model;
 
+use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injector;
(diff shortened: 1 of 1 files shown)


tractorcow:
Fixed and merged up to other branchesl

tractorcow:
Well, I broke all the tests. Will take responsibility and fix soon, sorry team.

PR #832, opened 2024-03-24
fix: check for request prior to access

tractorcow:
fix: check for request prior to access


tractorcow:
Looks like tests didnt run on 5.1 branch so I didn't get the fix across.

PR #838, opened 2024-04-04
ENH: Localisation manager links available for CMS edit link capable models.

mfendeksilverstripe:
## Description

Allow all CMS edit link capable models to use a Title link in the Locale manager. This link is quite handy as you can switch between locales easily. Previously this was available only for models extending `SiteTree` (pages).

Screenshot 2024-04-05 at 11 35 15 AM

## Pull request checklist

- [x] CI is green

### DIFF
--- a/src/Extension/Traits/FluentObjectTrait.php
+++ b/src/Extension/Traits/FluentObjectTrait.php
@@ -2,6 +2,10 @@
 
 namespace TractorCow\Fluent\Extension\Traits;
 
+use SilverStripe\Admin\CMSEditLinkExtension;
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
+use SilverStripe\Core\Convert;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\GridField\GridFieldConfig;
@@ -10,6 +14,7 @@ use SilverStripe\Forms\GridField\GridFieldDataColumns;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataQuery;
+use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\Queries\SQLSelect;
 use TractorCow\Fluent\Model\Locale;
 use TractorCow\Fluent\State\FluentState;
@@ -88,7 +93,7 @@ trait FluentObjectTrait
      */
     protected function updateFluentCMSFields(FieldList $fields)
     {
-        /** @var DataObject $owner */
+        /** @var DataObject|CMSEditLinkExtension $owner */
         $owner = $this->owner;
         if (!$owner->ID) {
             return;
@@ -107,6 +112,37 @@ trait FluentObjectTrait
             'Title' => 'Title',
             'Locale' => 'Locale'
         ];
+
+        // Augment Localisation tab with clickable locale links to allow easy navigation between model localisations
+        if ($owner->hasExtension(CMSEditLinkExtension::class)) {
+            $controller = Controller::has_curr() ? Controller::curr() : null;
+            $request = $controller?->getRequest();
+
+            // Pass getVars separately so we can process them later
+            $params = $request?->getVars() ?? [];
+
+            // This is to get URL only, getVars are not part of the URL
+            $url = $owner->CMSEditLink();
+            $url = Director::makeRelative($url);
+
+            $summaryColumns['Title'] = [
+                'title' => 'Title',
+                'callback' => function (Locale $object) use ($url, $params): ?DBField {
+                    if (!$object->RecordLocale()) {
+                        return null;
+                    }
+
+                    $recordLocale = $object->RecordLocale();
+                    $locale = $recordLocale->getLocale();
+                    $params['l'] = $locale;
+                    $localeLink = Controller::join_links($url, '?' . http_build_query($params));
+                    $localeTitle = Convert::raw2xml($recordLocale->getTitle());
+                    $render = sprintf('<a href="%s" target="_top">%s</a>', $localeLink, $localeTitle);
+
+                    return DBField::create_field('HTMLVarchar', $render);
+                }
+            ];
+        }
 
         // Let extensions override columns
         $owner->extend('updateLocalisationTabColumns', $summaryColumns);
--- a/src/Extension/Traits/FluentObjectTrait.php
+++ b/src/Extension/Traits/FluentObjectTrait.php
@@ -125,23 +125,25 @@ trait FluentObjectTrait
             $url = $owner->CMSEditLink();
             $url = Director::makeRelative($url);
 
-            $summaryColumns['Title'] = [
-                'title' => 'Title',
-                'callback' => function (Locale $object) use ($url, $params): ?DBField {
-                    if (!$object->RecordLocale()) {
-                        return null;
+            if ($url) {
+                $summaryColumns['Title'] = [
+                    'title' => 'Title',
+                    'callback' => function (Locale $object) use ($url, $params): ?DBField {
+                        if (!$object->RecordLocale()) {
+                            return null;
+                        }
+
+                        $recordLocale = $object->RecordLocale();
+                        $locale = $recordLocale->getLocale();
+                        $params['l'] = $locale;
+                        $localeLink = Controller::join_links($url, '?' . http_build_query($params));
+                        $localeTitle = Convert::raw2xml($recordLocale->getTitle());
+                        $render = sprintf('<a href="%s" target="_top">%s</a>', $localeLink, $localeTitle);
+
+                        return DBField::create_field('HTMLVarchar', $render);
                     }
-
-                    $recordLocale = $object->RecordLocale();
-                    $locale = $recordLocale->getLocale();
-                    $params['l'] = $locale;
-                    $localeLink = Controller::join_links($url, '?' . http_build_query($params));
-                    $localeTitle = Convert::raw2xml($recordLocale->getTitle());
-                    $render = sprintf('<a href="%s" target="_top">%s</a>', $localeLink, $localeTitle);
-
-                    return DBField::create_field('HTMLVarchar', $render);
-                }
-            ];
+                ];
+            }
         }
 
         // Let extensions override columns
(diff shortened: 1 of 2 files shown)


GuySartorelli:
Please change the commit prefix to `FIX` as per the [commit message guidelines](https://docs.silverstripe.org/en/5/contributing/code/#commit-messages) - I've unticked that box in the checklist, feel free to tick it again when you've made this change.

As this is a bug fix that doesn't introduce new API, please also retarget this to the `7.0` branch so it can be released as a patch. I've unticked the relevant checkbox for this in the checklist as well.

Note you may need to [reset your commits](https://gist.github.com/GuySartorelli/74627843aeb75e12b777b1f44cb22d2f) after retargetting the PR.

mfendeksilverstripe:
I've pushed up the requested changes @GuySartorelli , please review.

GuySartorelli:
There are failing unit tests to address.
It looks like you changed the base **after** pushing the commit, so you may need to force push your changes to get CI to rerun against the correct parent branch.

mfendeksilverstripe:
Tests passing now @GuySartorelli , please review 🙏 .

src/Extension/Traits/FluentObjectTrait.php
```diff
@@ -109,12 +115,48 @@ protected function updateFluentCMSFields(FieldList $fields)
...
+                'callback' => function (Locale $object) use ($url, $params) {
+                    if (!$object->RecordLocale()) {
+                        return null;
+                    }
+
+                    $recordLocale = $object->RecordLocale();
+                    $locale = $recordLocale->getLocale();
+                    $params['l'] = $locale;
+                    $localeLink = Controller::join_links($url, '?' . http_build_query($params));
+                    $localeTitle = Convert::raw2xml($recordLocale->getTitle());
```
maxime-rainville: Presumably this is there to sure that someone couldn't put a XSS payload in the Locale Title.

Are we happy with the assumption that there's no XSS vector in the locale link?

src/Extension/Traits/FluentObjectTrait.php
```diff
@@ -109,12 +115,48 @@ protected function updateFluentCMSFields(FieldList $fields)
...
+                    $render = sprintf('<a href="%s" target="_top">%s</a>', $localeLink, $localeTitle);
```
maxime-rainville: Maybe we could put this in a SS template? I tend not to like putting HTML directly in PHP when we have an entire templating engine for this purpose.

Not a big deal.

maxime-rainville (commented):
Not a full review, just a few things that were jumping at me when I glance at it.

src/Extension/Traits/FluentObjectTrait.php
ChloeSartorelli: I feel like a whole template for a single HTML tag is overkill - especially since that means rendering it many times for any given gridfield.

src/Extension/Traits/FluentObjectTrait.php
mfendeksilverstripe: I would say that this is out of scope of this PR. Not really changing how the link is generated here.

ChloeSartorelli: Yup it's largely a copy/paste. The only real change for the purposes of what's rendered is that it's not being handled by `DBHTMLVarchar` anymore, and that class didn't do any sanitisation. This PR doesn't introduce any risks that weren't already there.

src/Extension/Traits/FluentObjectTrait.php
```diff
@@ -109,12 +115,48 @@ protected function updateFluentCMSFields(FieldList $fields)
...
+
+                    return $render;
```
ChloeSartorelli: I tried this with `return DBField::create_field('HTMLVarchar', $render);` and that works for me locally.
When I checked out the `7.0` branch I found that that works for me as well.

Did you replicate the original issue in a fresh installation?

mfendeksilverstripe: yup, tested this with vanilla install with this composer file (ran composer update just before the test): (replaced fluent folder with the one from my branch though)

```
{
    "name": "silverstripe/installer",
    "type": "silverstripe-recipe",
    "description": "The SilverStripe Framework Installer",
    "require": {
        "php": "^8",
        "silverstripe/recipe-plugin": "^2",
        "silverstripe/recipe-cms": "~5.x-dev",
        "silverstripe-themes/simple": "~3.2.0",
        "silverstripe/login-forms": "^5",
        "dnadesign/silverstripe-elemental": "^5",
        "maxime-rainville/anyfield": "^0.0.0",
        "silverstripe/tagfield": "^3.1",
        "silverstripe-terraformers/gridfield-rich-filter-header": "dev-master",
        "tractorcow/silverstripe-fluent": "^7"
    },
    "require-dev": {
        "silverstripe/recipe-testing": "^3",
        "silverstripe/graphql-devtools": "^1.0"
    },
    "extra": {
        "resources-dir": "_resources",
        "project-files-installed": [
            "app/.htaccess",
            "app/_config.php",
            "app/_config/mimevalidator.yml",
            "app/_config/mysite.yml",
            "app/src/Page.php",
            "app/src/PageController.php",
            "behat.yml",
            "phpcs.xml.dist",
            "phpunit.xml.dist"
        ],
        "public-files-installed": [
            ".htaccess",
            "index.php",
            "web.config"
        ]
    },
    "repositories": {
        "ichaber/silverstripe-swiftype": {
            "type": "git",
            "url": "https://github.com/silverstripe-terraformers/silverstripe-swiftype.git"
        }
    },
    "config": {
        "process-timeout": 600,
        "allow-plugins": {
            "composer/installers": true,
            "silverstripe/vendor-plugin": true,
            "silverstripe/recipe-plugin": true
        }
    },
    "prefer-stable": true,
    "minimum-stability": "dev"
}

```

ChloeSartorelli: It's very important to note that that is **not** a vanilla installation - there are a lot of optional modules there which you won't get when you just do `composer create-project silverstripe/installer && composer require tractorcow/silverstripe-fluent`

With this set of modules I do experience the problem, but not when I just do a truly vanilla installation. I suspect one of the additional optional modules in this set is interfering with the way the value is displayed.

ChloeSartorelli: Aha! It's the constraint for `silverstripe/recipe-cms` here that changes things.

`"silverstripe/recipe-cms": "5.1.x-dev"` works as expected without this PR
`"silverstripe/recipe-cms": "~5.x-dev"` does not work as expected without this PR

Looks like there's a regression in framework or somewhere similar, and the problem should be addressed wherever that root cause regression lies.

mfendeksilverstripe: Excellent find @GuySartorelli ❤️ . I will change this PR into enhancement only covering the part that expands the display of title links which was my original intention. I'll check if I can find the source of the regression separately.

GuySartorelli:
re: https://github.com/tractorcow-farm/silverstripe-fluent/pull/838#discussion_r1555045471

I'll mark this as a draft in the meantime to avoid it being merged in its current state.

mfendeksilverstripe:
This is now ready to be reviewed as a standalone enhancement @GuySartorelli

GuySartorelli:
Please retarget this to the `7` branch since this is no longer fixing a bug

mfendeksilverstripe:
Created https://github.com/silverstripe/silverstripe-framework/issues/11191 to cover the malformed HTML

mfendeksilverstripe:
@GuySartorelli Updated base branch to `7` as requested. Please review.

src/Extension/Traits/FluentObjectTrait.php
```diff
@@ -108,6 +113,37 @@ protected function updateFluentCMSFields(FieldList $fields)
...
+                    $localeTitle = Convert::raw2xml($recordLocale->getTitle());
+                    $render = sprintf('<a href="%s" target="_top">%s</a>', $localeLink, $localeTitle);
+
+                    return DBField::create_field('HTMLVarchar', $render);
+                }
+            ];
+        }
+
         // Let extensions override columns
         $owner->extend('updateLocalisationTabColumns', $summaryColumns);
```
ChloeSartorelli, line 150: Having the link added before this hook does change how the value is when it's passed into the extension hook - however, because it was manipulated in the extension hook in the past, the value would be different depending on the order of extension hooks being called.

This is now more consistent, so IMO it's better than it was before in that regards.

src/Extension/Traits/FluentObjectTrait.php
```diff
@@ -108,6 +113,37 @@ protected function updateFluentCMSFields(FieldList $fields)
...
+        if ($owner->hasExtension(CMSEditLinkExtension::class)) {
+            $controller = Controller::has_curr() ? Controller::curr() : null;
+            $request = $controller?->getRequest();
+
+            // Pass getVars separately so we can process them later
+            $params = $request?->getVars() ?? [];
+
+            // This is to get URL only, getVars are not part of the URL
+            $url = $owner->CMSEditLink();
+            $url = Director::makeRelative($url);
```
ChloeSartorelli, line 126: There was some defensive programming in the old implementation against falsy `$url` which I think we should keep. If an empty URL is returned for any reason we shouldn't try to add a link.

src/Extension/Traits/FluentObjectTrait.php
```diff
@@ -108,6 +113,37 @@ protected function updateFluentCMSFields(FieldList $fields)
             'Locale' => 'Locale'
         ];
 
+        // Augment Localisation tab with clickable locale links to allow easy navigation between model localisations
+        if ($owner->hasExtension(CMSEditLinkExtension::class)) {
+            $controller = Controller::has_curr() ? Controller::curr() : null;
+            $request = $controller?->getRequest();
+
+            // Pass getVars separately so we can process them later
+            $params = $request?->getVars() ?? [];
```
ChloeSartorelli, line 122: This is different to the previous behaviour - but I think the previous behaviour was overly cautious. If there's no request or controller somehow (which is very unlikely for this) it makes sense to just have an empty params set here.

GuySartorelli (changes requested):

GuySartorelli:
Works well locally, just one suggested change above.

mfendeksilverstripe:
I've added the requested change @GuySartorelli , please review.

src/Extension/Traits/FluentObjectTrait.php
```diff
@@ -108,6 +113,39 @@ protected function updateFluentCMSFields(FieldList $fields)
...
+
+            // This is to get URL only, getVars are not part of the URL
+            $url = $owner->CMSEditLink();
+            $url = Director::makeRelative($url);
+
+            if ($url) {
```
ChloeSartorelli, line 128: I double checked and `Director::makeRelative()` returns an empty string if an empty string is passed in, so while this path is slightly different to how it used to be, it should be fine.

GuySartorelli (approved):
LGTM and works as expected locally.
Merge on green

PR #840, opened 2024-04-05
FIX: Source locale indicator correction.

mfendeksilverstripe:
# FIX: Source locale indicator correction.

## Description

Source locale detection corrected to take into account locale inheritance settings for a specific model. Impacted methods are only used in the CMS UI to give indication to content authors on what to expect (what is the source locale of the content). Before this change the indicator could be inaccurate in a sense that it would indicate content and there would actually not be any content to display.

319797975-24c993d8-8ac2-4704-8a61-694797fa6d64

This change should have no impact on actual content rendering.

**Change-set breakdown**

* `RecordLocale::getSourceLocale()` now takes into account locale inheritance mode
* `FluentExtension::getSourceLocale()` is now only a shorthand method to the above method instead of maintaining its own implementation that doesn't cover all the cases

## Manual testing steps

* assumes CMS 5 installer + Fluent
* navigate to the CMS under locales admin
* create first locale - English (Any english) and set this locale to be the global default
* create second locale - German and set the this locale to fall back to the first locale
* create a campaign page (code snippet below)

```php
class CampaignPage extends Page
{
    private static string $frontend_publish_required = FluentExtension::INHERITANCE_MODE_EXACT;
}
```

* run dev/build with flush
* reload the CMS admin pages and navigate to the pages section
* make sure you are in the English locale (see locale selector in the top left)
* create a campaign page (fill any data in)
* navigate to the localisation manager tab within the campaign page edit form
* you should see two records displayed (one for each locale)
* check the source locale for the German locale for this campaign page
* the expected value is "No source" (state after the change), before this change-set it would show English locale

Most common scenarios:

* Page uses fallback inheritance mode (default)
* Assets use any inheritance mode
* Campaign pages use exact inheritance mode

The source locale indicator should be in line with what is shown on the frontend.

## Issues

- https://github.com/tractorcow-farm/silverstripe-fluent/issues/842

## Pull request checklist
- [x] CI is green

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -25,7 +25,6 @@ use SilverStripe\ORM\FieldType\DBVarchar;
 use SilverStripe\ORM\Queries\SQLConditionGroup;
 use SilverStripe\ORM\Queries\SQLSelect;
 use SilverStripe\ORM\ValidationException;
-use SilverStripe\Security\Permission;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\View\HTML;
 use TractorCow\Fluent\Extension\Traits\FluentObjectTrait;
@@ -992,19 +991,24 @@ class FluentExtension extends DataExtension
         return Locale::getCurrentLocale();
     }
 
-
     /**
      * Returns the source locale that will display the content for this record
      *
      * @return Locale|null
      */
-    public function getSourceLocale()
+    public function getSourceLocale(): ?Locale
     {
-        $sourceLocale = $this->owner->getField('SourceLocale');
-        if ($sourceLocale) {
-            return Locale::getByLocale($sourceLocale);
+        $currentLocale = FluentState::singleton()->getLocale();
+
+        // We do not have any locales set up yet, so there is no source locale to find
+        if (!$currentLocale) {
+            return null;
         }
-        return Locale::getDefault();
+
+        $owner = $this->owner;
+        $localeInformation = $owner->LocaleInformation($currentLocale);
+
+        return $localeInformation->getSourceLocale();
     }
 
     /**
@@ -1311,11 +1315,21 @@ class FluentExtension extends DataExtension
         $summaryColumns['Source'] = [
             'title'    => 'Source',
             'callback' => function (Locale $object) {
-                if (!$object->RecordLocale()) {
+                $localeInformation = $object->RecordLocale();
+
+                if (!$localeInformation) {
                     return '';
                 }
 
-                $sourceLocale = $object->RecordLocale()->getSourceLocale();
+                $sourceLocale = FluentState::singleton()->withState(
+                    static function (FluentState $state) use ($localeInformation): ?Locale {
+                        // We are currently in the CMS context, but we want to show to the content author
+                        // what the data state is in the frontend context
+                        $state->setIsFrontend(true);
+
+                        return $localeInformation->getSourceLocale();
+                    }
+                );
 
                 if ($sourceLocale) {
                     return $sourceLocale->getLongTitle();
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -996,7 +996,7 @@ class FluentExtension extends DataExtension
      *
      * @return Locale|null
      */
-    public function getSourceLocale(): ?Locale
+    public function getSourceLocale()
     {
         $currentLocale = FluentState::singleton()->getLocale();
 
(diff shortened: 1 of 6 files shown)


mfendeksilverstripe:
This is ready for a review @GuySartorelli

src/Extension/FluentExtension.php
```diff
@@ -48,6 +48,7 @@
  *
  * @template T of DataObject
  * @extends DataExtension<T&static>
+ * @property DataObject|$this $owner
```
```suggestion
```
That's not needed because of the use of generic typing.

src/Extension/FluentExtension.php
```diff
@@ -968,16 +969,27 @@ protected function getRecordLocale()
...
+     * Source locale for frontend context is used by default as this is the most common use case,
+     * but you can optionally specify CMS context as well
+     * Passing null will fall back to whatever context is currently in use in global fluent state
+     *
+     * @param bool|null $isFrontend
      *
      * @return Locale|null
      */
-    public function getSourceLocale()
+    public function getSourceLocale(?bool $isFrontend = true): ?Locale
```
```suggestion
    public function getSourceLocale()
```
Additional parameters (even optional ones) and explicit typehints cannot be added outside of a major release.
See [the definition of public API](https://docs.silverstripe.org/en/5/project_governance/public_api/)

ChloeSartorelli: If those changes are required, they will need to target `8` so they can be included in the next major release.

src/Model/RecordLocale.php
```diff
@@ -360,23 +360,43 @@ public function getStagesDiffer(): bool
...
      * Get the locale which is the source of content for this record
+     * Source locale for frontend context is used by default as this is the most common use case,
+     * but you can optionally specify CMS context as well
+     * Passing null will fall back to whatever context is currently in use in global fluent state
      *
+     * @param bool|null $isFrontend
      * @return Locale|null
      */
-    public function getSourceLocale(): ?Locale
+    public function getSourceLocale(?bool $isFrontend = true): ?Locale
```
ChloeSartorelli:
```suggestion
    public function getSourceLocale(): ?Locale
```
As above.

GuySartorelli (changes requested):
I haven't checked anything beyond what I've requested changes for below, since these changes will require either retargetting the PR or fundamentally changing the approach.

GuySartorelli:
Please also avoid `@`ing me on each PR unless it is something you crucially need to be checked in the immediate future.

src/Extension/FluentExtension.php
mfendeksilverstripe: I'll see if I can find another way 🤔

mfendeksilverstripe:
New changes added: updated API so they match `minor` constraints, so this can be based on `7` branch.

src/Extension/FluentExtension.php
```diff
@@ -965,19 +964,24 @@ protected function getRecordLocale()
...
     }
 
-
     /**
      * Returns the source locale that will display the content for this record
      *
      * @return Locale|null
      */
-    public function getSourceLocale()
+    public function getSourceLocale(): ?Locale
```
```suggestion
    public function getSourceLocale()
```
as per https://github.com/tractorcow-farm/silverstripe-fluent/pull/840#discussion_r1556482111

GuySartorelli (dismissed):
Haven't reviewed properly yet. I'll add this to my list of things to try get to this week unless someone else gets to it before me

src/Extension/FluentExtension.php
mfendeksilverstripe: whoops, forgot about this, will make the change

src/Model/RecordLocale.php
```diff
@@ -367,16 +375,31 @@ public function getSourceLocale(): ?Locale
...
 
+        // This model has localised data in the current locale so the current locale is also the source locale
         if ($record->existsInLocale($this->getLocale())) {
             return $this->getLocaleObject();
         }
 
+        // This model requires localisation so fallback of any kind is not allowed
+        // hence the content can't come from another locale
+        // We don't have a source locale for such case
+        if ($inheritanceMode === FluentExtension::INHERITANCE_MODE_EXACT) {
```
tractorcow, line 393: Looks good! I was trying to remind myself how inheritance worked, but this line seems to obviously fix the issue. We should never care about fallbacks if inheritance is exact mode only.

GuySartorelli:
Can you please update the PR description to give actual step-by-step instructions for how to reproduce this from a fresh installation? Include any code samples for setting configuration, and exact steps required - assume I don't know anything about fluent :p

As it stands I'm not really sure what the behaviour was before this PR, what it should be afterward, and how to confirm either of those.

mfendeksilverstripe:
I've added test setup details into the description of this PR as requested @GuySartorelli

GuySartorelli:
Rebased - sorry it's taken me so long to look at this. Gonna take some time this afternoon to understand it.

GuySartorelli:
CI is failing for unrelated reasons - assuming it all works well for me locally, I'll spin off a separate CI run to make sure the tests are happy when allowed to actually run.

GuySartorelli (approved):
LGTM. Works well locally. Just waiting for a CI run to check the tests go green: https://github.com/creative-commoners/silverstripe-fluent/actions/runs/13126357953

PR #860, opened 2024-06-06
FIX Issues when running with other modules installed

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/271

There are two separate commits in here fixing different issues, do not squash merge

[FIX Only select columns from the current database](https://github.com/tractorcow-farm/silverstripe-fluent/pull/860/commits/eaece0a7e1c8e3a4492d3d290c84993b52c82b4c) fixes the `Duplicate columns in query` issue

[FIX Create a temp locale if there is not a default](https://github.com/tractorcow-farm/silverstripe-fluent/pull/860/commits/2714bff77c5baebcabf63e5aa281be57f9b05dbf) fixes the `TractorCow\Fluent\Model\RecordLocale::__construct()` issue

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -1055,6 +1055,14 @@ class FluentExtension extends DataExtension
             $localeObj = Locale::getDefault();
         }
 
+        if (!$locale && !$localeObj) {
+            // There is no default locale, this can happen when if fluent is installed though locales have been setup
+            // This will happen when doing integration unit testing, though can also happen during
+            // regular website operation
+            // This temporary Locale is created to prevent a invalid argument exception in RecordLocale::__construct()
+            $localeObj = Locale::create(['Title' => 'Temp locale']);
+        }
+
         return RecordLocale::create($this->owner, $localeObj);
     }
 
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -3,6 +3,7 @@
 namespace TractorCow\Fluent\Extension;
 
 use LogicException;
+use SilverStripe\i18n\i18n;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Convert;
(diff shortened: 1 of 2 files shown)


src/Extension/FluentExtension.php
```diff
@@ -1055,6 +1055,14 @@ public function LocaleInformation($locale = null)
             $localeObj = Locale::getDefault();
         }
 
+        if (!$locale && !$localeObj) {
```
```suggestion
        if (!$localeObj) {
```
We only need the `$localeObj` don't we?

src/Extension/FluentExtension.php
```diff
@@ -1055,6 +1055,14 @@ public function LocaleInformation($locale = null)
...
+            // There is no default locale, this can happen when if fluent is installed though locales have been setup
```
```suggestion
            // There is no default locale, this can happen if no locales have been setup
```
I assume you meant that NO locales have been set up.
The fact that fluent is installed is a given when fluent code is executed, so removed that for brevity.

src/Extension/FluentExtension.php
```diff
@@ -1055,6 +1055,14 @@ public function LocaleInformation($locale = null)
...
+            // This will happen when doing integration unit testing, though can also happen during
+            // regular website operation
+            // This temporary Locale is created to prevent a invalid argument exception in RecordLocale::__construct()
+            $localeObj = Locale::create(['Title' => 'Temp locale']);
```
ChloeSartorelli: Probably should set the `Locale` db field to the current locale defined in [`i18n::get_locale()`](https://api.silverstripe.org/5/SilverStripe/i18n/i18n.html#method_get_locale) - and don't set a title.

GuySartorelli (changes requested):
Please add details about why each commit is actually needed. What errors are popping up, and which commit fixes which error(s)?

It's hard to test this without knowing what's being fixed by which code change.

src/Extension/FluentExtension.php
emteknetnz: Possibly, though the issue I was fixing was specifically for code path where `$locale = null` so it. 

I've moved the code so that it's underneath Locale::getDefault() just to show the intent better. This seems like the most place I can put this that will have the least side-effects. I don't know if there will be any unintended negative side-effects also creating a fake locale for when `Locale::getByLocale($locale)` returns null, or creating a fake locale inside Locale::getDefault().

src/Extension/FluentExtension.php
emteknetnz: Updated

src/Extension/FluentExtension.php
emteknetnz: Updated

src/Extension/FluentExtension.php
emteknetnz: Updated

emteknetnz:
@GuySartorelli have added description to explain which issues commits are for

GuySartorelli (approved):
LGTM

PR #861, opened 2024-06-10
FIX Preview link when no locale set

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/272

Fixes vendor/silverstripe/cms/tests/behat/features/preview-a-page.feature:12 - which breaks when silverstripe/recipe-kitchen-sink is run including silverstripe/cms behat tests.

The issue that behat is running into is that the preview link will not show anything as no locale has been defined in a behat context

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -1056,15 +1056,6 @@ class FluentExtension extends DataExtension
             $localeObj = Locale::getDefault();
         }
 
-        if (!$localeObj) {
-            // There is no default locale, this can happen if no locales have been setup
-            // This will happen when doing integration unit testing, though can also happen during regular
-            // website operation
-            // This temporary Locale is created to prevent a invalid argument exception in
-            // RecordLocale::__construct()
-            $localeObj = Locale::create(['Locale' => i18n::get_locale()]);
-        }
-
         return RecordLocale::create($this->owner, $localeObj);
     }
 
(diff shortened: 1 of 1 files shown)


GuySartorelli (changes requested):
This is starting to feel really flaky.

I've done a root cause analysis and [this PR](https://github.com/tractorcow-farm/silverstripe-fluent/pull/781) is what causes the problem that results in needing this "temp locale" in the first place.

This fixes it without needing a temp locale:
```diff
     public function updatePreviewLink(&$link): void
     {
         $owner = $this->owner;
-        $info = $owner->LocaleInformation(FluentState::singleton()->getLocale());
+        $locale = FluentState::singleton()->getLocale();
+        if ($locale === null || $locale === '') {
+            return;
+        }
+        $info = $owner->LocaleInformation($locale);

         if (!$info->getSourceLocale()) {
             $link = null;
         }
     }
```

With this change, if there is no current locale (i.e. locales haven't been set up yet) it won't update the preview URL. It'll only update it when there _is_ a current locale (i.e. locales _have_ been set up).

emteknetnz:
Updated

GuySartorelli (changes requested):
Should also revert the changes that were made in https://github.com/tractorcow-farm/silverstripe-fluent/pull/860/commits/2714bff77c5baebcabf63e5aa281be57f9b05dbf and https://github.com/tractorcow-farm/silverstripe-fluent/pull/860/commits/b14687a9d09eb8ad8ef54c722e87895e4d256939 - the temporary locale isn't needed anymore.

emteknetnz:
Surprisingly yes, looks like the previous CI failure was actually just from rendering the preview link from a template.

Reverted as second commit, so do not squash merge

GuySartorelli (approved):
LGTM

PR #880, opened 2024-08-20
Fix for non RecordLocale Entites when checking Owner Links

MichaelHoughton:
## Description

This PR fixes issue on some modules where there are missing methods.  The code was creating a bug such as:

Call to undefined method Purl\Fragment::getSourceLocale()

The bug is likely related to legacy Silverstripe applications, which have since been upgraded to SS 5.

In our context, we use Silverstripe for the Irish Parliament website.

## Issues

- Not sure how to reproduce this, however it appears to be related to building the site subtree in Silverstripe.  It appears that not all entities which pass through Fluent are `RecordLocale`, and thus a 500 error is thrown on SS.  This was fine when using SS4, but appears to have come after we upgraded to SS5.

## Pull request checklist

- [ ] This change is covered with tests (or tests aren't necessary for this change)
- [ ] CI is green

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -1068,7 +1068,7 @@ class FluentExtension extends DataExtension
     {
         $locales = [];
         foreach ($this->owner->Locales() as $info) {
-            if ($info->IsDraft()) {
+            if (method_exists($info, 'IsDraft') && $info->IsDraft()) {
                 $locales[] = $info->getLocaleObject();
             }
         }
--- a/src/Extension/FluentSiteTreeExtension.php
+++ b/src/Extension/FluentSiteTreeExtension.php
@@ -279,7 +279,7 @@ class FluentSiteTreeExtension extends FluentVersionedExtension
             $info = $owner->LocaleInformation($locale);
 
             // Our content hasn't been drafted or published.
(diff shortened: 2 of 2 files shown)


src/Extension/FluentExtension.php
```diff
@@ -1068,7 +1068,7 @@ public function getLocaleInstances(): array
     {
         $locales = [];
         foreach ($this->owner->Locales() as $info) {
-            if ($info->IsDraft()) {
+            if (method_exists($info, 'IsDraft') && $info->IsDraft()) {
```
tractorcow, line 1071: if method_exists returns false, what is $info? Is it not an `RecordLocale` instance ?

tractorcow:
Can you give more context on the error cases and what is causing this? I want to make sure I understand the cause before addressing the side effects. This fix is nice but might not address the core issue. E.g. do we need to use something other than `$this->owner` to query locale information?

GuySartorelli:
@MichaelHoughton I've added back the pull request template that you must have removed when you created this pull request. Please fill it out in full.

Note that it has a space for an issue - please create an issue, with details about the bug you're trying to fix including steps to reproduce it.
It's hard to tell for sure because there's not a lot of information in your PR description, but this sounds like it's probably a project-specific bug rather than one that should be addressed with a path in fluent.

Issue #931, opened 2025-03-10
Localised `has_one` relation not duplicated even with `localised_copy` defined


n8-dev:
### Module version(s) affected

^7 (observed in 7.1.4)

### Description

When using the CMS to duplicate a page with Elemental, everything works nicely except for the final update onto the pages `{Page}_Localised` detail.

Elemental post duplication hooks are all happy across things like `DNADesign\Elemental\TopPageFluentExtension` etc.

Something appears to be up within `FluentExtension::onAfterDuplicate()` as it is copying the OLD elemental area ID from the localised table, and no subsequent steps are updating it.

### How to reproduce

Have both Elemental and Fluent installed.

Localise the ElementalAreaID

```php

private static array $field_include = [
        'ElementalAreaID',
    ];

    private static array $localised_copy = [
        'ElementalArea',
    ];
```

Add a block to your elemental page. give it some details

Right click and duplicate the page.

Now interact with the block and you'll notice its the same bound across both pages.

Checking the database you'll see all other references created correctly across tables.

However checking with 
```php
$fields->addFieldToTab(
    'Root.Main',
    ReadOnlyField::create('ElementalAreaID', 'ElementalArea'),
    'Title'
);
```

You can confirm it's loading the wrong ElementalAreaID on the duplicate page.
(which passes into the GraphQL query)

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/tractorcow-farm/silverstripe-fluent/pull/932

GuySartorelli:
PR merged. This will be automatically tagged by GitHub Actions

PR #932, opened 2025-03-10
FIX Correctly set duplicated has_one records for localised records

GuySartorelli:
## Description

When duplicating a localised parent record, any `has_one` relations that should be localised with it still have the child ID of the original parent record. This means any changes to the duplicate child also apply to the original child because they're the same record.

The following configuration all needs to be set on the parent record, listing the child has_one relation in all of them
- `field_include`
- `localised_copy`
- `cascade_duplicates`

## Manual testing steps

1. Create a record with a `has_one` relation and localise that relation. I used elemental applied to a page, but the below config applied to the parent class should work:
    ```php
    private static array $has_one = [
        'MyRelation' => MyDataObject::class,
    ];

    private static array $field_include = [
        'MyRelationID',
    ];
    
    private static array $localised_copy = [
        'MyRelation',
    ];
    
    private static array $cascade_duplicates = [
        'MyRelation',
    ];

    private static array $extensions = [
        FluentExtension::class,
    ];
    ```
1. Create a locale in the CMS through `admin/locales`
1. Make sure you swap to the new locale in the CMS
1. Create the parent record in the CMS
1. Add a child record in the CMS
1. Duplicate the parent record.

Without this PR, the child attached to the new duplicate record is the _same_ as the one attached to the original, even though it should be a new one.

## Issues

- https://github.com/tractorcow-farm/silverstripe-fluent/issues/931

## Pull request checklist

- [x] CI is green

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -24,6 +24,7 @@ use SilverStripe\ORM\FieldType\DBText;
 use SilverStripe\ORM\FieldType\DBVarchar;
 use SilverStripe\ORM\Queries\SQLConditionGroup;
 use SilverStripe\ORM\Queries\SQLSelect;
+use SilverStripe\ORM\Queries\SQLUpdate;
 use SilverStripe\ORM\ValidationException;
 use SilverStripe\Security\Permission;
 use SilverStripe\Versioned\Versioned;
@@ -636,7 +637,18 @@ class FluentExtension extends DataExtension
      */
     public function onAfterDuplicate($original, $doWrite, $relations): void
     {
+        $owner = $this->getOwner();
         $localisedTables = $this->owner->getLocalisedTables();
+        // Get the names of all has_one columns that will have new IDs
+        $copyRelations = (array) $owner->config()->get('localised_copy');
+        $hasOne = $owner->hasOne();
+        $copyHasOneRelations = [];
+        foreach ($copyRelations as $relationName) {
+            if (array_key_exists($relationName, $hasOne)) {
+                $copyHasOneRelations[] = $relationName . 'ID';
+            }
+        }
+        // Add row for new duplicated page in all relevant localised tables
         foreach ($localisedTables as $tableName => $fields) {
             // Target IDs
             $fromID = $original->ID;
@@ -654,6 +666,18 @@ class FluentExtension extends DataExtension
                     SELECT ? AS \"RecordID\", \"Locale\", $fields_str
                     FROM \"$localisedTable\"
                     WHERE \"RecordID\" = ?", [$toID, $fromID]);
+
+            // Make sure to update any has_one IDs - otherwise the new localised table entry will
+            // have the has_one ID from the original record.
+            // The current $owner (i.e. the newly duplicated record) may have already had its relation duplicated
+            // via cascade_duplicates prior to this extension hook being called.
+            $fieldsToUpdate = [];
+            foreach (array_intersect($fields, $copyHasOneRelations) as $copyFieldName) {
+                $fieldsToUpdate[$copyFieldName] = $owner->{$copyFieldName};
+            }
+            if (!empty($fieldsToUpdate)) {
+                SQLUpdate::create($localisedTable, $fieldsToUpdate, ['RecordID' => $toID])->execute();
+            }
         }
     }
 
--- a/tests/php/Extension/LocalisedCopyTest/DuplicationTest.php
+++ b/tests/php/Extension/LocalisedCopyTest/DuplicationTest.php
@@ -238,6 +238,40 @@ class DuplicationTest extends SapphireTest
         });
     }
 
+    /**
+     * case: duplicate() called on localised record
+     * desired outcome: has_one duplication is copied correctly to localised table
+     */
+    public function testDuplicate(): void
+    {
+        FluentState::singleton()->withState(function (FluentState $state): void {
+            $state->setLocale('en_NZ');
+
+            /** @var Horse|FluentExtension $originalHorse */
+            $originalHorse = $this->objFromFixture(Horse::class, 'horse1');
+            $tail = $originalHorse->Tail();
+            $originalTailID = $tail->ID;
+            $horseCountBefore = Horse::get()->count();
+            $tailsCountBefore = Tail::get()->count();
+
+            // Duplicate the horse (and its tail by association)
+            $duplicateHorse = $originalHorse->duplicate();
+
+            $horseCountAfter = Horse::get()->count();
+            $tailsCountAfter = Tail::get()->count();
+            $this->assertEquals($horseCountBefore + 1, $horseCountAfter);
+            $this->assertEquals($tailsCountBefore + 1, $tailsCountAfter);
+
+            // Re-fetch both horses so we are asserting on what's in the DB not just what's in memory
+            $originalHorse = Horse::get()->byID($originalHorse->ID);
+            $duplicateHorse = Horse::get()->byID($duplicateHorse->ID);
+
+            $this->assertNotSame($originalHorse->ID, $duplicateHorse->ID);
+            $this->assertNotSame($originalHorse->TailID, $duplicateHorse->TailID);
+            $this->assertSame($originalTailID, $originalHorse->TailID);
+        });
+    }
+
     public function localesProvider(): array
     {
         return [


src/Extension/FluentExtension.php
```diff
@@ -636,7 +637,18 @@ public function onAfterWrite(): void
      */
     public function onAfterDuplicate($original, $doWrite, $relations): void
     {
+        $owner = $this->getOwner();
         $localisedTables = $this->owner->getLocalisedTables();
+        // Get the names of all has_one columns that will have new IDs
+        $copyRelations = (array) $owner->config()->get('localised_copy');
```
mfendeksilverstripe, line 643: Worth adding a check for `localisedCopyActive`?

src/Extension/FluentExtension.php
```diff
@@ -654,6 +666,16 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
                     WHERE \"RecordID\" = ?", [$toID, $fromID]);
+
+            // Make sure to update any has_one IDs - otherwise the new localised table entry will
+            // have the has_one ID from the original record.
+            $fieldsToUpdate = [];
+            foreach (array_intersect($fields, $copyHasOneRelations) as $copyFieldName) {
+                $fieldsToUpdate[$copyFieldName] = $owner->$copyFieldName;
+            }
+            if (!empty($fieldsToUpdate)) {
+                SQLUpdate::create($localisedTable, $fieldsToUpdate, ['RecordID' => $toID])->execute();
```
mfendeksilverstripe, line 679: This is technically functional but I recommend something that is more ORM friendly - an approach similar to what we already have in `onBeforeWrite`. In our case we should be able to use `onBeforeDuplicate` extension point to achieve this.

mfendeksilverstripe (changes requested):
Current approach is alright but I made a suggestion to try a more ORM friendly approach.

src/Extension/FluentExtension.php
ChloeSartorelli: I'm not sure what that is or how it works. Can you please check if this is also a problem when that is used, and if so, add the necessary context and replication steps to the issue?

src/Extension/FluentExtension.php
ChloeSartorelli: I'm using a low-level API intentionally to mirror the `insert` query above. Ideally setting these values would all be done in a single call, but I wasn't sure how to merge using the correct values into the existing `insert this row based on that existing row` query.

I'm not sure how `onBeforeDuplicate` applies here - can you please explain further?

mfendeksilverstripe:
Looking into other approach (draft only so far):

https://github.com/tractorcow-farm/silverstripe-fluent/pull/933

src/Extension/FluentExtension.php
mfendeksilverstripe: Had a more detailed look on this one and I think it's fine to not have it as the intended case doesn't apply here.
 Context: this property allows you do avoid content duplication when models are being localised for some edge cases like in unit tests. This is not relevant for duplication scenario when a new model is being created.

src/Extension/FluentExtension.php
mfendeksilverstripe: Tried out a different approach using ORM as raw queries should be used only when there is no other option. It turns out that ORM approach has some other issues so I think your approach is overall better. Happy to keep this as is.

tests/php/Extension/LocalisedCopyTest/DuplicationTest.php
```diff
@@ -238,6 +238,31 @@ public function testEditWithInheritedRelation(string $locale, bool $duplicated,
...
+            // Re-select both horses so we are getting what's in the DB not just what's in memory
+            $horse = Horse::get()->byID($horse->ID);
+            $duplicateHorse = Horse::get()->byID($duplicateHorse->ID);
+            $this->assertNotSame($horse, $duplicateHorse->TailID);
+            $this->assertNotSame($originalTailID, $duplicateHorse->TailID);
+            $this->assertNotSame($horse->TailID, $duplicateHorse->TailID);
+            $this->assertSame($originalTailID, $horse->TailID);
+        });
+    }
+
```
mfendeksilverstripe, line 274: Added a more detail test, specifically looking for data redundancy.

```php
/**
 * case: duplicate() called on localised record
 * desired outcome: has_one duplication is copied correctly to localised table
 */
public function testDuplicate(): void
{
    FluentState::singleton()->withState(function (FluentState $state): void {
        $state->setLocale('en_NZ');

        /** @var Horse|FluentExtension $originalHorse */
        $originalHorse = $this->objFromFixture(Horse::class, 'horse1');

        $tail = $originalHorse->Tail();
        $originalTailID = $tail->ID;

        $horseCountBefore = Horse::get()->count();
        $tailsCountBefore = Tail::get()->count();

        $duplicateHorse = $originalHorse->duplicate();

        // Re-fetch the original horse so we are certain we are asserting on the latest data
        $originalHorse = Horse::get()->byID($originalHorse->ID);

        $horseCountAfter = Horse::get()->count();
        $tailsCountAfter = Tail::get()->count();

        $this->assertEquals($horseCountBefore + 1, $horseCountAfter, 'We expect to see only one new horse after duplication');
        $this->assertEquals($tailsCountBefore + 1, $tailsCountAfter, 'We expect to see only one new tail after duplication');

        // Re-select both horses so we are getting what's in the DB not just what's in memory
        $originalHorse = Horse::get()->byID($originalHorse->ID);
        $duplicateHorse = Horse::get()->byID($duplicateHorse->ID);

        $this->assertNotSame($originalHorse->ID, $duplicateHorse->ID, 'We expect original and duplicated horse to have different IDs');
        $this->assertNotSame($originalHorse->TailID, $duplicateHorse->TailID, 'We expect original and duplicated tail to have different IDs');
        $this->assertSame($originalTailID, $originalHorse->TailID, 'We expect the original horse to still have the original tail');
    });
}
```

src/Extension/FluentExtension.php
```diff
@@ -654,6 +666,16 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
                     SELECT ? AS \"RecordID\", \"Locale\", $fields_str
                     FROM \"$localisedTable\"
                     WHERE \"RecordID\" = ?", [$toID, $fromID]);
+
+            // Make sure to update any has_one IDs - otherwise the new localised table entry will
+            // have the has_one ID from the original record.
+            $fieldsToUpdate = [];
+            foreach (array_intersect($fields, $copyHasOneRelations) as $copyFieldName) {
+                $fieldsToUpdate[$copyFieldName] = $owner->$copyFieldName;
```
mfendeksilverstripe: Can you please add to this comment why the model (owner) in this has the correct data? I think that's the bit that's missing here.

mfendeksilverstripe:
```suggestion
                $fieldsToUpdate[$copyFieldName] = $owner->{$copyFieldName};
```
[nitpick]

mfendeksilverstripe:
Hey @GuySartorelli , I finished my testing and assessment. I think your approach is overall better so I closed https://github.com/tractorcow-farm/silverstripe-fluent/pull/933.

I left some minor comments to action but overall your solution works great. Tested on my local and I'm getting the desired outcome (new elemental area ID is assigned with correct contents).

Screenshot 2025-03-11 at 9 24 54 AM

GuySartorelli:
Thanks for taking the time to review, Mojmir. I'll make those requested changes today.

tests/php/Extension/LocalisedCopyTest/DuplicationTest.php
ChloeSartorelli: Done (copied in the new stuff and updated to match the style of the existing stuff)
Didn't re-fetch the original horse twice, though, as that doesn't seem necessary.

src/Extension/FluentExtension.php
ChloeSartorelli: Done

ChloeSartorelli: Done

mfendeksilverstripe (approved):
Nice work 👍 GTG 🚀

emteknetnz:
@mfendeksilverstripe You're probably allowed to merge things you approve :-) You have better knowledge of this PR and module in general than I do.

PR #934, opened 2025-03-11
ENH Add config to set samesite for fluent locale cookie

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/10342


src/Middleware/DetectLocaleMiddleware.php
```diff
@@ -221,12 +226,13 @@ protected function setPersistLocale(HTTPRequest $request, $locale)
                 && Session::config()->get('cookie_secure');
             Cookie::set(
                 $key,
-                $locale,
+                $locale ?? false,
```
ChloeSartorelli, line 229: Needed to respect strong typing

tractorcow (approved):

PR #935, opened 2025-03-12
Merge branch '7.2' into 7

GuySartorelli:
Just a straight merge-up from `7.2` to `7.3`.
After this is merged I'll need to open a new PR to merge to `7` then to `8.0`.

## Issue
- https://github.com/silverstripe/.github/issues/376


prij:
This merge seems to break Elemental with the documented fluent setup:
https://docs.silverstripe.org/en/5/optional_features/elemental/advanced_setup/#:~:text=%24result%3B%20%7D%20%7D-,Elemental%20with%20fluent%20setup,directly%20is%20the%20ElementalArea%20relation.

When I Copy & Publish a page to a new locale, the ElementalAreaID is now getting duplicated.
Is there now a better way to set up indirect localisation with ElementalAreaID that is not documented?

GuySartorelli:
@prij Please open a new GitHub Issue so that we can correctly track and resolve the issue you're experiencing.

Issue #940, opened 2025-03-25
Localised has_one relation not duplicated on extend Classes even with localised_copy defined on Parent Classes


n8-dev:
### Module version(s) affected

7.2.4

### Description

Following on from https://github.com/tractorcow-farm/silverstripe-fluent/issues/931

Say you have the following structure.

`SiteTree` -> `BlockPage` -> `FormPage`

`BlockPage` is where elemental is attached.

Duplicating a `BlockPage` grants a unique `ElementalArea` ✅ 
Duplicating a `FormPage` does not. ⛔ 

### How to reproduce

If you place into the `vendor/tractorcow/silverstripe-fluent/src/Extension/FluentExtension::onAfterDuplicate()`

The following:
```php
            foreach (array_intersect($fields, $copyHasOneRelations) as $copyFieldName) {

                Debug::dump([
                    '$copyFieldName' => $copyFieldName,
                    '$original->{$copyFieldName}' => $original->{$copyFieldName},
                    '$owner->{$copyFieldName}' => $owner->{$copyFieldName},
                ]);

                $fieldsToUpdate[$copyFieldName] = $owner->{$copyFieldName};
            }
```

You'll observe the following.
BlockPage

```
array (
  '$copyFieldName' => 'ElementalAreaID',
  '$original->{$copyFieldName}' => 9,
  '$owner->{$copyFieldName}' => 53,
)
```

FormPage  
```
array (
  '$copyFieldName' => 'ElementalAreaID',
  '$original->{$copyFieldName}' => 49,
  '$owner->{$copyFieldName}' => 49,
)
```
**note the values are the same when they should not be (`49`)**

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/tractorcow-farm/silverstripe-fluent/pull/941
- https://github.com/silverstripe/silverstripe-elemental/pull/1342

mfendeksilverstripe:
I tested this on a subclass of BlockPage on another project and the existing fix was working as expected.

n8-dev:
Oh dear.

Time to dig deeper into the project itself then.

mfendeksilverstripe:
[...]

PR #941, opened 2025-03-25
FIX Multi locale localised copy correction.

mfendeksilverstripe:
# Multi locale localised copy correction.

Fixes https://github.com/tractorcow-farm/silverstripe-fluent/issues/940

Fix spans multiple modules. https://github.com/silverstripe/silverstripe-elemental/pull/1342

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -698,32 +698,10 @@ class FluentExtension extends DataExtension
             return;
         }
 
-        $locales = [];
+        $locales = $this->getLocaleCodesFormModel($owner);
 
-        /** @var RecordLocale $localeInformation */
-        foreach ($owner->Locales() as $localeInformation) {
-            $sourceLocale = $localeInformation->getSourceLocale();
-            $modelLocale = $localeInformation->getLocaleObject();
-
-            if (!$sourceLocale) {
-                // We don't have any source locale, so we can bail out
-                continue;
-            }
-
-            if ($modelLocale->Locale !== $sourceLocale->Locale) {
-                // Source of this locale is different from current locale, so we can skip it
-                // as this locale content is being inherited
-                continue;
-            }
-
-            if ($modelLocale->Locale === $currentLocale) {
-                // Current locale can be skipped as it was already handled correctly
-                continue;
-            }
-
-            // Add locale which uses current locale as a source to our list
-            $locales[] = $modelLocale->Locale;
-        }
+        // Current locale can be skipped as it was already handled correctly
+        $locales = array_diff($locales, [$currentLocale]);
 
         // No locales need to be actioned
         if (count($locales) === 0) {
@@ -1603,5 +1581,39 @@ class FluentExtension extends DataExtension
 
         // all other cases should not duplicate (normal edits)
         return false;
+    }
+
+    /**
+     * Get a list of locale codes that represent locales that the model is localised in
+     * TODO make this method public and use $owner instead as this is a useful utility method
+     *
+     * @param DataObject $model
+     * @return array
+     */
+    private function getLocaleCodesFormModel(DataObject $model): array
+    {
+        $locales = [];
+
+        /** @var RecordLocale $localeInformation */
+        foreach ($model->Locales() as $localeInformation) {
+            $sourceLocale = $localeInformation->getSourceLocale();
+            $modelLocale = $localeInformation->getLocaleObject();
+
+            if (!$sourceLocale) {
+                // We don't have any source locale, so we can bail out
+                continue;
+            }
+
+            if ($modelLocale->Locale !== $sourceLocale->Locale) {
+                // Source of this locale is different from current locale, so we can skip it
+                // as this locale content is being inherited
+                continue;
+            }
+
+            // Add locale which uses current locale as a source to our list
+            $locales[] = $modelLocale->Locale;
+        }
+
+        return $locales;
     }
 }
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -712,19 +712,20 @@ class FluentExtension extends DataExtension
 
         foreach ($locales as $locale) {
             FluentState::singleton()->withState(
-                static function (FluentState $state) use ($owner, $ownerVersioned, $locale, $copyRelations): void {
+                static function (FluentState $state) use ($owner, $original, $ownerVersioned, $locale, $copyRelations): void {
                     $state->setLocale($locale);
 
                     $localisedOwner = DataObject::get($owner->ClassName)->byID($owner->ID);
+                    $localisedOriginal = DataObject::get($original->ClassName)->byID($original->ID);
 
                     // Couldn't find localised data to work with
-                    if (!$localisedOwner) {
+                    if (!$localisedOwner || !$localisedOriginal) {
                         return;
                     }
 
                     // Duplicate all localised relations
                     foreach ($copyRelations as $relation) {
-                        $originalRelation = $localisedOwner->getComponent($relation);
+                        $originalRelation = $localisedOriginal->getComponent($relation);
 
                         if (!$originalRelation instanceof DataObject) {
                             continue;
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -642,12 +642,15 @@ class FluentExtension extends DataExtension
 
         // Get the names of all has_one columns that will have new IDs
         $copyRelations = (array) $owner->config()->get('localised_copy');
+        $cascadeDuplicates = (array) $owner->config()->get('cascade_duplicates');
         $hasOne = $owner->hasOne();
         $copyHasOneRelations = [];
 
         foreach ($copyRelations as $relationName) {
-            if (array_key_exists($relationName, $hasOne)) {
-                $copyHasOneRelations[] = $relationName . 'ID';
+            // We only need to process has_one relations which are covered in both localised_copy & cascade_duplicates
+            // because only such cases need special handling for duplication
+            if (array_key_exists($relationName, $hasOne) && in_array($relationName, $cascadeDuplicates)) {
+                $copyHasOneRelations[$relationName] = $relationName . 'ID';
             }
         }
 
@@ -708,14 +711,17 @@ class FluentExtension extends DataExtension
             return;
         }
 
-        $ownerVersioned = $owner->hasExtension(Versioned::class);
+        $ownerIsVersioned = $owner->hasExtension(Versioned::class);
 
         foreach ($locales as $locale) {
             FluentState::singleton()->withState(
-                static function (FluentState $state) use ($owner, $original, $ownerVersioned, $locale, $copyRelations): void {
+                static function (FluentState $state) use ($owner, $original, $ownerIsVersioned, $locale, $copyRelations): void {
                     $state->setLocale($locale);
 
+                    // This is model which was created by the duplication
                     $localisedOwner = DataObject::get($owner->ClassName)->byID($owner->ID);
+
+                    // This is the model that we were duplicating
                     $localisedOriginal = DataObject::get($original->ClassName)->byID($original->ID);
 
                     // Couldn't find localised data to work with
@@ -724,13 +730,9 @@ class FluentExtension extends DataExtension
                     }
 
                     // Duplicate all localised relations
-                    foreach ($copyRelations as $relation) {
+                    foreach ($copyRelations as $relation => $relationIDField) {
                         $localisedRelation = $localisedOwner->getComponent($relation);
                         $originalRelation = $localisedOriginal->getComponent($relation);
-
-                        if (!$originalRelation instanceof DataObject) {
-                            continue;
-                        }
 
                         if (!$originalRelation->isInDB()) {
                             continue;
@@ -751,17 +753,17 @@ class FluentExtension extends DataExtension
                                 }
                             );
                         } else {
-                            $duplicate = $originalRelation->duplicate();
+                            $duplicate = $originalRelation->duplicate(false);
                         }
 
                         $localisedOwner->setComponent($relation, $duplicate);
                     }
 
                     // Update localised data (without version as this is considered a part of the duplication action)
-                    $localisedOwner->withLocalisedCopyState(function () use ($ownerVersioned, $localisedOwner): void {
+                    $localisedOwner->withLocalisedCopyState(function () use ($ownerIsVersioned, $localisedOwner): void {
                         // Prevent unintended interactions with localised copy feature
                         $localisedOwner->setLocalisedCopyActive(false);
-                        $ownerVersioned
+                        $ownerIsVersioned
                             ? $localisedOwner->writeWithoutVersion()
                             : $localisedOwner->write();
                     });
@@ -1602,13 +1604,6 @@ class FluentExtension extends DataExtension
         return false;
     }
 
-    /**
-     * Get a list of locale codes that represent locales that the model is localised in
-     * TODO make this method public and use $owner instead as this is a useful utility method
-     *
-     * @param DataObject $model
-     * @return array
-     */
     private function getLocaleCodesForModel(DataObject $model): array
     {
         $locales = [];
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -638,21 +638,26 @@ class FluentExtension extends DataExtension
     public function onAfterDuplicate($original, $doWrite, $relations): void
     {
         $owner = $this->getOwner();
-        $localisedTables = $this->owner->getLocalisedTables();
+        $localisedTables = $owner->getLocalisedTables();
+
         // Get the names of all has_one columns that will have new IDs
         $copyRelations = (array) $owner->config()->get('localised_copy');
         $hasOne = $owner->hasOne();
         $copyHasOneRelations = [];
+
         foreach ($copyRelations as $relationName) {
-            if (array_key_exists($relationName, $hasOne)) {
-                $copyHasOneRelations[] = $relationName . 'ID';
+            if (!array_key_exists($relationName, $hasOne)) {
+                continue;
             }
+
+            $copyHasOneRelations[] = $relationName . 'ID';
         }
+
         // Add row for new duplicated page in all relevant localised tables
         foreach ($localisedTables as $tableName => $fields) {
             // Target IDs
             $fromID = $original->ID;
-            $toID = $this->owner->ID;
+            $toID = $owner->ID;
 
             // Get localised table
             $localisedTable = $this->getLocalisedTable($tableName);
@@ -671,6 +676,7 @@ class FluentExtension extends DataExtension
             // have the has_one ID from the original record.
             // The current $owner (i.e. the newly duplicated record) may have already had its relation duplicated
             // via cascade_duplicates prior to this extension hook being called.
+            // This only covers the current locale though
             $fieldsToUpdate = [];
             foreach (array_intersect($fields, $copyHasOneRelations) as $copyFieldName) {
                 $fieldsToUpdate[$copyFieldName] = $owner->{$copyFieldName};
@@ -678,6 +684,86 @@ class FluentExtension extends DataExtension
             if (!empty($fieldsToUpdate)) {
                 SQLUpdate::create($localisedTable, $fieldsToUpdate, ['RecordID' => $toID])->execute();
             }
+        }
+
+        // We don't have any localised relations to cover
+        if (count($copyHasOneRelations) === 0) {
+            return;
+        }
+
+        $currentLocale = FluentState::singleton()->getLocale();
+
+        // We don't have a current locale which indicates that the Fluent setup is incomplete - bail out
+        if (!$currentLocale) {
+            return;
+        }
+
+        $locales = [];
+
+        /** @var RecordLocale $localeInformation */
+        foreach ($owner->Locales() as $localeInformation) {
+            $sourceLocale = $localeInformation->getSourceLocale();
+            $modelLocale = $localeInformation->getLocaleObject();
+
+            if (!$sourceLocale) {
+                // We don't have any source locale, so we can bail out
+                continue;
+            }
+
+            if ($modelLocale->Locale !== $sourceLocale->Locale) {
+                // Source of this locale is different from current locale, so we can skip it
+                // as this locale content is being inherited
+                continue;
+            }
+
+            if ($modelLocale->Locale === $currentLocale) {
+                // Current locale can be skipped as it was already handled correctly
+                continue;
+            }
+
+            // Add locale which uses current locale as a source to our list
+            $locales[] = $modelLocale->Locale;
+        }
+
+        // No locales need to be actioned
+        if (count($locales) === 0) {
+            return;
+        }
+
+        foreach ($locales as $locale) {
+            FluentState::singleton()->withState(
+                static function (FluentState $state) use ($owner, $locale, $copyRelations): void {
+                    $state->setLocale($locale);
+
+                    $localisedOwner = DataObject::get($owner->ClassName)->byID($owner->ID);
+
+                    // Couldn't find localised to to work with
+                    if (!$localisedOwner->exists()) {
+                        return;
+                    }
+
+                    // Duplicate all localised relations
+                    foreach ($copyRelations as $relation) {
+                        $original = $localisedOwner->{$relation}();
+
+                        if (!$original instanceof DataObject) {
+                            continue;
+                        }
+
+                        if (!$original->exists()) {
+                            continue;
+                        }
+
+                        $duplicate = $original->duplicate();
+                        $localisedOwner->setComponent($relation, $duplicate);
+                    }
+
+                    // Update localised data (without version as this is considered a part of the duplication action)
+                    $localisedOwner->hasExtension(Versioned::class)
+                        ? $localisedOwner->writeWithoutVersion()
+                        : $localisedOwner->write();
+                }
+            );
         }
     }
 
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -715,7 +715,7 @@ class FluentExtension extends DataExtension
 
         foreach ($locales as $locale) {
             FluentState::singleton()->withState(
-                static function (FluentState $state) use ($owner, $original, $ownerIsVersioned, $locale, $copyRelations): void {
+                static function (FluentState $state) use ($owner, $original, $ownerIsVersioned, $locale, $copyHasOneRelations): void {
                     $state->setLocale($locale);
 
                     // This is model which was created by the duplication
@@ -730,7 +730,7 @@ class FluentExtension extends DataExtension
                     }
 
                     // Duplicate all localised relations
-                    foreach ($copyRelations as $relation => $relationIDField) {
+                    foreach ($copyHasOneRelations as $relation => $relationIDField) {
                         $localisedRelation = $localisedOwner->getComponent($relation);
                         $originalRelation = $localisedOriginal->getComponent($relation);
 
@@ -744,7 +744,7 @@ class FluentExtension extends DataExtension
                             : 0;
 
                         // Elemental module compatibility
-                        if ($topPageID && $originalRelation->hasMethod('withFixedTopPage')) {
+                        if ($topPageID && $originalRelation->hasExtension(\DNADesign\Elemental\TopPage\DataExtension::class)) {
                             $duplicate = $originalRelation->withFixedTopPage(
                                 $topPageID,
                                 static function () use ($originalRelation): DataObject {
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -642,14 +642,13 @@ class FluentExtension extends DataExtension
 
         // Get the names of all has_one columns that will have new IDs
         $copyRelations = (array) $owner->config()->get('localised_copy');
-        $cascadeDuplicates = (array) $owner->config()->get('cascade_duplicates');
         $hasOne = $owner->hasOne();
         $copyHasOneRelations = [];
 
         foreach ($copyRelations as $relationName) {
-            // We only need to process has_one relations which are covered in both localised_copy & cascade_duplicates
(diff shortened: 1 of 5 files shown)


src/Extension/FluentExtension.php
```diff
@@ -638,21 +638,26 @@ public function onAfterWrite(): void
...
         $copyHasOneRelations = [];
+
         foreach ($copyRelations as $relationName) {
-            if (array_key_exists($relationName, $hasOne)) {
-                $copyHasOneRelations[] = $relationName . 'ID';
+            if (!array_key_exists($relationName, $hasOne)) {
+                continue;
             }
+
+            $copyHasOneRelations[] = $relationName . 'ID';
```
ChloeSartorelli: Why is this change made? Was this just a subjective preference or does it actually affect things somehow?
If the former, please try to avoid introducing subjective preference code changes as it makes PRs more difficult to review.

mfendeksilverstripe:
Can you please have a look at this @GuySartorelli since you already have context 🙏

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+                    $localisedOwner = DataObject::get($owner->ClassName)->byID($owner->ID);
+
+                    // Couldn't find localised data to work with
+                    if (!$localisedOwner->exists()) {
+                        return;
+                    }
+
+                    // Duplicate all localised relations
+                    foreach ($copyRelations as $relation) {
+                        $original = $localisedOwner->{$relation}();
```
ChloeSartorelli: Don't reuse the `$original` variable - it introduces confusion since in `onAfterDuplicate()` that variable usually refers to the original record that was duplicated, resulting in this extension hook being called

```suggestion
                        $originalRelation = $localisedOwner->{$relation}();
```

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+
+                        if (!$original instanceof DataObject) {
+                            continue;
+                        }
```
ChloeSartorelli: How would this ever _not_ be true? Are we silently failing for incorrect configuration here? Or are we saying we only care about `has_one` and `belongs_to` - not the other relation types (if so refer to comment about using `$copyHasOneRelations`) - though if `belongs_to` matters then it probably matters for the current locale too so that needs updating? Or something else?

GuySartorelli:
You don't need to `@` me both on the issue _and_ here :p
Yes I will look at it when I have time

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+
+                        if (!$original->exists()) {
+                            continue;
+                        }
```
```suggestion
                        if (!$original->isInDB()) {
                            continue;
                        }
```
I presume we don't care, for example, if a file doesn't exist in the filesystem - we only care about whether there's a record in the database to duplicate.

e.g. see `DataObject::duplicateHasOneRelation()` which is effectively what we're doing here.

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+
+                        $duplicate = $original->duplicate();
+                        $localisedOwner->setComponent($relation, $duplicate);
```
ChloeSartorelli: This explicitly duplicates the record - but it doesn't check `cascade_duplicates`. Is that going to cause problems?

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
                 SQLUpdate::create($localisedTable, $fieldsToUpdate, ['RecordID' => $toID])->execute();
             }
         }
+
+        // We don't have any localised relations to cover
+        if (count($copyHasOneRelations) === 0) {
```
ChloeSartorelli, line 690:
```suggestion
        if (empty($copyHasOneRelations)) {
```
This is just a personal preference so feel free to leave as is if you want.

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+
+                    $localisedOwner = DataObject::get($owner->ClassName)->byID($owner->ID);
+
+                    // Couldn't find localised data to work with
+                    if (!$localisedOwner->exists()) {
+                        return;
+                    }
+
+                    // Duplicate all localised relations
+                    foreach ($copyRelations as $relation) {
```
ChloeSartorelli: Feels like we should be using `$copyHasOneRelations` if it's only the `has_one` relations we care about (which the early exit earlier implies is the case)

We could for example
1. put the relation name (on its own) as the keys in that array, or else 
1. use the relation name as the values and not add the `ID` suffix until we need to when assigning `$fieldsToUpdate[$copyFieldName]`

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+            }
+
+            if ($modelLocale->Locale === $currentLocale) {
+                // Current locale can be skipped as it was already handled correctly
+                continue;
+            }
+
+            // Add locale which uses current locale as a source to our list
+            $locales[] = $modelLocale->Locale;
+        }
```
ChloeSartorelli: This feels like it should probably be its own method

src/Extension/FluentExtension.php
```diff
@@ -671,6 +676,7 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
             // have the has_one ID from the original record.
             // The current $owner (i.e. the newly duplicated record) may have already had its relation duplicated
             // via cascade_duplicates prior to this extension hook being called.
+            // This only covers the current locale though
```
```suggestion
            // This only covers the current locale - see below for handling additional locales.
```

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
                 SQLUpdate::create($localisedTable, $fieldsToUpdate, ['RecordID' => $toID])->execute();
             }
         }
+
```
ChloeSartorelli, line 686:
```suggestion

// We need to handle duplicating relations in `localised_copy` into additional locales

File: src/Extension/FluentExtension.php [hunk 11]
```diff
@@ -638,21 +638,26 @@ public function onAfterWrite(): void
...
         $copyRelations = (array) $owner->config()->get('localised_copy');
         $hasOne = $owner->hasOne();
         $copyHasOneRelations = [];
+
         foreach ($copyRelations as $relationName) {
-            if (array_key_exists($relationName, $hasOne)) {
-                $copyHasOneRelations[] = $relationName . 'ID';
+            if (!array_key_exists($relationName, $hasOne)) {
+                continue;
             }
```
ChloeSartorelli: We should probably be checking if the relation is in `cascade_duplicates` as well, which I didn't think of when doing the original PR.
Edit: Actually we should be checking against `$relations` - which defaults to `cascade_duplicates` but could be anything if called manually.

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+
+        foreach ($locales as $locale) {
+            FluentState::singleton()->withState(
+                static function (FluentState $state) use ($owner, $locale, $copyRelations): void {
+                    $state->setLocale($locale);
+
+                    $localisedOwner = DataObject::get($owner->ClassName)->byID($owner->ID);
+
+                    // Couldn't find localised data to work with
+                    if (!$localisedOwner->exists()) {
```
```suggestion
                    if (!$localisedOwner) {
```
`DataList::byID()` returns `null` if there's no record.

src/Extension/FluentExtension.php
```suggestion
                        $original = $localisedOwner->getComponent($relation);
```
Just cuts out the magic getters etc and brings us closer to the implementation of `DataObject::duplicateHasOneRelation()`

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+
+                        if (!$original instanceof DataObject) {
+                            continue;
+                        }
+
+                        if (!$original->exists()) {
+                            continue;
+                        }
+
+                        $duplicate = $original->duplicate();
```
```suggestion
                        $duplicate = $original->duplicate(false);
```
Will be written as part of `setComponent()`. Again see `DataObject::duplicateHasOneRelation()`

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+                        $localisedOwner->setComponent($relation, $duplicate);
+                    }
+
+                    // Update localised data (without version as this is considered a part of the duplication action)
+                    $localisedOwner->hasExtension(Versioned::class)
```
ChloeSartorelli: We could do the `hasExtension` check outside the loop since we know all of the `$localisedOwner` will be the same class and have the same extensions. Unsure how much of a performance boost it would be though, probably negligible until you get into insane numbers of locales, so I'm not too fussed either way.

src/Extension/FluentExtension.php
```diff
@@ -679,6 +685,86 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+                        ? $localisedOwner->writeWithoutVersion()
+                        : $localisedOwner->write();
```
ChloeSartorelli: This might result in `makeLocalisedCopy()` being called in some cases - do we need to wrap this in `withLocalisedCopyState()` with `setLocalisedCopyActive(false)`?

GuySartorelli (changes requested):

src/Extension/FluentExtension.php
mfendeksilverstripe: This is just linting (exist early rule), happy to revert it.

src/Extension/FluentExtension.php
mfendeksilverstripe: Good point, will make the change.

src/Extension/FluentExtension.php
mfendeksilverstripe: Localised copy currently only supports `has_one`. This check is likely not needed, can remove it if needed.

src/Extension/FluentExtension.php
mfendeksilverstripe: Yes, agree.

src/Extension/FluentExtension.php
mfendeksilverstripe: `duplicate()` is recursive by default as far as I know, see method signature so it should be covered.

src/Extension/FluentExtension.php
mfendeksilverstripe: Will leave this as it as it's just a preference, personally I don't like using empty() as it's too greedy (will work on many variable types).

src/Extension/FluentExtension.php
mfendeksilverstripe: The approach sounds good but my aim was to keep the code consistent on how localised copy handles this. I would be keen on refactoring this code a bit to make it more dry. I don't think I can do that though under a patch constraint.

src/Extension/FluentExtension.php
mfendeksilverstripe: Yes, agree  - particularly it would be useful to extract out a separate method that gives you a list of locale codes that the model is localised in for a particular stage as we already use a similar feature in multiple projects. I'm happy to make these changes but I don't think I can make it in this PR and stay within patch version constraints.

ChloeSartorelli: Making a private method is perfectly fine in a patch.
Its visibility could then be changed later in a minor release if it seems like it could be useful for purposes outside of this specific class.

mfendeksilverstripe:
Found some more issues related to duplication of other locales, looking into this further.

src/Extension/FluentExtension.php
tractorcow: It wouldn't hurt for you to put the shared logic into a trait, and you could have a discussion about whether you should be only copying has_ones or not there.

The "instanceof DataObject" below basically means this code only works for has_one (or belongs_to) anyway, not *-to-many, so use that as a guide. :)

You probably want to have a discussion about how it should work for belongs_to, and then apply the same logic for *-to-ones for both duplicate and localised copy.

tractorcow: I've been out of the game for a while, so happy for you guys to make the call on how you want it to work at a high level, and I can just provide advice or suggestions. :)

mfendeksilverstripe:
Hey @GuySartorelli I've pushed up new changes, please review.

* fixed data source for duplicated models
* expanded test to cover the above and also the nested models duplication
* added Elemental module compatibility support (not sure if this is the best way to do it, keen on your feedback)

src/Extension/FluentExtension.php
```diff
@@ -1512,4 +1601,38 @@ protected function localisedCopyNeeded(): bool
         // all other cases should not duplicate (normal edits)
         return false;
     }
+
+    /**
+     * Get a list of locale codes that represent locales that the model is localised in
+     * TODO make this method public and use $owner instead as this is a useful utility method
+     *
+     * @param DataObject $model
+     * @return array
```
```suggestion
     * Get a list of locale codes that represent locales that the model is localised in
```
TODO comments in code get forgotten. Raise an issue instead so it can be tracked, discussed, and actioned.

PHPDocs aren't needed if they only provide the same information the actual typehints provide. They're likely to get stale if the actual hints change.

src/Extension/FluentExtension.php
```diff
@@ -638,21 +638,24 @@ public function onAfterWrite(): void
...
         // Get the names of all has_one columns that will have new IDs
         $copyRelations = (array) $owner->config()->get('localised_copy');
         $hasOne = $owner->hasOne();
         $copyHasOneRelations = [];
+
         foreach ($copyRelations as $relationName) {
             if (array_key_exists($relationName, $hasOne)) {
                 $copyHasOneRelations[] = $relationName . 'ID';
             }
         }
```
ChloeSartorelli: We still need to take the `$relations` array into account as per https://github.com/tractorcow-farm/silverstripe-fluent/pull/941#discussion_r2013297096

src/Extension/FluentExtension.php
```diff
@@ -679,6 +683,91 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+                        if ($topPageID && $originalRelation->hasMethod('withFixedTopPage')) {
+                            $duplicate = $originalRelation->withFixedTopPage(
+                                $topPageID,
+                                static function () use ($originalRelation): DataObject {
+                                    // Duplicate needs to include write to store the top page data
+                                    return $originalRelation->duplicate();
+                                }
+                            );
+                        } else {
+                            $duplicate = $originalRelation->duplicate();
```
```suggestion
                            $duplicate = $originalRelation->duplicate(false);
```
There's no comment on this one so I assume we don't need an explicit write here and can rely on `setComponent` calling `write()` as per https://github.com/tractorcow-farm/silverstripe-fluent/pull/941#discussion_r2013304006

src/Extension/FluentExtension.php
```diff
@@ -679,6 +683,91 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+
+                        // Elemental module compatibility
+                        if ($topPageID && $originalRelation->hasMethod('withFixedTopPage')) {
+                            $duplicate = $originalRelation->withFixedTopPage(
+                                $topPageID,
+                                static function () use ($originalRelation): DataObject {
+                                    // Duplicate needs to include write to store the top page data
+                                    return $originalRelation->duplicate();
+                                }
+                            );
```
ChloeSartorelli: Why is this actually needed? What happens if we don't have it?

src/Extension/FluentExtension.php
```diff
@@ -679,6 +683,91 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+
+        // Current locale can be skipped as it was already handled correctly
+        $locales = array_diff($locales, [$currentLocale]);
+
+        // No locales need to be actioned
+        if (count($locales) === 0) {
+            return;
+        }
+
+        $ownerVersioned = $owner->hasExtension(Versioned::class);
```
```suggestion
        $ownerIsVersioned = $owner->hasExtension(Versioned::class);
```
Makes the name very clearly a boolean

src/Extension/FluentExtension.php
```diff
@@ -679,6 +683,91 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+                            continue;
+                        }
+
+                        // Determine if Top page ID information is available
+                        $topPageID = (int) $localisedRelation->hasField('TopPageID')
+                            ? $localisedRelation->TopPageID
+                            : 0;
+
+                        // Elemental module compatibility
+                        if ($topPageID && $originalRelation->hasMethod('withFixedTopPage')) {
```
ChloeSartorelli: Assuming this is specifically for `DNADesign\Elemental\TopPage\DataExtension` or `DNADesign\Elemental\TopPage\FluentExtension` or `DNADesign\Elemental\TopPage\SiteTreeExtension`, we should check specifically for the extension(s) we care about instead of checking for methods and fields.
That method and field could be added to other models for other purposes.

More likely, given there is a fluent extension in the `TopPage` namespace in elemental, it feels like perhaps this should be handled on that end instead?

That said, my question about why this is actually needed still stands - it's hard to know where this should live without knowing why we need it.

src/Extension/FluentExtension.php
ChloeSartorelli: It's not - that's the whole reason `cascade_duplicates` exists. If you don't define a relation as needing to be duplicated, it won't be.

That said, this will be handled by https://github.com/tractorcow-farm/silverstripe-fluent/pull/941#discussion_r2013297096

src/Extension/FluentExtension.php
ChloeSartorelli: I would prefer this be removed if it's not necessary. It makes it look like something more complex is happening or possible here.

src/Extension/FluentExtension.php
ChloeSartorelli: Given we already find the specific relations we care about earlier in this method, I would prefer to keep using that same variable here.

It has several advantages:

1. It makes it immediately clear that the relationships we're dealing with _here_ are the same ones we're dealing with above.
1. It ensures that we don't _accidentally_ deal with different relationships here than we did above - meaning the relationships that are handled for the current locale are _always_ the same ones handled for other locales.
1. It reduces the amount of mental overhead required to understand the code (because of 1)
1. It reduces the amount of work being done (because we already identified the relationships - we don't have to do it again here

While it would be nice if this code and code elsewhere that does the same thing were shared, that's not what's happening here right now.
I think refactoring this into a trait might make sense in a minor release in the future - for this patch I think we should avoid adding new public API. I also think we should avoid mirroring what code elsewhere is doing if that code ends up duplicating the effort that we've already done above.

src/Extension/FluentExtension.php
```diff
@@ -679,6 +683,91 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+
+        $ownerVersioned = $owner->hasExtension(Versioned::class);
+
+        foreach ($locales as $locale) {
+            FluentState::singleton()->withState(
+                static function (FluentState $state) use ($owner, $original, $ownerVersioned, $locale, $copyRelations): void {
+                    $state->setLocale($locale);
+
+                    $localisedOwner = DataObject::get($owner->ClassName)->byID($owner->ID);
+                    $localisedOriginal = DataObject::get($original->ClassName)->byID($original->ID);
```
ChloeSartorelli: What's the difference between these, and why do we need both? I assume this is what you meant by "fixed data source for duplicated models" - we might need some inline comments here so it's clear what these data sources represent and why we need both of them.

GuySartorelli (changes requested):

src/Extension/FluentExtension.php
mfendeksilverstripe: All good, raised https://github.com/tractorcow-farm/silverstripe-fluent/issues/942 to cover this.

src/Extension/FluentExtension.php
mfendeksilverstripe: We get incomplete data - the impact ranges from worse performance to broken features based on your specific project setup.

ChloeSartorelli: Can you please be more specific? What data should I be looking at specifically? What should I be expecting as I test this locally?

src/Extension/FluentExtension.php
mfendeksilverstripe: Added some more comments, here is how this looks like in a local test:

* original page (Int locale)
* -> original text block (Int locale)
* original page (Au locale)
* -> original text block (Au locale)

Trigger duplicate on original page in Int locale.

We get this:

* duplicated page (Int locale)
* -> duplicated text block (Int locale)
* duplicated page (Au locale)
* -> duplicated text block (Au locale)

Content mapping:

* original page (Int locale) - duplicated page (Int locale)
* original text block (Int locale) - duplicated text block (Int locale)
* original page (Au locale) - duplicated page (Au locale)
* original text block (Au locale) - duplicated text block (Au locale)

src/Extension/FluentExtension.php
mfendeksilverstripe: `TopPageID` and `TopPageLocale` DB columns. These are expected to be on ElementalArea and BaseElement records.

src/Extension/FluentExtension.php
mfendeksilverstripe: Have changed the condition to use extension instead of method.

src/Extension/FluentExtension.php
mfendeksilverstripe: See more details on the setup in [Elemental docs](https://github.com/silverstripe/silverstripe-elemental/blob/5/docs/en/02_advanced_setup.md#top-page-reference-performance-enhancement).

Here is my local test data that should help with this.

### Test page (source)

Page

Screenshot 2025-03-27 at 1 28 22 PM

Elemental area

Screenshot 2025-03-27 at 1 29 23 PM

Block

Screenshot 2025-03-27 at 1 30 34 PM

### Duplicated page (after duplication)

Page

Screenshot 2025-03-27 at 1 32 57 PM

Elemental area

Screenshot 2025-03-27 at 1 32 23 PM

Block

Screenshot 2025-03-27 at 1 31 10 PM

Screenshot 2025-03-27 at 1 32 00 PM

Hopefully this helps.

mfendeksilverstripe:
I've pushed up new changes based on your feedback @GuySartorelli , please review.

src/Extension/FluentExtension.php
```diff
@@ -638,21 +638,27 @@ public function onAfterWrite(): void
...
+        $cascadeDuplicates = (array) $owner->config()->get('cascade_duplicates');
         $hasOne = $owner->hasOne();
         $copyHasOneRelations = [];
+
         foreach ($copyRelations as $relationName) {
-            if (array_key_exists($relationName, $hasOne)) {
-                $copyHasOneRelations[] = $relationName . 'ID';
+            // We only need to process has_one relations which are covered in both localised_copy & cascade_duplicates
+            // because only such cases need special handling for duplication
+            if (array_key_exists($relationName, $hasOne) && in_array($relationName, $cascadeDuplicates)) {
```
```suggestion
            // We only need to process has_one relations which are covered in localised_copy and are part of this duplication
            // because only such cases need special handling for duplication
            if (array_key_exists($relationName, $hasOne) && in_array($relationName, $relations)) {
```
The `$relations` array contains the relations which are relevant for this duplication action. Note that you can pass in a custom array to `duplicate()` which will be used instead of `cascade_duplicates`.

Please also add a test scenario where a custom `$relations` array is passed into `duplicate()` which differs from the `cascade_duplicates` configuration array.

src/Extension/FluentExtension.php
```diff
@@ -638,21 +638,27 @@ public function onAfterWrite(): void
     public function onAfterDuplicate($original, $doWrite, $relations): void
     {
         $owner = $this->getOwner();
-        $localisedTables = $this->owner->getLocalisedTables();
+        $localisedTables = $owner->getLocalisedTables();
+
         // Get the names of all has_one columns that will have new IDs
         $copyRelations = (array) $owner->config()->get('localised_copy');
+        $cascadeDuplicates = (array) $owner->config()->get('cascade_duplicates');
```
```suggestion
```
As per https://github.com/tractorcow-farm/silverstripe-fluent/pull/941#discussion_r2015176328

src/Extension/FluentExtension.php
```diff
@@ -679,6 +686,90 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+                            ? $localisedRelation->TopPageID
+                            : 0;
+
+                        // Elemental module compatibility
+                        if ($topPageID && $originalRelation->hasExtension(\DNADesign\Elemental\TopPage\DataExtension::class)) {
+                            $duplicate = $originalRelation->withFixedTopPage(
+                                $topPageID,
+                                static function () use ($originalRelation): DataObject {
+                                    // Duplicate needs to include write to store the top page data
+                                    return $originalRelation->duplicate();
```
ChloeSartorelli: > [!IMPORTANT]
> Ignore this unless we're keeping the `TopPage` stuff in this PR. See https://github.com/tractorcow-farm/silverstripe-fluent/pull/941#discussion_r2020387712

I'm a bit confused about why the call to `duplicate` specifically needs to write here - this still ends up being passed into `setComponent` which will write the record.

Is something special happening in a specific `onAfterDuplicate` or `onBeforeDuplicate` extension hook in the top page extension? If so please call that out specifically in the comment.

src/Extension/FluentExtension.php
```diff
@@ -679,6 +686,90 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+        foreach ($locales as $locale) {
+            FluentState::singleton()->withState(
+                static function (FluentState $state) use ($owner, $original, $ownerIsVersioned, $locale, $copyHasOneRelations): void {
+                    $state->setLocale($locale);
+
+                    // This is model which was created by the duplication
+                    $localisedOwner = DataObject::get($owner->ClassName)->byID($owner->ID);
+
+                    // This is the model that we were duplicating
+                    $localisedOriginal = DataObject::get($original->ClassName)->byID($original->ID);
```
```suggestion
                    // This is localised copy of the record which was created by the duplication
                    $localisedOwner = DataObject::get($owner->ClassName)->byID($owner->ID);

                    // This is the localised copy of the original record that we were duplicating
                    $localisedOriginal = DataObject::get($original->ClassName)->byID($original->ID);
```
This additional context is sort-of self explanatory by the fact we're in `withState()` - but it could be easy to get confused for developers not used to working with fluent or who are just having an off day.

src/Extension/FluentExtension.php
```diff
@@ -679,6 +686,90 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+
+                    // Couldn't find localised data to work with
+                    if (!$localisedOwner || !$localisedOriginal) {
+                        return;
```
ChloeSartorelli: Are there legitimate cases where this would happen? My understanding is we should always have both of these records by now. Should we be logging an error here?

src/Extension/FluentExtension.php
```diff
@@ -679,6 +686,90 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+                    }
+
+                    // Duplicate all localised relations
+                    foreach ($copyHasOneRelations as $relation => $relationIDField) {
+                        $localisedRelation = $localisedOwner->getComponent($relation);
+                        $originalRelation = $localisedOriginal->getComponent($relation);
```
ChloeSartorelli: `$originalRelation` is clear enough - but what is `$localisedRelation` here? That's the same record as `$original->getComponent($relation)` isn't it? i.e. the original unduplicated unlocalised relation?

Why do we only use it to get the `TopPageID`?

src/Extension/FluentExtension.php
```diff
@@ -679,6 +686,90 @@ public function onAfterDuplicate($original, $doWrite, $relations): void
...
+                        $topPageID = (int) $localisedRelation->hasField('TopPageID')
+                            ? $localisedRelation->TopPageID
+                            : 0;
+
+                        // Elemental module compatibility
+                        if ($topPageID && $originalRelation->hasExtension(\DNADesign\Elemental\TopPage\DataExtension::class)) {
+                            $duplicate = $originalRelation->withFixedTopPage(
+                                $topPageID,
+                                static function () use ($originalRelation): DataObject {
+                                    // Duplicate needs to include write to store the top page data
```
ChloeSartorelli: I think handling `TopPage` needs to be moved into `dnadesign/silverstripe-elemental` for the following reasons:

1. `dnadesign/silverstripe-elemental` already has coupling for fluent, but fluent doesn't have coupling for elemental. This would be adding coupling in a new direction that so far we haven't had to deal with, which is complexity I'd prefer to avoid.
1. The work this PR is doing to set the top page will be overridden by the existing code in `DNADesign\Elemental\TopPage\SiteTreeExtension::writeDuplication()` from elemental if the `ElementalPageExtension` is applied to `SiteTree` after `FluentSiteTreeExtension` due to the "first come first served" ordering of extension hook execution.

One option is to call an extension hook from here (e.g. `$originalRelation->extend('onBeforeDuplicateToLocale`, $relation, $relationIDField, $localisedOwner)`) and then use that extension hook in elemental to set the top page after duplication in a way that doesn't have weird interactions with `FluentSiteTreeExtension`'s existing logic.

GuySartorelli (changes requested):

src/Extension/FluentExtension.php
mfendeksilverstripe: Will look into separating this to the Elemental module via an extension as that's where this code belongs.

mfendeksilverstripe: This is now split into https://github.com/silverstripe/silverstripe-elemental/pull/1342

mfendeksilverstripe:
I've pushed up new changes @GuySartorelli please review. The Elemental module related code is now in a separate PR: https://github.com/silverstripe/silverstripe-elemental/pull/1342

GuySartorelli (approved):
Thank you for your patience with me and for your hard work. This is working as expected.

Issue #944, opened 2025-03-30
`Title` and `URLSegment` fields for new `Locale` records should default to the chosen locale.


GuySartorelli:
Currently when you create a new locale, the `Title` and `URLSegment` form fields have a pre-set value which matches the default locale in the `Locale` form field.

It seems like these should be empty by default (as indicated by the placeholder which is set in `getCMSFields()`), and then if the user hasn't explicitly defined them they should be set to values matching the locale the user selected.

This has bugged me every time I've had to set up a locale for testing 😅

## Acceptance Criteria

- [ ] When creating a new `Locale` record, the `Title` and `URLSegment` fields don't have a preset value
- [ ] If the `Title` and `URLSegment` fields are empty when creating the record, they are set to match the chosen locale.
- [ ] If the user has entered a value for `Title` or `URLSegment`, the user's value is used.

## PRs
- https://github.com/tractorcow-farm/silverstripe-fluent/pull/945

GuySartorelli:
PR merged. This will be included in the 8.1 minor release in October.

PR #945, opened 2025-03-30
ENH Set Title and URLSegment fields on write

GuySartorelli:
Currently the logic will automatically set the value for the Title and URLSegment form fields based on the default locale.

These should instead be set to the locale that is chosen unless otherwise explicitly overridden.

Note that this is targetting `8` to be included in a future `8.1` minor release.

## Issue
- https://github.com/tractorcow-farm/silverstripe-fluent/issues/944

### DIFF
--- a/src/Model/Locale.php
+++ b/src/Model/Locale.php
@@ -8,6 +8,7 @@ use Exception;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
 use SilverStripe\Core\Config\Config;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\CheckboxField;
 use SilverStripe\Forms\DropdownField;
 use SilverStripe\Forms\FieldList;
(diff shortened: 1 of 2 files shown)


src/Model/Locale.php
```diff
@@ -155,20 +156,6 @@ public function requireDefaultRecords(): void
         }
     }
 
-    /**
-     * Get internal title for this locale
-     *
-     * @return string
-     */
-    public function getTitle()
```
ChloeSartorelli, line 163: Doesn't need deprecating first - it'll inherit `DataObject::getTitle()`

emteknetnz (changes requested):
Why aren't we just doing this directly to 8.0 (CMS 6 beta) and deprecated in v7 (CMS 5)?

GuySartorelli:
> Why aren't we just doing this directly to 8.0 (CMS 6 beta) and deprecated in v7 (CMS 5)?

This isn't a regression. The breaking change also isn't necessary in order for this change to work, so there's no need for it to be done in 8.0 which has already had its beta release. We should only be doing post-beta breaking changes if they're necessary.

Since this can be done with a deprecation instead of a direct API break, there's no need for it to be in 8.0.

emteknetnz (approved):
Works well

PR #960, opened 2025-05-04
Merge branch '7' into 8.0

GuySartorelli:
Couldn't merge the `7` branch up directly due to conflicts.
image

This PR resolves the conflicts. See https://github.com/tractorcow-farm/silverstripe-fluent/compare/8.0...7 for the raw diff without conflict resolution.

### DIFF
--- a/src/Extension/FluentGridFieldExtension.php
+++ b/src/Extension/FluentGridFieldExtension.php
@@ -31,13 +31,13 @@ class FluentGridFieldExtension extends Extension
      *
      * @param DBField|null $badgeField
      * @see VersionedGridFieldItemRequest::Breadcrumbs()
-     * @deprecated 7.3.0 Will be replaced with TractorCow\Fluent\Extension\FluentExtension::updateStatusFlags()
+     * @deprecated 7.3.0 Will be replaced with TractorCow\Fluent\Extension\FluentExtension::updateStatusFlags() in a future major release
      */
     public function updateBadge(&$badgeField)
     {
         Deprecation::noticeWithNoReplacment(
             '7.3.0',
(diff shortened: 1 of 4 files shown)


tractorcow:
Shouldn't 7 go into 8, not 8.0?

7 could introduce new features, so it should go into a feature branch, not a patch branch.

tractorcow:
It looks like 8.0.0 hasn't been released, so it's fine, but I'd probably not have a 8.0 branch at all if that was the case, and then you could keep the merge strategy consistent. Branch 8 into 8.0 once you tag 8.0.0.

- x.y merge into x to merge patches into feature branch at same version
- x into x+1 to merge new features from older feature branches into newer feature branches
- x.y into (x+1).z to forward-port bugfixes from older versions (with discretion)

GuySartorelli:
Our process currently is to merge up through the patch branch (so 7.3 to 7, 7 to 8.0, 8.0 to 8). Happy to do that differently for this repository if you like though I can't promise we'll remember that every time since that's now how we're handling it with the other supported modules.

The reasoning for the way we currently do it is that only one stable major release line is actually getting new features at a time. 7.3 should therefore be the last stable minor release in the 7.x release line - or at least that's how all other commercially supported modules work and what our release policies imply if not outright say.

tractorcow:
As long as you aren't introducing new features into already released patches, you're fine. If I'm upgrading from 8.0.0 to 8.0.1 I don't want to see any features released in 7.4.0, as a hypothetical example. I'll leave it up to you to use your judgement. :)

GuySartorelli:
> If I'm upgrading from 8.0.0 to 8.0.1 I don't want to see any features released in 7.4.0, as a hypothetical example.

Yup, we don't want that either. 👍 Sounds like we're aligned.

emteknetnz (approved):

PR #975, opened 2025-06-26
ENH Use new DataList caching

GuySartorelli:
Replaces deprecated `DataObject::get_one()` and `DataObject::get_by_id()` in favour of the new DataList query caching.

Also caches the current locale object separately from the `getCached()` result set.

> [!IMPORTANT]
> Relies on https://github.com/silverstripe/silverstripe-framework/pull/11768
> DO NOT MERGE until that PR has been merged in

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11767

### DIFF
--- a/src/Model/Locale.php
+++ b/src/Model/Locale.php
@@ -388,7 +388,6 @@ class Locale extends DataObject implements PermissionProvider
     public static function getCurrentLocale(): ?Locale
     {
         $locale = FluentState::singleton()->getLocale();
-
         return static::getByLocale($locale);
     }
 
@@ -409,7 +408,7 @@ class Locale extends DataObject implements PermissionProvider
         }
 
         // Get filtered locale
-        return Locale::getCached()->find('Locale', $locale);
+        return Locale::get()->setUseCache(true)->find('Locale', $locale);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/Model/Locale.php
```diff
@@ -409,7 +408,7 @@ public static function getByLocale($locale): ?Locale
         }
 
         // Get filtered locale
-        return Locale::getCached()->find('Locale', $locale);
+        return Locale::get()->setUseCache(true)->find('Locale', $locale);
```
ChloeSartorelli, line 411: `getCached()` caches a bunch of records in an `ArrayList`.
Calling `find()` on that list requires filtering through _the full list of locales_ every time this is called, even though we just want the same record every time.

Caching this record separately ends up being more efficient (`O(1)` instead of `O(n)` if that's useful to anyone) and was the main driver for finally adding this level of caching to `DataList` in the first place.

This method gets called _a lot_ (when rendering the site tree it was around 200 times with a site with 10 pages. Imagine a big site...) so it's well worth caching separately.

emteknetnz: Is it worth deprecating `Locale::getCached()`?

emteknetnz (changes requested):

GuySartorelli:
That's still  used in other places.

emteknetnz (approved):

PR #1006, opened 2025-08-27
ENH Hide font icons from screen readers

GuySartorelli:
- Updates CSS to support icons in inner span elements
- Hides decorative icons from screen readers (usually by adding an inner `span`)

I didn't notice any visual difference without the admin PR so we don't need to bump the constraint.

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1957


tractorcow (approved):

tractorcow:
Approved, and I've learned a bit about accessibility too.

client/src/styles/fluent.scss
```diff
@@ -122,7 +122,7 @@ $fluent-color-localised: rgb(102, 15, 242);
     font-size: 1em;
     color: #303b4d;
 
-    &:before {
+    span[class*="font-icon-"]:before {
```
tractorcow: I guess it could just be `span.font-icon` right? Since you have a new span element, you don't need the ":before" anymore? I think the accessibility guidelines in the parent ticket suggested not using ::before?

tractorcow: > Replace ::before pseudo element with a background-image or other CSS that displays the icon and doesn't inject text into the DOM

ChloeSartorelli: > I guess it could just be span.font-icon right? Since you have a new span element, you don't need the ":before" anymore?

Not sure, I'll try that and see if it works without targeting the `:before` pseudo element.

> Replace ::before pseudo element with a background-image or other CSS that displays the icon and doesn't inject text into the DOM

That is an option we have, but would require having the icons available in a non-font format. I thought about adding a svg spritesheet or something but for most icons I think that would have ended up being more work without a clear benefit over just using `aria-hidden` to hide them.

ChloeSartorelli: Looks fine without targeting the `:before` pseudo element so I've updated that

tractorcow (approved):
Great job

client/src/styles/fluent.scss
tractorcow: Looks good to me. :)

emteknetnz (approved):

Issue #913, opened 2024-12-11
Fluent/Elemental: all translations but the last one saved are marked as modified


wernerkrauss:
### Module version(s) affected

5.2, 5.3

### Description

When working with Elemental and BaseElement has FluentVersionedExtension applied, all translations of a block but the one saved last are marked as "Modified". That's cause `Versioned::isLiveVersion()` doesn't know about Fluent and having different Versions per Locale.

### How to reproduce

* Create a project with Fluent and Elemental; 
* Apply FluentVersionedExtension to BaseElement
  ```yml
  DNADesign\Elemental\Models\BaseElement:
    extensions:
      fluent: 'TractorCow\Fluent\Extension\FluentVersionedExtension'
  ```
* Create a bunch of locales
* Save a block and translate it to different Locales; publish the block in each locale.

### Possible Solution

A possible solution would be to overwrite isLiveVersion() in FluentVersionedExtension:
```php
    /**
     * Fixes a bug in Elemental/Fluent, where all translations but the last one saved are marked as modified,
     * cause Fluent saves a new Version per locale.
     * @return bool
     */
    public function isLiveVersion()
    {
        return !$this->getOwner()->stagesDifferInLocale();
    }
```

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs

- https://github.com/tractorcow-farm/silverstripe-fluent/pull/1007 - revisit for major release

GuySartorelli:
Can you please provide some specific step-by-step details for these two? For the extension, are any additional configurations set at all? For translating the block to different locales, there's multiple ways to do that.

> - Apply FluentVersionedExtension to BaseElement
> ...
> - Save a block and translate it to different Locales; publish the block in each locale.

lerni:
@wernerkrauss - Thanks for the workaround! Working perfectly for me. Was about to investigate this myself, so you saved me considerable time. 
```
silverstripe/framework              5.4.9
tractorcow/silverstripe-fluent      7.3.2
dnadesign/silverstripe-elemental    5.4.2
```
Would love to see this make it into an upstream fix! 🤞

GuySartorelli:
> Would love to see this make it into an upstream fix!

[...]

PR #1007, opened 2025-08-29
FIX Ensure Version column is included in draft_localised table

GuySartorelli:
> [!IMPORTANT]
> Needs https://github.com/silverstripe/silverstripe-versioned/pull/526 to be merged for CI to go green. Merge that first, then rerun CI.

Various versioning functionality relies on having the correct Version value in the draft record - for example `Versioned::isLiveVersion()` checks the most recent live version against the current (usually draft) record version.

Fluent updates the values for `get_versionnumber_by_stage` through an extension hook, but this results in having the correctly localised live version number with the incorrect non-localised draft version number - which don't match.

Follow reproduction steps in the linked issue for testing. Note that due to https://github.com/silverstripe/silverstripe-elemental/issues/1266 you will need to add this workaround to allow you to edit the block:

```php
<?php

use SilverStripe\Core\Extension;
use SilverStripe\Forms\FieldList;

class BaseElementExtension extends Extension
{
    protected function updateCMSFields(FieldList $fields): void
    {
        if ($this->getOwner()::config()->get('inline_editable')) {
            $fields->removeByName('RecordLocales');
        }
    }
}
```

```yml
DNADesign\Elemental\Models\BaseElement:
  extensions:
    - 'BaseElementExtension'
```

## Issue

- https://github.com/tractorcow-farm/silverstripe-fluent/issues/913

### DIFF
--- a/src/Extension/FluentVersionedExtension.php
+++ b/src/Extension/FluentVersionedExtension.php
@@ -151,6 +151,15 @@ class FluentVersionedExtension extends FluentExtension implements Resettable
         DB::require_table($localisedTable . FluentVersionedExtension::SUFFIX_VERSIONS, $versionsFields, $versionsIndexes, false);
     }
 
+    protected function isFieldLocalised($field, $type, $class)
+    {
+        // Explicitly localise version columns, which are necessary for some versioning functionality to work correctly.
+        if (array_key_exists($field, $this->defaultVersionsFields)) {
+            return true;
+        }
+        return parent::isFieldLocalised($field, $type, $class);
+    }
+
     /**
      * {@inheritDoc}
      *
--- a/tests/php/Extension/FluentSiteTreeExtensionTest.yml
+++ b/tests/php/Extension/FluentSiteTreeExtensionTest.yml
@@ -54,22 +54,3 @@ Page:
     Title: Staff
     URLSegment: my-staff
     Parent: =>Page.about
-
-SiteTree_Localised:
-  nz-page:
-    RecordID: =>Page.nz-page
-    Title: A Page
-    URLSegment: a-page
-    Locale: de_DE
-  about:
-    RecordID: =>Page.about
-    Title: About Us
-    URLSegment: about-us
-    Locale: de_DE
-
-SiteTree_Localised_Live:
-  nz-page:
-    RecordID: =>Page.nz-page
-    Title: A Page
-    URLSegment: a-page
-    Locale: de_DE
--- a/tests/php/Extension/FluentVersionedExtensionTest.php
+++ b/tests/php/Extension/FluentVersionedExtensionTest.php
@@ -11,10 +11,15 @@ use TractorCow\Fluent\Extension\FluentVersionedExtension;
 use TractorCow\Fluent\Model\Domain;
 use TractorCow\Fluent\Model\Locale;
 use TractorCow\Fluent\State\FluentState;
+use TractorCow\Fluent\Tests\Extension\FluentVersionedExtensionTest\TestVersionedModel;
 
 class FluentVersionedExtensionTest extends SapphireTest
 {
     protected static $fixture_file = 'FluentVersionedExtensionTest.yml';
+
+    protected static $extra_dataobjects = [
+        TestVersionedModel::class
+    ];
 
     protected static $required_extensions = [
         SiteTree::class => [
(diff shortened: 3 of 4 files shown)


tests/php/Extension/FluentSiteTreeExtensionTest.yml
```diff
@@ -54,22 +54,3 @@ Page:
...
-  nz-page:
-    RecordID: =>Page.nz-page
-    Title: A Page
-    URLSegment: a-page
-    Locale: de_DE
-  about:
-    RecordID: =>Page.about
-    Title: About Us
-    URLSegment: about-us
-    Locale: de_DE
```
ChloeSartorelli, line 68: Removing this is necessary for `FluentSiteTreeExtensionTest::testUpdateCMSActionsDrafted()` to pass - see failure in https://github.com/tractorcow-farm/silverstripe-fluent/actions/runs/17311586305/job/49146679097

This fixture was adding incomplete localisation data to the localised table. Removing it allows fluent to build the correct localisation data and means we're testing the _actual_ functionality rather than just testing whether the fixture has the right localisation data or not.

Removing the fixture data for `nz-page` isn't strictly necessary for _this_ PR but it should be removed for the same reasons.

tests/php/Extension/FluentVersionedExtensionTest.php
```diff
@@ -206,4 +211,127 @@ public function testStagesDifferInLocale(): void
         // No Live version
         $this->assertTrue($page->stagesDifferInLocale());
     }
+
+    /**
+     * Tests various Versioned methods correctly identify versions in different locales
+     */
+    public function testCorrectVersionIdentified(): void
```
ChloeSartorelli, line 218: Intentionally not using data providers here because while there are several scenarios here (draft and published in base record and in localised record), they each rely on the previous scenario happening successfully first.

tests/php/Extension/FluentVersionedExtensionTest.php
```diff
@@ -206,4 +211,164 @@ public function testStagesDifferInLocale(): void
...
+            $this->assertFalse($localisedRecord->isOnLiveOnly());
+        });
+
+        // Check still only true for draft stuff in *alternate* locale
+        FluentState::singleton()->withState(function (FluentState $newState) use ($recordID) {
+            $newState->setLocale('en_US');
+            $localisedRecord = TestVersionedModel::get()->byID($recordID);
+            $this->assertTrue($localisedRecord->isOnDraft());
+            $this->assertFalse($localisedRecord->isPublished());
+            $this->assertTrue($localisedRecord->isLatestDraftVersion());
```
ChloeSartorelli, line 333: This specific call to `isLatestDraftVersion()` is where the tests started failing without the fix.

GuySartorelli:
@mfendeksilverstripe You're probably the world's leading expert in this module at this point - do you have a few minutes to look at this? If so, can you please let me know if there's any obvious problems with this fix?

tests/php/Extension/FluentVersionedExtensionTest.php
```diff
@@ -206,4 +211,164 @@ public function testStagesDifferInLocale(): void
...
+            $localisedRecord = TestVersionedModel::get()->byID($recordID);
+            $this->assertTrue($localisedRecord->isOnDraft());
+            $this->assertTrue($localisedRecord->isPublished());
+            $this->assertTrue($localisedRecord->isLatestDraftVersion());
+            $this->assertTrue($localisedRecord->isLatestVersion());
+            $this->assertTrue($localisedRecord->isLiveVersion());
+            $this->assertFalse($localisedRecord->isModifiedOnDraft());
+            $this->assertFalse($localisedRecord->isOnLiveOnly());
+        });
+    }
```
mfendeksilverstripe, line 373: Can you please add some test cases for the UK locale? This one inherits content from another locale.

tests/php/Extension/FluentVersionedExtensionTest/TestVersionedModel.php
```diff
@@ -0,0 +1,25 @@
...
+
+use SilverStripe\Dev\TestOnly;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
+use TractorCow\Fluent\Extension\FluentVersionedExtension;
+
+/**
+ * @mixin Versioned
+ */
+class TestVersionedModel extends DataObject implements TestOnly
```
mfendeksilverstripe, line 13: I think your test model should explicitly define the locale inheritance mode rather than relying on default to prevent ambiguity.

src/Extension/FluentVersionedExtension.php
```diff
@@ -151,6 +151,15 @@ protected function augmentDatabaseRequireTable($localisedTable, $fields, $indexe
         DB::require_table($localisedTable . FluentVersionedExtension::SUFFIX_VERSIONS, $versionsFields, $versionsIndexes, false);
     }
 
+    protected function isFieldLocalised($field, $type, $class)
```
mfendeksilverstripe, line 154: This might have impact on dev/build, note to self: need to cover this by detailed locale testing.

tests/php/Extension/FluentVersionedExtensionTest.php
```diff
@@ -206,4 +211,164 @@ public function testStagesDifferInLocale(): void
...
     }
+
+    /**
+     * Tests various Versioned methods correctly identify versions in different locales
+     */
+    public function testCorrectVersionIdentified(): void
+    {
+        $recordID = null;
+        // Create a new record in the base record only, no localisation
+        FluentState::singleton()->withState(function (FluentState $newState) use (&$recordID) {
```
mfendeksilverstripe, line 222:
```suggestion
        FluentState::singleton()->withState(function (FluentState $state) use (&$recordID) {
```
[nitpick] Simplify variable name.

tests/php/Extension/FluentVersionedExtensionTest.php
```diff
@@ -206,4 +211,164 @@ public function testStagesDifferInLocale(): void
...
+            $newState->setLocale(null);
+            $record = new TestVersionedModel(['Title' => 'Initial value']);
```
mfendeksilverstripe, line 224: [nitpick] Prefer property assignment over array data.

mfendeksilverstripe (changes requested):
Left some early feedback, need to do some more detailed testing as localised versions is a complex area.

mfendeksilverstripe:
@GuySartorelli I had a detailed look today and I think we need to consider broader context here before we can make a change to fix the issue.

The current approach on versioned data is as follows:

* version history data does not follow locale inheritance rules
* this was a decision that was made in Fluent 5 I think, it was based on the impact on features such as how page archive would work (glossing over a lot of detail here but can provide further context if needed)
* this leads to the implementation approach of using the extension points in Versioned for specific methods and provide overrides, for example `Versioned::stagesDiffer()` has an extension point `updateStagesDiffer` which is then leveraged by `FluentVersionedExtension::updateStagesDiffer()` to alter the behaviour

With this context, I think the proposed fix doesn't align with this approach so instead we should add a fix via the `Versioned::isOnDraft()` which has the `updateIsOnDraft` extension hook. `FluentVersionedExtension::updateIsOnDraft()` should be used to override the draft version number to be the correct one. This should then also fix `isModifiedOnDraft()` as a consequence.

I wanted to highlight that we can choose to change the current approach to versioned history localisation but that requires a dedicated discussion and should probably go into a major release.

I also managed to confirm that the reported issue can be reproduced as this is a known issue on several projects that I'm across - we avoid using `isModifiedOnDraft()` and instead leverage `isPublished()` and `stagesDiffer()` as a workaround.

tractorcow:
I can give a little bit of context. Because the _versioned and _versioned_localised tables are intended to work alongside the existing versioned extension, and the silverstripe convention for additional tables, the "version" itself isn't technically a localised field, in the same way that "id" is not a localised field. It's an identifier for a version.

There is always a new version for each new change, even if that change only occurs in one locale. I.e. version 1 = first version, localised in en_UK. version 2 = next version, localised in zh_CN. This has the effect of "isLatestVersion" only ever, at least historically, been true for the last locale written. I'm not sure how versioned works in modern time. :) Versioned had a LOT of static methods that couldn't be extended, so I think I just ignored them for the most part.

To bring it back this PR in context, I'm not sure that isFieldLocalised being true for "Version" is correct, in the same way that "ID" isn't localised either.

tests/php/Extension/FluentVersionedExtensionTest/TestVersionedModel.php
ChloeSartorelli: Can you please provide me with an appropriate snippet for that? I'm not sure which locale inheritance mode this should use if declared explicitly.

GuySartorelli:
> To bring it back this PR in context, I'm not sure that isFieldLocalised being true for "Version" is correct, in the same way that "ID" isn't localised either.

Can you think of another way to resolve this bug?

`Versioned` (and consequently everything that consumes the `Versioned` API downstream) doesn't have a concept of locales. Because of this from a pure `Versioned` perspective it's unclear whether `isLatestVersion()` should return a localised value or a cross-locales value.

That said, I can't think of any scenario where you'd want to call `isLatestVersion()` to check whether this _locale_ was the most recently updated _locale_ - rather, I would expect that you'd want to know _within this locale_ is the current version of this record the most recent version of this record.

In other words, when I'm in `en_US` and I call `isLatestVersion()` on a record, I would personally expect that method call (and most if not all method calls) to give me a response that is relevant to the `en_US` locale specifically.

GuySartorelli:
> With this context, I think the proposed fix doesn't align with this approach so instead we should add a fix via the Versioned::isOnDraft() which has the updateIsOnDraft extension hook. FluentVersionedExtension::updateIsOnDraft() should be used to override the draft version number to be the correct one. This should then also fix isModifiedOnDraft() as a consequence.

This would probably resolve the narrow scope of this specific bug, but it seems inconsitent to me that `isOnDraft()` and `isModifiedOnDraft()` would be localised but `isLatestVersion()` would not. There may be other related bugs lurking undiscovered as a result of this inconsistency.

That said, if you and @tractorcow agree that's the appropriate way forward, I'll update the PR accordingly.

GuySartorelli:
@mfendeksilverstripe Just to clarify, is the thumbs up react "Yes I agree with what you're saying, leave it as is"? Or is it meant to convey "Yes please update the PR"?

mfendeksilverstripe:
@GuySartorelli Yes, please update the PR as per our recommendation.

GuySartorelli:
@mfendeksilverstripe That hasn't worked. I tried more-or-less blindly implementing `updateIsOnDraft()` but that extension hook doesn't even get called and so doesn't solve the problem. The implementation I had, in case that helped discussion later on, is:
```php
    public function updateIsOnDraft(bool &$isOnDraft): void
    {
        $locale = FluentState::singleton()->getLocale();
        if (!$locale) {
            return;
        }
        $isOnDraft = $this->getOwner()->isDraftedInLocale($locale);
    }
```

But ultimately that's moot. As discussed in https://github.com/tractorcow-farm/silverstripe-fluent/issues/913#issuecomment-3235267565 where I mention the root cause of this problem, the `Versioned::isLiveVersion()` method is where we're getting `false` even though we do have the latest _localised_ version.

`isLiveVersion()` compares the value of `get_versionnumber_by_stage()` (which is a localised result because of `FluentVersionedExtension`) against `$this->owner->Version` - which is _not localised_. This naturally results in a mismatch, because 9 times out of 10 the latest localised version _won't_ match the _unlocalised_ latest version. That's the heart of the problem here.

I guess I could add an extension hook to `isLiveVersion()` and then in `FluentVersionedExtension::updateIsLiveVersion()` we can return the correct comparison? I'm not sure what that comparison would look like yet to be honest (maybe if it's published and not modified on draft? That's not really asking about _this version_ of the record though since we don't know what version _this_ version is because that value isn't localised... so it gets a bit messy) and this feels like we're just putting a bandaid on the problem.

I really do strongly think that _not_ having the `Version` field localised is causing more problems than it's solving.

mfendeksilverstripe:
@GuySartorelli 

I've booked some time to look into the patch for this issue next week. As mentioned previously, if we are aiming to patch this we are likely to use the suggested approach not the initial one due to the latter being a breaking change that would need to go into a major release.

GuySartorelli:
I don't understand how it's a breaking change (what does it break?) but if you can find a sensible way to implement a fix without localising the Version column that would be fine.

mfendeksilverstripe:
@GuySartorelli Managed to complete the detailed assessment of your change.

Overall, I think this change is good in terms that it delivers what it's meant to do. The current behaviour likely has gaps.

Base table has a `Version` DB field present.

Screenshot 2025-10-13 at 4 37 10 PM

Looking at localised tables:

Before your change - `Version` DB field absent.

Screenshot 2025-10-13 at 4 36 31 PM

After your change - `Version` DB field present.

Screenshot 2025-10-13 at 4 36 40 PM

As I suspected, this is indeed a breaking change as just applying the patch and running dev/build will not populate the newly created field with correct values, instead it will default to `0`.

You will either need to provide a data migration dev task or at least guidance on how to populate this new value (considering some projects have tens of thousands of models in their DB).

This brings me back to my original question - would you like to implement a `7` patch or are you happy to just fix this in `8`?

For backwards compatible workaround, it might be possible to find a place where we can override `Version` model property with the correct value but based on my assessment, this likely isn't trivial.

Context:

`Versioned::augmentWrite()` assigns to `Version` property coming from base table - this needs to be coming from localised version instead.

GuySartorelli:
> As I suspected, this is indeed a breaking change as just applying the patch and running dev/build will not populate the newly created field with correct values, instead it will default to 0.

Ahhh thank you, that was the context I was missing - and should have been obvious to me, but that's hindsight for ya.

> This brings me back to my original question - would you like to implement a 7 patch or are you happy to just fix this in 8?

I'll have a quick play around hopefully this week to see if I can get `$record->Version` to fetch the localised version (property fetching goes through `__get()` so it _might) be possible). If I can't do that, it will have to wait not for a new minor version in `8` but for the next major version since breaking changes must be in majors, not any sooner.

Marking draft in the mean time.

_maybe_ an okay halfway point is to implement Werner's workaround from the linked issue? I'd have to look at what side effects that might have and it's not a complete solution... as I say, I'll try getting `->Version` to return the correct value first.

GuySartorelli:
Actually nevermind, the idea of tricking `__get()` into getting the localised value falls flat on its face because I don't know _which version_ of the record it is... because we don't have that value yet :rofl: 

I'll close this PR - it will be suitable for a major release but that can't be merged until we open a new major branch which we don't do until we're ready to start working on the next release across the board.

PR #1010, opened 2025-11-17
DOC Add composer install command and repository link

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/465

### DIFF
--- a/docs/en/index.md
+++ b/docs/en/index.md
@@ -10,6 +10,10 @@ The Fluent module provides multi-language translation and localisation for your 
 
 If you were looking for localisation of hardcoded strings, go to the [i18n docs](https://docs.silverstripe.org/en/developer_guides/i18n/)
 
+## Installation
+
+Use composer to install fluent, see [installation](./01_installation.md).
+
 ## GitHub repository
 
 <https://github.com/tractorcow-farm/silverstripe-fluent>


docs/en/index.md
```diff
@@ -10,6 +10,10 @@ The Fluent module provides multi-language translation and localisation for your
 
 If you were looking for localisation of hardcoded strings, go to the [i18n docs](https://docs.silverstripe.org/en/developer_guides/i18n/)
 
+## Installation
+
+Use composer to install fluent, see [installation](./01_installation.md).
```
ChloeSartorelli, line 15: Why is this one done differently to all the others? We should be putting a composer command here and renaming the old installation page (or deleting if no longer relevant)

GuySartorelli (changes requested):

docs/en/index.md
emteknetnz: The installation page also has the following on it:

```md
After installing, here are some common steps you should do:

- Run a `dev/build` with `flush=1` to ensure all additional table fields have been generated
- Configure your locales in the `/admin/locales` section
- Publish pages in each of the locales you want them to be visible in

Fluent will automatically localise SiteTree objects. If you want to localise other DataObjects you will need to
add the appropriate extension yourself.

Please note that if your DataObject is versioned you will need to use the
[`FluentVersionedExtension`](api:TractorCow\Fluent\Extension\FluentVersionedExtension), and it must be applied *after* the [`Versioned`](api:SilverStripe\Versioned\Versioned) extension. You can achieve this by
using an `after: '#versionedfiles'` condition in your YAML configuration block title.

For more information please see [Configuration](./03_configuration.md).
```

Feels like that should stay on its own standalone installation page.

ChloeSartorelli: Fair enough

GuySartorelli (approved):
LGTM

PR #1023, opened 2026-02-17
Merge branch '7' into 8.1

GuySartorelli:
Straight merge-up

After this we need to do 8.1 to 8

## Issue

- https://github.com/silverstripe/.github/issues/476

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,56 +2,48 @@
 # yarn lockfile v1
 
 
-"@ampproject/remapping@^2.2.0":
-  version "2.3.0"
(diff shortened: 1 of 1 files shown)


emteknetnz (changes requested):
Normally we don't merge up yarn.lock - though the deps between the two majors look very similar

I did notice that flat-cache we from 6.1.12 to 5.0.0 (i.e. down a version) which is unexpected.

Perhaps create a new PR branch where you do the merge up and either checkout the yarn.lock file (will be no changes, not sure if github will allow it), or, the merge-up and run yarn update + yarn build and submit that PR.

Issue #252, opened 2017-03-13
"duplicate page" feature leads to conflict in URL segment


leonixyz:
Hello,

Fluent version 3.8.0 seems to have a problem with CMS' built-in "duplicate page" feature.
Tested with silverstripe/cms and silverstripe/framework version 3.5.2.

If one duplicates a page, the URL segment is updated only for the page in the main language (a "-2" is appended to the segment). For other languages, however, the segment is kept untouched,  leading to a conflict with the original page.

This is a big problem if one is using the CMS in split mode, and tries to edit the new page for a non-main language: the CMS automatically redirects the user to the original page.

I think there should be some code in Fluent handling page duplication, and updating the URL segment for all languages.

Hope I explained it well.

Thanks

tractorcow:
Thanks for reporting this. It seems like a necessary and easy to achieve feature; We should be able to simply hook into onAfterDuplicate and reset the urlsegment on all locales.

Zauberfisch:
still an issue in silverstripe 5.
Without testing it in detail, my guess is that it's caused by the raw sql insert being run in  `FluentExtension->onAfterDuplicate()`.

As a workaround I added the following to my Page class which fixes the issue:
```

        public function duplicate(bool $doWrite = true, array|null $relations = null): static
        {
            $new = parent::duplicate($doWrite, $relations);
            $new->forceChange();
            $new->write();
            return $new;
        }
```

adiwidjaja:
This is still an issue in Silverstripe 6 and it's more visible because silverstripe-admin redirects to the original page when you try to edit the new page.

I added a pull request fixing this.

PR #1031, opened 2026-05-19
FIX Re-uniquify localised URLSegment after duplicate (#252)

adiwidjaja:
## Description

onAfterDuplicate raw-copies the original record's localised rows into the duplicate (including URLSegment), bypassing SiteTree::validURLSegment which only checks the base table. The duplicate ends up with a unique base URLSegment but a colliding localised URLSegment, which makes the duplicate unreachable on the frontend and causes the CMS preview to re-route the edit form to the original page via the x-page-id/x-cms-edit-link sync.

After the localised copy, walk each locale row and append a numeric suffix until no other record in the same locale uses the same URLSegment. Mirror the resulting value into the localised _Versions table so version history stays consistent with the live record.

Adds a regression test on FluentSiteTreeExtensionTest that asserts unique URLSegments in both the live and _Versions localised tables after duplicate.

## Manual testing steps

On a Silverstripe CMS 6 installation with fluent: Use the duplicate function in the tree. The duplicate will have a different URLSegment and you can edit it.

## Issues
#252 

## Pull request checklist
- [x] CI is green

### DIFF
--- a/src/Extension/FluentExtension.php
+++ b/src/Extension/FluentExtension.php
@@ -21,9 +21,9 @@ use SilverStripe\ORM\FieldType\DBHTMLVarchar;
 use SilverStripe\ORM\FieldType\DBText;
 use SilverStripe\ORM\FieldType\DBVarchar;
 use SilverStripe\ORM\Queries\SQLConditionGroup;
+use SilverStripe\ORM\Queries\SQLInsert;
 use SilverStripe\ORM\Queries\SQLSelect;
 use SilverStripe\Core\Validation\ValidationException;
-use SilverStripe\ORM\Queries\SQLUpdate;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\View\HTML;
 use TractorCow\Fluent\Extension\Traits\FluentBadgeTrait;
@@ -642,6 +642,13 @@ class FluentExtension extends Extension
 
     /**
      * If an object is duplicated also duplicate existing localised values from original to new object.
+     *
+     * Each localised row is read from the original, prepared via
+     * {@see self::prepareLocalisedRowForDuplicate()} (extension point for subclasses
+     * to mutate field values before they hit the DB), and then inserted as a single
+     * write per row. This avoids the previous DELETE + INSERT…SELECT + post-hoc
+     * UPDATE pattern and lets subclasses (e.g. FluentSiteTreeExtension) provide
+     * already-correct values rather than fix them up afterwards.
      */
     public function onAfterDuplicate($original, $doWrite, $relations): void
     {
@@ -663,35 +670,20 @@ class FluentExtension extends Extension
 
         // Add row for new duplicated page in all relevant localised tables
         foreach ($localisedTables as $tableName => $fields) {
-            // Target IDs
             $fromID = $original->ID;
             $toID = $owner->ID;
-
-            // Get localised table
             $localisedTable = $this->getLocalisedTable($tableName);
 
-            // Remove existing translations from duplicated object
-            DB::prepared_query("DELETE FROM \"$localisedTable\" WHERE \"RecordID\" = ?", [$toID]);
-
-            // Copy translations to duplicated object
-            $fields_str = '"' . implode('","', $fields) . '"';
-            DB::prepared_query("INSERT INTO \"$localisedTable\" ( \"RecordID\", \"Locale\", $fields_str)
-                    SELECT ? AS \"RecordID\", \"Locale\", $fields_str
-                    FROM \"$localisedTable\"
-                    WHERE \"RecordID\" = ?", [$toID, $fromID]);
-
-            // Make sure to update any has_one IDs - otherwise the new localised table entry will
-            // have the has_one ID from the original record.
-            // The current $owner (i.e. the newly duplicated record) may have already had its relation duplicated
-            // via cascade_duplicates prior to this extension hook being called.
-            // This only covers the current locale - see below for handling additional locales.
-            $fieldsToUpdate = [];
-            foreach (array_intersect($fields, $copyHasOneRelations) as $copyFieldName) {
-                $fieldsToUpdate[$copyFieldName] = $owner->{$copyFieldName};
-            }
-            if (!empty($fieldsToUpdate)) {
-                SQLUpdate::create($localisedTable, $fieldsToUpdate, ['RecordID' => $toID])->execute();
-            }
+            $this->duplicateLocalisedRows(
+                $tableName,
+                $localisedTable,
+                $fields,
+                $fromID,
+                $toID,
+                $original,
+                $owner,
+                $copyHasOneRelations
+            );
         }
 
         // We need to handle duplicating relations in `localised_copy` into additional locales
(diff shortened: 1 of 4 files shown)


src/Extension/FluentExtension.php
```diff
@@ -1745,4 +1750,79 @@ private function getLocaleCodesForModel(DataObject $model): array
...
+        }
+    }
+
+    private function generateUniqueLocalisedURLSegment(
+        string $localisedTable,
+        string $locale,
+        string $segment,
+        int $excludeRecordID
+    ): string {
+        $base = preg_replace('/-[0-9]+$/', '', $segment) ?: $segment;
```
wernerkrauss: IMHO this should use Silverstripe's URLSegmentFilter for replacing Umlauts and other chars.
```
$filter = URLSegmentFilter::create();
$base = $filter->filter($segment);
```

src/Extension/FluentExtension.php
```diff
@@ -1745,4 +1750,79 @@ private function getLocaleCodesForModel(DataObject $model): array
...
+        $candidate = $segment;
+        $count = 2;
+        while ($this->localisedURLSegmentExists($localisedTable, $locale, $candidate, $excludeRecordID)) {
+            $candidate = $base . '-' . $count++;
+        }
+        return $candidate;
+    }
+
+    private function localisedURLSegmentExists(
```
wernerkrauss: doesn't seem to take parent into account if owner has Hierarchy extension.

/foo/bar and /baz/bar are valid urls, but same URLSegment

tests/php/Extension/FluentSiteTreeExtensionTest.php
```diff
@@ -520,6 +520,64 @@ public static function localeFallbackProvider(): array
         ];
     }
 
+    /**
+     * Regression test for #252: duplicating a page must produce a unique
+     * URLSegment in every localised row, not just the base table.
+     */
+    public function testDuplicateProducesUniqueLocalisedURLSegment(): void
```
wernerkrauss, line 530: A test for having two pages with different parents but same URLSlug (e.g. foo/innovation-2 and bar/innovation) would be useful. 
duplicate bar/innovation and expect the URLSlug to be bar/innovation-2

tests/php/Extension/FluentSiteTreeExtensionTest.php
```diff
@@ -520,6 +520,64 @@ public static function localeFallbackProvider(): array
...
+    {
+        FluentState::singleton()->withState(function (FluentState $state): void {
```
wernerkrauss: This test doesn't seem to check, if any translations also have updated URLSegments? Where is another locale for this page?

wernerkrauss (changes requested):
Nice fix. Could you please improve it a bit?

adiwidjaja:
I have re-pushed with the changes in the right place and I think I have followed everything you have reviewed.

tractorcow:
Looks ok, but I have a general question. Could we fix the duplication logic so that it generates a unique URL at the actual duplication stage, rather than temporarily letting it persist a clash, and then attempt to fix it later? This might be more reliable, reduces DB writes, and avoids the scenario where we have a temporarily inconsistent state.

adiwidjaja:
For non-Fluent pages, the unique URL is created in SiteTree::onBeforeWrite, so that's in the duplication stage. The localised rows are created in FluentExtension::onAfterDuplicate, so we cannot correct them before. So IMO it's the right place, what do you think?

tractorcow:
> For non-Fluent pages, the unique URL is created in SiteTree::onBeforeWrite, so that's in the duplication stage. The localised rows are created in FluentExtension::onAfterDuplicate, so we cannot correct them before. So IMO it's the right place, what do you think?

Why can't we correct it in FluentExtension::onAfterDuplicate then? 

My feeling is that the current logic for preparing each localised row for copy deserves a rewrite.

```
            // Copy translations to duplicated object
            $fields_str = '"' . implode('","', $fields) . '"';
            DB::prepared_query("INSERT INTO \"$localisedTable\" ( \"RecordID\", \"Locale\", $fields_str)
                    SELECT ? AS \"RecordID\", \"Locale\", $fields_str
                    FROM \"$localisedTable\"
                    WHERE \"RecordID\" = ?", [$toID, $fromID]);
```


Perhaps we could shift the logic for preparing the row for the duplication into a separate function, one that can be overridden by both FluentVersionedExtension and FluentSiteTreeExtension? I.e. something called on each localised / version row?

Some logic like (pseudocode)


```
for each localised row:
  - get original localised row
  - generate duplicate localisation (our new function)
  - insert new row
```

and then we can override the duplicate generation function in any subclass that needs extra logic (e.g. handling URLSegment, versions).

The existing logic in onAfterDuplicate already seems to do a lot of unnecessary writes that should be condensed into a single write per row. If we have the ability to prepare a row before writing it to the DB, we shouldn't write it "wrong" once and then just SQLUpdate until it's fixed.

tractorcow:
Here's what I'm talking about; There's a delete, an insert, and possibly multiple updates. This could be condensed into a single conditional SQL update / insert.

```
            // Remove existing translations from duplicated object
            DB::prepared_query("DELETE FROM \"$localisedTable\" WHERE \"RecordID\" = ?", [$toID]);

            // Copy translations to duplicated object
            $fields_str = '"' . implode('","', $fields) . '"';
            DB::prepared_query("INSERT INTO \"$localisedTable\" ( \"RecordID\", \"Locale\", $fields_str)
                    SELECT ? AS \"RecordID\", \"Locale\", $fields_str
                    FROM \"$localisedTable\"
                    WHERE \"RecordID\" = ?", [$toID, $fromID]);

            // Make sure to update any has_one IDs - otherwise the new localised table entry will
            // have the has_one ID from the original record.
            // The current $owner (i.e. the newly duplicated record) may have already had its relation duplicated
            // via cascade_duplicates prior to this extension hook being called.
            // This only covers the current locale - see below for handling additional locales.
            $fieldsToUpdate = [];
            foreach (array_intersect($fields, $copyHasOneRelations) as $copyFieldName) {
                $fieldsToUpdate[$copyFieldName] = $owner->{$copyFieldName};
            }
            if (!empty($fieldsToUpdate)) {
                SQLUpdate::create($localisedTable, $fieldsToUpdate, ['RecordID' => $toID])->execute();
            }
```

adiwidjaja:
@tractorcow Ok, I tried to implement what you proposed. I implemented a `duplicateLocalisedRows` with a `prepareLocalisedRowForDuplicate` extension point in which FluentSiteTreeExtension does the URLSegment search. Is this what you meant?

src/Extension/FluentSiteTreeExtension.php
```diff
@@ -427,4 +430,148 @@ public function actionComplete($form, $message)
...
+        string $localisedTable,
+        string $baseTable,
+        string $locale,
+        string $segment,
+        int $excludeRecordID,
+        ?int $parentID
+    ): string {
+        $filter = URLSegmentFilter::create();
+        $segment = $filter->filter($segment) ?: $segment;
+        $base = preg_replace('/-[0-9]+$/', '', $segment) ?: $segment;
```
wernerkrauss, line 542: would be nice to get the existing numer directly saved to $count, so we can start from there

tractorcow:
General structure looks pretty good. I'll let @wernerkrauss and you tidy it up for merge 👍

src/Extension/FluentSiteTreeExtension.php
adiwidjaja: Ok, I finally got to change this, please review.

src/Extension/FluentExtension.php
```diff
@@ -934,6 +926,80 @@ protected function localiseManipulationTable(&$manipulation, $table, $localeTabl
...
+        $sourceRows = DB::prepared_query(
+            "SELECT $columnList FROM \"$localisedTable\" WHERE \"RecordID\" = ?",
+            [$fromID]
+        );
+
+        // Remove the localised row(s) that the initial duplicate write created so we can re-insert
+        // from the source row data in a controlled, single write per row.
+        DB::prepared_query("DELETE FROM \"$localisedTable\" WHERE \"RecordID\" = ?", [$toID]);
+
+        // The original SQLUpdate did not filter by locale, so all locales received the current
```
wernerkrauss: Should this comment be in the next Fluent versions? I doubt it...

wernerkrauss (changes requested):
looks good for me, but one comment looks like it was useful during development but should not be part of the PR

src/Extension/FluentExtension.php
adiwidjaja: Ok, I removed the comment.

wernerkrauss:
@tractorcow it should be tidied up now. Can you run the tests and merge?

adiwidjaja:
Hello @tractorcow, this is still open, do you see a problem merging it?

FYI @wernerkrauss

tractorcow:
@wernerkrauss you can make the final call

wernerkrauss (approved):

wernerkrauss:
@tractorcow I don't have write access, so I cannot merge it, and GitHub needs an approve from someone with write access...

